PR #14, opened 2016-06-16
Expose LineErr

pocke:
Thanks for cool library and  tool!

I develop a linter for dockerfile by golang. https://github.com/pocke/dflint
In this tool, I use this library as a shell parser.

I've implemented a feature to detect syntax errors of shell. 
However, `lineErr` is not exported. So, I haven't implemented displaying positions of error yet.

I propose to expose `lineErr`.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -502,25 +502,25 @@ func (p *parser) errPass(err error) {
 	}
 }
 
-type lineErr struct {
-	name string
-	pos  Pos
-	text string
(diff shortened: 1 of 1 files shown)


mvdan:
Makes sense, thanks!

pocke:
Thanks!

mvdan:
Note that now that we're exposing it I changed it a bit (hadn't thought hard about names and such before).

PR #40, opened 2016-11-01
cmd/shfmt: allow whitespace in shebangs so "#! interpreter [optional-arg]"  is allowed (ref.: man 2 execve)

dertuxmalwieder:


mvdan:
Where exactly? Can't find it on `man 2 execve`.

dertuxmalwieder:
An interpreter script is a text file that has execute permission enabled and
    whose first line is of the form:
        #! interpreter [optional-arg]

Note the whitespace, also note that the original [design specification](https://en.wikipedia.org/wiki/Shebang_(Unix)#History) included support for whitespaces in shebangs, so did [4.2BSD](http://www.in-ulm.de/~mascheck/various/shebang/#blankrequired).

mvdan:
You're right, it works under linux with both a space and a tab.

Then please:

* Add a test
* Use whitespace (`\s`) in the regex, not a space
* Change the commit message to `cmd/shfmt: allow whitespace in shebangs`

And if you quoted the manual page or the fact that it works under linux, even better. Otherwise by looking at the commit it's not obvious why this change was made.

mvdan:
Commit message titles can't be this long. Also, if you're having trouble fixing the regex or adding tests, I can just do the change myself.

dertuxmalwieder:
> Commit message titles can't be this long. 

I had no problem with a long commit message. -- I thought I fixed the regex?

mvdan:
Commit message titles shouldn't be more than 52 characters.

Regex is still the same as it was in the beginning - see the changes tab.

dertuxmalwieder:
`git` is repeatedly overtaxing me. This time it worked.

mvdan:
There was a weirdness with merge commits and this was still missing tests, so I've applied the changes myself while giving credit.

dertuxmalwieder:
Oh, OK. Thanks though!

PR #60, opened 2016-12-08
Add -version flag for reporting version

meatballhat:
I'm working on getting `shfmt` added to the default preinstalled stuff on the Travis CI VMs, and I'd like to be able to include the `shfmt` version in the VM metadata.  I realize that having a hardcoded version is a bit of a pain, and I'd typically implement this as an injected string of `git describe --always --tags` or similar, but didn't want to introduce such a big change without first proposing the flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: go
 
 go:
-  - 1.7
+  - 1.7.x
 
 script:
   - bash --version
   - go test ./...
   - go test -short -race ./...
+  - shfmt -version
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -17,10 +17,11 @@ import (
 )
 
 var (
-	write  = flag.Bool("w", false, "write result to file instead of stdout")
-	list   = flag.Bool("l", false, "list files whose formatting differs from shfmt's")
-	indent = flag.Int("i", 0, "indent: 0 for tabs (default), >0 for number of spaces")
-	posix  = flag.Bool("p", false, "parse POSIX shell code instead of bash")
+	write       = flag.Bool("w", false, "write result to file instead of stdout")
+	list        = flag.Bool("l", false, "list files whose formatting differs from shfmt's")
+	indent      = flag.Int("i", 0, "indent: 0 for tabs (default), >0 for number of spaces")
+	posix       = flag.Bool("p", false, "parse POSIX shell code instead of bash")
+	showVersion = flag.Bool("V", false, "show version and exit")
 
 	parseMode         syntax.ParseMode
 	printConfig       syntax.PrintConfig
@@ -29,10 +30,17 @@ var (
 	copyBuf = make([]byte, 32*1024)
 
 	out io.Writer
+
+	version = "v0.6.0"
 )
 
 func main() {
 	flag.Parse()
+
+	if *showVersion {
+		fmt.Println(version)
+		os.Exit(0)
+	}
 
 	out = os.Stdout
 	printConfig.Spaces = *indent
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -35,6 +35,10 @@ var (
 )
 
 func main() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s %s:\n", os.Args[0], version)
+		flag.PrintDefaults()
+	}
 	flag.Parse()
 
 	if *showVersion {
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -21,7 +21,7 @@ var (
 	list        = flag.Bool("l", false, "list files whose formatting differs from shfmt's")
 	indent      = flag.Int("i", 0, "indent: 0 for tabs (default), >0 for number of spaces")
 	posix       = flag.Bool("p", false, "parse POSIX shell code instead of bash")
-	showVersion = flag.Bool("V", false, "show version and exit")
+	showVersion = flag.Bool("version", false, "show version and exit")
 
 	parseMode         syntax.ParseMode
 	printConfig       syntax.PrintConfig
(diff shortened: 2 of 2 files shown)


mvdan:
> I'm working on getting shfmt added to the default preinstalled stuff on the Travis CI VMs

That's great! How would it be available?

> I realize that having a hardcoded version is a bit of a pain, and I'd typically implement this as an injected string of git describe --always --tags or similar, but didn't want to introduce such a big change without first proposing the flag.

Makes sense. I don't think I would go with build flags since that would break the regular `go get`/`go install`. Doing `go generate` with `git describe` is probably overkill as I would not release that often anyway.

So I think I'll leave it like this for now.

cmd/shfmt/main.go
```diff
@@ -29,11 +30,22 @@ var (
 	copyBuf = make([]byte, 32*1024)
 
 	out io.Writer
+
+	version = "v0.6.0"
 )
 
 func main() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s %s:\n", os.Args[0], version)
```
mvdan: I would rather keep the usage as it is now.

cmd/shfmt/main.go
```diff
@@ -29,11 +30,22 @@ var (
...
+		flag.PrintDefaults()
+	}
 	flag.Parse()
 
+	if *showVersion {
+		fmt.Println(version)
+		os.Exit(0)
```
mvdan: `return`

cmd/shfmt/main.go
```diff
@@ -17,10 +17,11 @@ import (
...
 var (
-	write  = flag.Bool("w", false, "write result to file instead of stdout")
-	list   = flag.Bool("l", false, "list files whose formatting differs from shfmt's")
-	indent = flag.Int("i", 0, "indent: 0 for tabs (default), >0 for number of spaces")
-	posix  = flag.Bool("p", false, "parse POSIX shell code instead of bash")
+	write       = flag.Bool("w", false, "write result to file instead of stdout")
+	list        = flag.Bool("l", false, "list files whose formatting differs from shfmt's")
+	indent      = flag.Int("i", 0, "indent: 0 for tabs (default), >0 for number of spaces")
+	posix       = flag.Bool("p", false, "parse POSIX shell code instead of bash")
+	showVersion = flag.Bool("V", false, "show version and exit")
```
mvdan: I realise `-v` is a bad idea since it can also be `verbose`, but I would use `-version` instead. Is `-V` popular in the wild? Haven't seen it yet.

meatballhat:
@mvdan I'm planning to get it added to all of the Ubuntu Trusty images such that a preinstalled version would be available most likely at `/usr/local/bin/shfmt`.  We have some bits in the image baking process that gather system info, which is what shows up in the `system_info` fold at the top, which is where I'd expect to show the `shfmt` version.

cmd/shfmt/main.go
meatballhat: I suspect `-version`/`--version` are more popular than `-V`, so def switch per your recommendation.

.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: go
 
 go:
-  - 1.7
+  - 1.7.x
```
mvdan: Out of curiosity, what does this do? Does it use the latest 1.7.x?

This should be clarified in the docs, as `a version with x in place of the minor or patch level` doesn't mean that to me.

cmd/shfmt/main.go
mvdan: Did you forget to amend this?

.travis.yml
meatballhat: 👍 I'll fix up the docs, and yes `1.7.x` means `1.7.{latest}`.

mvdan:
I'm probably doing some of this review wrong. I'm used to both the old PR system and gerrit, but not this new gerrit-like PR review system Github has. I'm probably not doing it in the simplest way.

.travis.yml
mvdan: I've seen tons of repos use `1.7` or `1.7.4`, so definitely be clear about this and recommend it in the docs.

meatballhat:
I don't know why this change didn't get picked up https://github.com/meatballhat/sh/blob/a7ffbd94456c4d465cafcdbef22049b143806ce8/cmd/shfmt/main.go#L40-L44 😁

cmd/shfmt/main.go
meatballhat: :shrug:

mvdan:
Oh - the point is that `os.Exit(0)` should be replaced by `return`. Both at the same time are redundant. And `os.Exit(0)` alone is unnecessarily using `os`.

meatballhat:
@mvdan ah, sorry.  I was trying to match style, but of course `os.Exit(0)` is different than the reasoning for `os.Exit(1)` https://github.com/meatballhat/sh/blob/a7ffbd94456c4d465cafcdbef22049b143806ce8/cmd/shfmt/main.go#L52-L58

.travis.yml
mvdan: Cue in commits on all my repos using this :)

mvdan:
I've squashed, amended the title and merged. Thanks again!

> I'm planning to get it added to all of the Ubuntu Trusty images such that a preinstalled version would be available most likely at /usr/local/bin/shfmt.

Does that mean it will be available on a default `sudo: false` Travis job? Or will it require `sudo: true` or the beta ubuntu images?

meatballhat:
@mvdan once I get it integrated and pushed out, it will be available on both:

``` yaml
sudo: false
dist: trusty
```

and

``` yaml
sudo: required
dist: trusty
```

mvdan:
@meatballhat please let me know once it's available - I'll add a note in the README since people are just downloading binaries or installing by hand now.

Also, how often will that preinstalled shfmt be updated?

.travis.yml
shurcooL: Mine too! https://github.com/shurcooL/cmd/pull/5 This is incredible!

meatballhat:
@mvdan I'll def ping here when available.  Updates are mostly sporadic.  If you think of it, getting the relevant bits in travis-cookbooks updated whenever you push a release is all that's needed.  I'm aiming to get accumulated updates pushed to "stable" on a monthly-ish cycle.

mvdan:
@meatballhat cool, thanks for the info. FWIW I intend to get 1.0 out next week, in case you want to wait for that.

meatballhat:
Nice!  That means 1.0 can ship with the January trusty push :tada:

mvdan:
@meatballhat 1.0 released.

PR #75, opened 2017-03-26
Fix Shellcheck errors

chrisfinazzo:


mvdan:
I appreciate the intent here, but please notice that this shell script is only to showcase the format that this tool enforces.

As such, any "potential bugs" aren't really applicable. If you have suggestions on how to make it more realistic they're welcome, but right now it's getting rid of examples like `$()` and `$(())`.

Issue #86, opened 2017-04-21
interp: add context to cancel/stop the interpreter


mvdan:
There is already a TODO on this, but this is to track it better.

Broken off #84 by @andreynering.

PR #88, opened 2017-04-22
interp: Add support for Context

andreynering:
Ref #86

I'm not familiar with the code base, so feel free to close if it's not the right way to do it.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -5,12 +5,14 @@ package interp
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/mvdan/sh/syntax"
 )
@@ -1036,5 +1038,32 @@ func TestRunnerOpts(t *testing.T) {
 					c.in, c.want, got)
 			}
 		})
+	}
+}
+
+func TestContext(t *testing.T) {
+	p, err := syntax.Parse(strings.NewReader("while true; do true; done"), "", 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	r := Runner{
+		File:    p,
+		Context: ctx,
+	}
+	endChan := make(chan struct{})
+
+	go func() {
+		_ = r.Run()
+		endChan <- struct{}{}
+	}()
+
+	cancel()
+
+	select {
+	case <-time.After(time.Millisecond * 100):
+		t.Error("Program was not killed in 0.15 seconds")
+	case <-endChan:
 	}
 }
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -5,6 +5,7 @@ package interp
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -67,7 +68,8 @@ type Runner struct {
 
 	bgShells sync.WaitGroup
 
-	// TODO: add context to kill the runner before it's done
+	// Context can be used to cancel the interpreter before it finishes
+	Context context.Context
 }
 
 // varValue can hold a string, an indexed array ([]string) or an
@@ -184,6 +186,9 @@ func (r *Runner) setFunc(name string, body *syntax.Stmt) {
 
 // Run starts the interpreter and returns any error.
 func (r *Runner) Run() error {
+	if r.Context == nil {
+		r.Context = context.Background()
+	}
 	if r.Env == nil {
 		r.Env = os.Environ()
 	}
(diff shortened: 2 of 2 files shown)


mvdan:
Thanks for the PR!

I'm on mobile, so this is a quick review.

Please add a test - like one that just runs `while true; do true; done` and cancels the runner immediately.

Also, I assume you want to pass the context to `exec.Cmd` somehow too.

mvdan:
Also, I think there would be ways for the runner to still keep on going forever with this patch. But this patch seems like a fine start; we can add more tests and fine-tune later.

Also, please lowercase the commit message and use "Fixes #86.".

andreynering:
@mvdan Updated.

You are right, it is still not ending in some situations (I tried the `sleep` command). But at least it finishes the infinite loop.

interp/interp_test.go
```diff
@@ -1038,3 +1041,34 @@ func TestRunnerOpts(t *testing.T) {
 		})
 	}
 }
+
+func TestContext(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	defer cancel()
+
+	p, err := syntax.Parse(strings.NewReader("while true; do true; done;"), "", 0)
```
mvdan: trailing `;` is redundant

interp/interp_test.go
```diff
@@ -1038,3 +1041,34 @@ func TestRunnerOpts(t *testing.T) {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	buff := bytes.NewBuffer(nil)
```
mvdan: `var buf bytes.Buffer`

interp/interp_test.go
```diff
@@ -1038,3 +1041,34 @@ func TestRunnerOpts(t *testing.T) {
...
 	}
 }
+
+func TestContext(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+	defer cancel()
+
+	p, err := syntax.Parse(strings.NewReader("while true; do true; done;"), "", 0)
+	if err != nil {
+		log.Fatal(err)
```
mvdan: `t.Fatal(err)`

interp/interp_test.go
```diff
@@ -1038,3 +1041,34 @@ func TestRunnerOpts(t *testing.T) {
 		})
 	}
 }
+
+func TestContext(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
```
mvdan: Why not make it cancel immediately? Otherwise, if I understand correctly, this test will take at least three seconds.

interp/interp_test.go
```diff
@@ -1038,3 +1041,34 @@ func TestRunnerOpts(t *testing.T) {
...
+	go func() {
+		if err := r.Run(); err != nil {
+			t.Fatal(err)
+		}
+		endChan <- struct{}{}
+	}()
+
+	select {
+	case <-time.After(time.Second * 6):
+		t.Error("Program was not killed in 6 seconds")
```
mvdan: If you make the timeout above much lower than `3s`, I assume this could easily be much lower like `1s`.

interp/interp_test.go
```diff
@@ -1038,3 +1041,34 @@ func TestRunnerOpts(t *testing.T) {
...
+	r := Runner{
+		File:    p,
+		Stdout:  buff,
+		Context: ctx,
+	}
+	endChan := make(chan struct{})
+
+	go func() {
+		if err := r.Run(); err != nil {
+			t.Fatal(err)
```
mvdan: I believe you can't run `t.Fatal` from a goroutine... Just ignore the error.

interp/interp_test.go
mvdan: In fact, I think you can just ignore Stdout (or if it crashes, `ioutil.Discard` and I'll fix that later)

interp/interp_test.go
mvdan: Ah, I see you call `cancel` immediately. Perhaps use `context.WithCancel` then?

interp/interp_test.go
```diff
@@ -1038,3 +1040,30 @@ func TestRunnerOpts(t *testing.T) {
...
+
+	go func() {
+		_ = r.Run()
+		endChan <- struct{}{}
+	}()
+
+	cancel()
+
+	select {
+	case <-time.After(time.Microsecond * 125):
```
mvdan: This is a little extreme :) use `0.1s` and I'm sure Travis will be happy.

interp/interp_test.go
andreynering: Updated

I kept `time.WithTimeout` because I think it can not end before the select below, but I change that if you want. Changed time to 0.125 seconds.

mvdan: You're right, but remember that `go test` already has a timeout as a fallback for us. Which is configurable, too. So no need to do this ourselves.

interp/interp_test.go
andreynering: Ops. I wanted to use Millisecond, not Microsecond. Updated.

Issue #68, opened 2017-02-22
cmd/shfmt: official docker images


mmlb:
I'd be great if you could provide a docker image on docker hub. My use case is to use it with drone.io, there are other CIs that can use docker images natively. Something like:

```
FROM alpine:latest
ENTRYPOINT ["/usr/bin/shfmt"]
ADD https://github.com/mvdan/sh/releases/download/v1.2.0/shfmt_v1.2.0_linux_amd64 /usr/bin/shfmt
RUN chmod +x /usr/bin/shfmt
```
would work for me.

And maybe the following for automated builds:
```
FROM golang:alpine
ENTRYPOINT ["shfmt"]
COPY . /go/src/app
WORKDIR /go/src/app
RUN set -ex \
       && apk add --no-cache --virtual .build-deps git \
       && go-wrapper download ./... \
       && go-wrapper install ./... \
       && apk del .build-deps 
```

mvdan:
Usually people would build their own docker images with multiple tools, not just `shfmt`:

https://github.com/search?l=Dockerfile&q=shfmt&type=Code&utf8=%E2%9C%93

In that list, there are also some people making small images with just this tool.

Why should I make an "official" image? How is that better than just an official binary?

mvdan:
reping @mmlb

mvdan:
I'm going to close this, as I don't see how an official docker image adds anything on top of an official static binary.

mmlb:
hey @mvdan 

So sorry to not respond, I somehow disabled email notifications and wasn't checking them from github. Its cool if you don't want to implement this, my use case was to not have to manage my own docker image and update whenever you updated, nor to have to download the static binary on each ci run.

Its kinda nice that shellcheck has docker images that track master. It would require more work than I would want to do if I wanted to replicate. I would need to wire up ifft or something else to watch github's rss feed of this repo and update my docker image.

I'll probably just have my own image and update that with your releases/master from time to time.

mvdan:
You should not always use `shfmt` from master :) Same goes for released versions, really. You should stick to a certain version and only update to newer versions after checking everything is still OK.

For example, I might add more simplification rules like the one that converts `` `foo` `` to `$(foo)`.

If you're updating to a newer `shfmt` version manually, building a tiny docker image doesn't keep you from fully automating the process on its own.

mmlb:
[...]

PR #121, opened 2017-06-10
Docker - Add examples to readme of using a dockersed image of `shfmt`.

jamesmstone:
I made a docker image of `shfmt`, so I thought I would share it. If you would like I would be happy to make this more official by putting it as part of your repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,8 @@ A shell parser, formatter and interpreter. Supports [POSIX Shell],
 [Bash] and [mksh]. Requires Go 1.7 or later.
 
 ### shfmt
-#### Go
+
 	go get -u github.com/mvdan/sh/cmd/shfmt
-#### Docker
-	docker pull jamesmstone/shfmt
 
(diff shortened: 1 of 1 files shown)


mvdan:
Thanks for your contribution!

Adding this to the related projects section seems like a great idea. Adding it at the top of the README not so much, especially given how this is not an official image.

@mmlb raised a similar proposal in #68, but we decided against official Docker images for now.

mmlb:
Right, @mvdan I've been thinking about adding `-d` support to `shfmt`, would you accept that? Missing `-d` is what what basically killed shfmt image for my use.

mvdan:
Could you open a separate issue about that? With a proposal on how it would be implemented, in particular.

jamesmstone:
cheers have now limited it to just the readme :)

mvdan:
I've taken the liberty of cleaning up the patch and applying it to master, keeping you as the author. Thanks again!

PR #124, opened 2017-06-14
gosh: allow -c flag

andreynering:
```bash
gosh -c "echo 'Hello, World'"
```

### DIFF
--- a/cmd/gosh/main.go
+++ b/cmd/gosh/main.go
@@ -6,17 +6,31 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io"
 	"os"
+	"strings"
 
 	"github.com/mvdan/sh/interp"
 	"github.com/mvdan/sh/syntax"
 )
 
-var parser *syntax.Parser
+var (
+	command = flag.String("c", "", "command to be executed")
+
+	parser *syntax.Parser
+)
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/gosh/main.go
```diff
@@ -15,9 +17,21 @@ import (
 var parser *syntax.Parser
 
 func main() {
+	var command bool
+	flag.BoolVar(&command, "c", false, "command to be executed")
```
mvdan: Put it as a global, like in `cmd/shfmt/main.go`

cmd/gosh/main.go
```diff
@@ -15,9 +17,21 @@ import (
...
 	flag.Parse()
 	parser = syntax.NewParser()
 
+	if command {
+		for _, c := range flag.Args() {
```
mvdan: Are you positive this is how `-c` works? Reading bash's man page just confused me. Saner shells seem to use `-c string` instead of arguments, which is how I assumed it worked.

In any case, your current implementation seems wrong:

```
$ bash -c 'echo foo' 'echo bar'
foo
```

andreynering: You are totally right! Updated.

mvdan:
Thanks!

PR #127, opened 2017-06-24
interp: fix redirect path when Runner.Dir option is used

andreynering:


mvdan:
Nice find - but please add a test :) I have others like this, just grep for `mkdir .* cd` in `interp_test.go`.

andreynering:
Test case added.

mvdan:
Oh no; what I meant is something like `mkdir -p a/b && cd a && cd b && cd ../..`, which tests it for `cd`. There is no need to add a full `Runner` test.

You could do something like: `mkdir a && cd a && echo foo >b && cd .. && cat a/b`

andreynering:
Hmm... you're right. Updated. :smile:

mvdan:
Thanks!

PR #128, opened 2017-07-01
interp: fix path in "test" builtin for a different Runner.Dir

andreynering:


interp/test.go
```diff
@@ -95,40 +95,43 @@ func (r *Runner) binTest(op syntax.BinTestOperator, x, y string) bool {
...
 	//case syntax.TsCharSp:
 	//case syntax.TsBlckSp:
 	case syntax.TsNmPipe:
-		return statMode(x, os.ModeNamedPipe)
+		return r.statMode(x, os.ModeNamedPipe)
 	case syntax.TsSocket:
-		return statMode(x, os.ModeSocket)
+		return r.statMode(x, os.ModeSocket)
 	case syntax.TsSmbLink:
 		info, _ := os.Lstat(x)
```
mvdan: You'll need to fix this one too, and any other os.Foo direct calls. Or if you're lazy, just leave TODOs.

andreynering: @mvdan I missed those. I abstracted it in a function to reduce code duplication.

mvdan:
Appreciated!

Issue #141, opened 2017-07-24
interp: consider portability of Linux and Unix-isms


mvdan:
A lot of the shell scripts nowadays do some or all of the following:

* Hard-code `/` as a path separator
* Make use of `/dev/` and `/proc/`, e.g. `/dev/null` and `/dev/urandom`
* Use `./foo` to call binaries in the current directory

This will break with `interp` on Windows, most notably.

The current solution to this is doing nothing about it. It's up to the shell script to be portable.

A slightly better version of this solution would be to provide some helper funcs like `filepath.Join`. This way, shell scripts could be updated to be more portable. Would require work from the user's side, though.

Then we have the option to automatically translate/emulate Linux/Unix-isms to other systems like Windows. This is sort of what Bash on Windows does. What it does (I think) is translate system calls, so at the low level, when one opens the file `foo/bar`, it actually opens `foo\bar`.

Unfortunately, we can't do the same as we don't have control over the system calls. In the case of path separators, we could try to replace them while interpreting a program, but this would break strings that weren't paths like `echo "foo / bar"`.

I'm not convinced there's a clean, good solution to this. Input and suggestions are very welcome.

/cc @andreynering, re #140

andreynering:
Note that many programming languages, including Go, accept both `\` and `/` as path separator on Windows.

So as example, this will work only on Windows:

```go
f, err := os.Open("path\\to\\file")
```

But this works everywhere, including on Windows:

```go
f, err := os.Open("path/to/file")
```

This means that relative paths are already working on gosh, including calling an executable as your example:

```go
./binary
```

The only things missing are:

1) Absolute path. It understands `C:\` and `C:/`, but not `/`
2) `/dev/*` files

About (1), maybe scripts should handle it. About (2), maybe it's only possible to cover the most simple cases (`/dev/null`, etc).

andreynering:
I didn't tried it, but I'm thinking in something like this for `/dev/null`:

```go
type DevNull struct{}

func (DevNull) Read(r []byte) (int, error) {
	return 0, io.EOF
}

func (DevNull) Write(w []byte) (int, error) {
	return len(w), nil
}

func (DevNull) Close() error {
	return nil
}
```

mvdan:
> This means that relative paths are already working on gosh, including calling an executable as your example:

Wow, I didn't know this. If that's the case, then why is `filepath` even a thing? Or is this automated only in certain calls like `os.Open`?

> Absolute path

[...]

PR #143, opened 2017-07-27
interp: have a custom /dev/null implementation

andreynering:
This fix reading and writing to /dev/null on Windows, where this file is not available.

It may also add a very minimal performance improvement on other plataforms, since now interp don't need to open and redirect to a real file anymore.

Reference: #141 

---

@mvdan Consider this a proposal, and feel free to reject it if you decide against it, or have a better idea about how to implement this.

### DIFF
--- a/interp/dev.go
+++ b/interp/dev.go
@@ -1,0 +1,21 @@
+package interp
+
+import (
+	"io"
+)
+
+var _ io.ReadWriteCloser = devNull{}
+
+type devNull struct{}
+
+func (devNull) Read(_ []byte) (int, error) {
+	return 0, io.EOF
+}
+
+func (devNull) Write(p []byte) (int, error) {
+	return len(p), nil
+}
+
+func (devNull) Close() error {
+	return nil
+}
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -671,10 +671,17 @@ func (r *Runner) redir(rd *syntax.Redirect) (io.Closer, error) {
 	case syntax.RdrOut, syntax.RdrAll:
 		mode = os.O_RDWR | os.O_CREATE | os.O_TRUNC
 	}
-	f, err := os.OpenFile(r.relPath(arg), mode, 0644)
-	if err != nil {
-		// TODO: print to stderr?
-		return nil, err
+	var f io.ReadWriteCloser
+	switch arg {
+	case "/dev/null":
+		f = devNull{}
+	default:
+		var err error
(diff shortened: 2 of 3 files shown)


interp/dev.go
```diff
@@ -0,0 +1,23 @@
+package interp
+
+import (
+	"io"
+)
+
+var (
+	_ io.ReadWriteCloser = devNull{}
+)
```
mvdan: style nit, but make this a single line like `var _ ...`

interp/interp.go
```diff
@@ -670,10 +670,17 @@ func (r *Runner) redir(rd *syntax.Redirect) (io.Closer, error) {
 	case syntax.RdrOut, syntax.RdrAll:
 		mode = os.O_RDWR | os.O_CREATE | os.O_TRUNC
 	}
-	f, err := os.OpenFile(r.relPath(arg), mode, 0644)
-	if err != nil {
-		// TODO: print to stderr?
-		return nil, err
+	var f io.ReadWriteCloser
+	switch arg {
+	case "/dev/null":
```
mvdan: Perhaps we should use something like `os.Stat` to make symlinks and stuff like `/dev/../dev/null` work, even if they're a bit silly. But this is fine for the first version.

interp/dev.go
andreynering: Fixed

interp/interp.go
andreynering: Paths like `/dev/../dev/null` can be usually fixed with [`filepath.Clean`](https://golang.org/pkg/path/filepath/#Clean). [Example](https://play.golang.org/p/jvdGjn8VXz).

But on Windows I think it may be converted to something like `C:\dev\null`. Will do a try later.

mvdan: We can think about this at a later time.

interp/interp_test.go
```diff
@@ -1127,6 +1127,10 @@ var fileCases = []struct {
 		"mkdir a; touch a/b.x; echo */*.x; cd a; echo *.x",
 		"a/b.x\nb.x\n",
 	},
+
+	// /dev/null
+	{"echo foo >/dev/null", ""},
+	{"cat /dev/null", ""},
```
mvdan: This is a bad test - you probably meant `cat </dev/null`, or just `</dev/null`.

andreynering: Oops. Fixed.

mvdan:
Thanks again!

Issue #152, opened 2017-08-25
cmd/shfmt: treats symlinks to directories as files when walking


soulshake:
Reproduced in [this sample repo](https://github.com/soulshake/tmp-shfmt-symlink-issue):

```
$ shfmt .
open dir1/dir2: is a directory

$ echo $?
1
```

mvdan:
Seems like the root cause is that `filepath.Walk` does an `lstat` instead of a `stat`, so it doesn't follow symlinks. Will review the PR, thanks!

PR #153, opened 2017-08-25
Don't test symlinks to directories

soulshake:
Addresses #152

I can try adding a test this weekend if you want.

### DIFF
--- a/fileutil/file.go
+++ b/fileutil/file.go
@@ -41,7 +41,7 @@ func CouldBeScript(info os.FileInfo) ScriptConfidence {
 	switch {
 	case info.IsDir(), name[0] == '.':
 		return ConfNotScript
-	case info.IsDir(), info.Mode()&os.ModeSymlink == os.ModeSymlink:
+	case info.Mode()&os.ModeSymlink == os.ModeSymlink:
 		return ConfNotScript
 	case extRe.MatchString(name):
 		return ConfIsScript
--- a/fileutil/file.go
+++ b/fileutil/file.go
@@ -41,6 +41,8 @@ func CouldBeScript(info os.FileInfo) ScriptConfidence {
 	switch {
 	case info.IsDir(), name[0] == '.':
 		return ConfNotScript
+	case info.IsDir(), info.Mode()&os.ModeSymlink == os.ModeSymlink:
+		return ConfNotScript
 	case extRe.MatchString(name):
 		return ConfIsScript
 	case strings.IndexByte(name, '.') > 0:


fileutil/file.go
```diff
@@ -41,6 +41,8 @@ func CouldBeScript(info os.FileInfo) ScriptConfidence {
 	switch {
 	case info.IsDir(), name[0] == '.':
 		return ConfNotScript
+	case info.IsDir(), info.Mode()&os.ModeSymlink == os.ModeSymlink:
+		return ConfNotScript
```
mvdan: Note that the commas here are OR, not AND. So `info.IsDir()` is a no-op since it's caught above.

Effectively this discards all symlinks, so it should be wrong. If you add tests for both symlinks to files and symlinks to dirs, this change should break symlinks to files - which I believe we want to keep working.

Also, you can simplify the second part to `info.Mode()&os.ModeSymlink != 0`.

mvdan:
By the way, I had a look at `gofmt`'s source code and it actually doesn't handle this case. It's easier for them because they discard any file that doens't have a `.go` extension.

So, in theory, you can have a `foo.go` symlink targetting `bar/` and it would error. Although that's just a terrible idea, so it's fine for it to error.

We have to be more careful here because we don't require files to have a shell extension to open them. In fact, if they don't have an extension, we open them to read the shebang.

fileutil/file.go
soulshake: Re: OR/AND: Thanks, that was a bad assumption on my part.

Re: discarding symlinks: Doesn't testing symlinks result in testing the same file multiple times? Also, doesn't that possibly end up testing (and possibly modifying in-place) files outside the context provided by the user?

mvdan: Yes, that can potentially lead to doing the same files multiple times. Which is OK for now since we don't parse multiple files at once, but it could be problematic sometime in the future if we're not careful.

Regarding files outside the context, I guess that depends on how you use symlinks and how you expect tools to treat them. I'd say either using them or discarding them would be OK things to do.

But in general I agree - trying to follow symlinks can lead to tricky situations and unexpected behavior for little benefit. Simply skipping all symlinks seems like a good approach for now. I'd rather be conservative and only reconsider the logic if anyone raises an issue about it.

Tests are less important if we just skip all symlinks, but they'd still be useful IMO. As said, feel free to leave that for me.

soulshake: I pushed another commit intending to drop all symlinks. I left `info.Mode()&os.ModeSymlink == os.ModeSymlink` instead of replacing it with `info.Mode()&os.ModeSymlink != 0` because I find the former more readable.

mvdan:
Thank you - I'll merge this and add the tests :)

mvdan:
I took the liberty to squash commits (as one was a fixup) and to format the commit message like the rest. I hope that's ok.

soulshake:
Of course! Thanks :)

PR #160, opened 2017-09-13
interp: cd now checks for exec permissions

### LINKED ISSUE #148: interp: cd should check if the dir has the right permissions
I believe we only need execute permissons for the current user to `cd` to the directory.

As suggested in https://github.com/mvdan/sh/issues/137#issuecomment-320436970.

CC @ebfe

andreynering:
This commit reintroduces bc1c6f7cf372d921cb3594b2d23a2cab08d5fd23, but with build tags to fix build on Windows.

Closes #148

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,6 @@ script:
   - go test -short -race ./...
   - shfmt -version
   - goveralls
+  # at least make sure the builds work
+  - GOOS=windows go install ./interp
+  - GOOS=darwin go install ./interp
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -161,6 +161,9 @@ func (r *Runner) builtinCode(pos syntax.Pos, name string, args []string) int {
 		if err != nil || !info.IsDir() {
 			return 1
 		}
+		if !hasPermissionToDir(info) {
+			return 1
+		}
 		r.Dir = dir
 	case "wait":
 		if len(args) > 0 {
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -428,6 +428,30 @@ var fileCases = []struct {
 		`mkdir a; ln -s a b; [[ $(cd a && pwd) == $(cd b && pwd) ]]; echo $?`,
 		"1\n",
 	},
+	{
+		`mkdir a; chmod 0000 a; cd a`,
+		"exit status 1 #JUSTERR",
+	},
+	{
+		`mkdir a; chmod 0222 a; cd a`,
+		"exit status 1 #JUSTERR",
+	},
+	{
+		`mkdir a; chmod 0444 a; cd a`,
+		"exit status 1 #JUSTERR",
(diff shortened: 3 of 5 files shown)


fileutil/perm_other.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func HasPermissionToDir(info os.FileInfo) (bool, error) {
+	st, _ := info.Sys().(*syscall.Stat_t)
+	if st == nil {
+		return true, nil
+	}
+
+	perm := info.Mode().Perm()
+	user, err := user.Current() // TODO: do this at start-up
```
mvdan: Another point in favor of keeping this inside `interp` - we can more easily hook into the internal `Runner` state, such as an initialised `os/user.User`.

fileutil/perm_other.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	var (
+		uid, _ = strconv.Atoi(user.Uid)
+		gid, _ = strconv.Atoi(user.Gid)
+	)
```
mvdan: Simplify with `:=`

fileutil/perm_other.go
```diff
@@ -0,0 +1,43 @@
+// +build !windows
```
mvdan: missing copyright notice (put 2017, and if it's a new file mostly written by you, your name - maybe we could eventually switch to an "sh authors" like Go)

fileutil/perm_other.go
```diff
@@ -0,0 +1,43 @@
...
+package fileutil
+
+import (
+	"os"
+	"os/user"
+	"strconv"
+	"syscall"
+)
+
+func HasPermissionToDir(info os.FileInfo) (bool, error) {
```
mvdan: given how we only use the bool here, have you thought about omitting either of the return values?

fileutil/perm_other.go
andreynering: Seems that `os/user` already caches the user struct, so no need to worry about that.

mvdan: I was thinking more along the lines of getting the value at startup, not on the first `cd`. But I guess it can't change anyway, so feel free to just remove the TODO.

mvdan:
Also, I just realised, how does this support super-users, e.g. root on linux? It doesn't seem to me like it would work.

andreynering:
Updated.

As a suggestion, you may want to add these commands to the Travis build, to know if it builds to non-linux OSes.

```sh
GOOS=windows go install ./interp
GOOS=darwin go build ./interp
```

mvdan:
> And it will have group permission?

What determines that is your code, and your code checks for `uid` and `gid` exact matches. I suspect it wouldn't work. But I can work on that after merging this.

mvdan:
Awesome, thanks!

PR #170, opened 2017-11-02
interp: send Interrupt to processes instead of Kill when possible

andreynering:
The code as is works. But I don't know do you would prefer to add this. Should it always send Interrupt or have an options to choose how to handle it? For an option: what would be the default?

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -80,6 +80,19 @@ type Runner struct {
 	stopOnCmdErr bool // set -e
 
 	dirStack []string
+
+	// KillTimeout holds how much time the interpreter will wait for a
+	// program to stop after being sent an interrupt signal, after
+	// which a kill signal will be sent. This process will happen when the
+	// interpreter's context is cancelled.
+	//
+	// The zero value will default to 2 seconds.
+	//
+	// A negative value means that a kill signal will be sent immediately.
+	//
+	// On Windows, the kill signal is always sent immediately,
+	// because Go doesn't currently support sending Interrupt on Windows.
+	KillTimeout time.Duration
 }
 
 // Reset will set the unexported fields back to zero, fill any exported
@@ -92,15 +105,16 @@ type Runner struct {
 func (r *Runner) Reset() error {
 	// reset the internal state
 	*r = Runner{
-		Env:     r.Env,
-		Dir:     r.Dir,
-		Params:  r.Params,
-		Context: r.Context,
-		Stdin:   r.Stdin,
-		Stdout:  r.Stdout,
-		Stderr:  r.Stderr,
-		Exec:    r.Exec,
-		Open:    r.Open,
+		Env:         r.Env,
+		Dir:         r.Dir,
+		Params:      r.Params,
+		Context:     r.Context,
+		Stdin:       r.Stdin,
+		Stdout:      r.Stdout,
+		Stderr:      r.Stderr,
+		Exec:        r.Exec,
+		Open:        r.Open,
+		KillTimeout: r.KillTimeout,
 	}
 	if r.Context == nil {
 		r.Context = context.Background()
@@ -137,17 +151,21 @@ func (r *Runner) Reset() error {
 	if r.Open == nil {
 		r.Open = DefaultOpen
 	}
+	if r.KillTimeout == 0 {
+		r.KillTimeout = 2 * time.Second
+	}
 	return nil
 }
 
 func (r *Runner) ctx() Ctxt {
 	c := Ctxt{
-		Context: r.Context,
-		Env:     r.Env,
-		Dir:     r.Dir,
-		Stdin:   r.Stdin,
-		Stdout:  r.Stdout,
-		Stderr:  r.Stderr,
+		Context:     r.Context,
+		Env:         r.Env,
+		Dir:         r.Dir,
+		Stdin:       r.Stdin,
+		Stdout:      r.Stdout,
+		Stderr:      r.Stderr,
+		KillTimeout: r.KillTimeout,
 	}
 	for name, val := range r.cmdVars {
 		c.Env = append(c.Env, name+"="+r.varStr(val, 0))
--- a/interp/module.go
+++ b/interp/module.go
@@ -8,19 +8,22 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"runtime"
 	"syscall"
+	"time"
 )
 
 // Ctxt is the type passed to all the module functions. It contains some
 // of the current state of the Runner, as well as some fields necessary
 // to implement some of the modules.
 type Ctxt struct {
-	Context context.Context
-	Env     []string
-	Dir     string
-	Stdin   io.Reader
-	Stdout  io.Writer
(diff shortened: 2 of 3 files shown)


interp/module.go
```diff
@@ -36,12 +37,25 @@ type Ctxt struct {
...
+
+	// send Interrupt on cancel instead of Kill (Go's default) when possible
+	go func() {
+		<-ctx.Context.Done()
+		sig := os.Interrupt
+		// Go on Windows doesn't support send Interrupt:
+		// https://golang.org/pkg/os/#Process.Signal
+		if runtime.GOOS == "windows" {
+			sig = os.Kill
+		}
```
mvdan: Will this ever be fixed, or is it fixable?

interp/module.go
```diff
@@ -36,12 +37,25 @@ type Ctxt struct {
 type ModuleExec func(ctx Ctxt, name string, args []string) error
 
 func DefaultExec(ctx Ctxt, name string, args []string) error {
-	cmd := exec.CommandContext(ctx.Context, name, args...)
+	cmd := exec.CommandContext(context.Background(), name, args...)
```
mvdan: Maybe you can avoid the context in the call entirely.

mvdan:
Here's another interesting question - what do shells like Bash and others do in these cases? I assume they pass whatever signal they got to the children.

interp/module.go
andreynering: Seems that no: https://github.com/golang/go/issues/6720

andreynering:
> What happens if the child process never actually finishes? Shouldn't there be a timeout after which a KILL is sent?

Ideally, no process should wait forever, but if it happen, interp will wait for it, too. A timeout is resonable.

> What do shells like Bash and others do in these cases? I assume they pass whatever signal they got to the children.

In my tests, when a Go program receives Interrupt, it is redirected to all child processes automatically.

I think maybe the problem describe in the Task issue is because maybe a Kill was sent right after the Interrupt, so the process didn't have time to do cleanup. (But that's a guess).

This PR is currently about sending Interrupt on context cancel, not about redirecting signals.

mvdan:
> A timeout is resonable.

I assume we should add it, then. Otherwise one could lock up a goroutine running the interpreter quite easily, if allowed to run arbitrary programs.

As long as the timeout is small and/or configurable, I'm okay with leaving the interrupts as a default. I know that killing things immediately is a bit of a nuclear solution. To be honest, I didn't know that was what `os/exec` did. Perhaps this could be fixed there?

andreynering:
> To be honest, I didn't know that was what `os/exec` did. Perhaps this could be fixed there?

I think changing that on `os/exec` would be a breaking change, and probably wouldn't be accepted.

I couldn't find a proposal for that [in the issues](https://github.com/golang/go/issues).

mvdan:
Still fine by me to merge, as long as tests are added and this is made configurable somehow. You could, for example, expose a `time.Duration` timeout for `Interrupt`. If 0, then `Kill` would be sent immediately. It should also be documented that on Windows, `Kill` is always sent directly.

andreynering:
Timeout added.

Misses tests. I can add later, but I'll have to compile and run another Go executable inside tests to test it.

mvdan:
Not really - you could simply run `bash -c 'sleep 10'` as a program that would sleep for 10 seconds. I'm not sure that will ignore interrupt signals, but I'm sure there's a bash builtin of some sort to do it. The tests depend on bash already, so it's fine to depend on it.

interp/interp.go
```diff
@@ -77,6 +77,16 @@ type Runner struct {
 	Context context.Context
 
 	stopOnCmdErr bool // set -e
+
+	// KillTimeout holds how much time the interpreter wait until it force-kill
+	// an process, if it didn't self-exit after receive an Interrupt signal.
+	//
+	// The zero value means kill after 5 seconds.
```
mvdan: I'd say 5 seconds is too high for a default - 2 seconds?

interp/interp.go
```diff
@@ -77,6 +77,16 @@ type Runner struct {
...
+	//
+	// A negative duration means kill immediatelly
```
mvdan: A negative value means that a kill signal will be sent immediately.

interp/interp.go
```diff
@@ -77,6 +77,16 @@ type Runner struct {
...
+	//
+	// On Windows it always kill immediatelly.
```
mvdan: On Windows, the kill signal is always sent immediately. (also explain why)

interp/interp.go
mvdan: Also rewrite to something like "The zero value will default to N seconds."

interp/interp.go
```diff
@@ -77,6 +77,16 @@ type Runner struct {
 	Context context.Context
 
 	stopOnCmdErr bool // set -e
+
+	// KillTimeout holds how much time the interpreter wait until it force-kill
+	// an process, if it didn't self-exit after receive an Interrupt signal.
```
mvdan: KillTimeout holds how much time the interpreter will wait for a program to stop after being sent an interrupt signal, after which a kill signal will be sent. This process will happen when the interpreter's context is cancelled.

andreynering:
@mvdan 

> Not really - you could simply run `bash -c 'sleep 10'` as a program that would sleep for 10 seconds. I'm not sure that will ignore interrupt signals, but I'm sure there's a bash builtin of some sort to do it. The tests depend on bash already, so it's fine to depend on it.

The sleep exits either by Interrupt and Kill. There's the [`trap` command](http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_12_02.html) that allow running code on Interrupt, but it can't prevent the script from exiting AFAIK. Also, it's [listed as a bindin](https://github.com/mvdan/sh/blob/8be3b585fdbbf34fd9cedc003114f1db74bcc9a5/interp/builtin.go#L20), but not implemented, so I can't use it on tests yet.

mvdan:
I meant something like `bash -c 'trap xxx; sleep 10'`. It should not matter that we can't implement trap particularly well.

andreynering:
Oh, I got it.

andreynering:
@mvdan Tests added. Let me know if this was what you had in mind.

I could not solve a data race on calls to `Process.Signal()` so skipped it in this test. I'm open to ideas on how to solve that, if possible.

andreynering:
I just opened an issue to know what the maintainers think about having that on stdlib: https://github.com/golang/go/issues/22757

mvdan:
Thank you - I hadn't gotten to properly reviewing this, so I'll wait to see if anything happens on that issue first.

mvdan:
@andreynering is this ready now? I see you fixed the race, but just to be sure.

andreynering:
Yes, it is ready.

interp/module.go
```diff
@@ -58,7 +81,7 @@ func DefaultExec(ctx Ctxt, name string, args []string) error {
 		return ExitCode(127)
 		// TODO: print something?
 	default:
-		return nil
+		return err
```
mvdan: nice spot, by the way

PR #178, opened 2017-11-29
interp: unset CDPATH during tests.

ebfe:
bash prints the pwd after changing directories when CDPATH is set, which
made some of the TestFileConfirm tests fail.


mvdan:
Thanks!

PR #184, opened 2017-11-30
interp: handle more printf formats

ebfe:
The idea is to let the fmt package do the heavy lifting and just translate the
format char/arg type as required. There might be hidden gotchas I've not yet
discovered, but it seems to work as expected so far.


mvdan:
Ah, this is like translating to regexes, but for printf. Makes sense.

PR #188, opened 2017-12-02
interp: support set -a

ebfe:
Also never export arrays.

This is what I used for testing:

```
unset X A AA
X=0
A=(a b c)
declare -A AA=([foo]=bar)
set -a
X=1
A[0]=A
AA[foo]=rab
echo $X $A ${AA[foo]}
/bin/sh -c 'echo $X $A ${AA[foo]}'
X=(a b c)
echo $X
echo ${X[1]}
/bin/sh -c 'echo $X'
```

bash
```
1 A rab
1
a
b

```

gosh
```
1 A rab
1
a
b

```

Not sure how to translate this into a proper testcase.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -78,6 +78,7 @@ type Runner struct {
 
 	stopOnCmdErr bool // set -e
 	noGlob       bool // set -f
+	allExport    bool // set -a
 
 	dirStack []string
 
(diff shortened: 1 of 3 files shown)


mvdan:
Please add a test for a variable being set before this option is set - i.e. whether or not `set -a` affects existing variables. I assume it does not, and I assume you got it right, but I mean it more as a regression test.

mvdan:
And the same goes for non-string variables never being exported - would be good to have that as a test. See the `env | grep` tests.

ebfe:
Update with more tests.

mvdan:
Appreciated!

PR #190, opened 2017-12-03
interp: support printf %d 1 2 3

ebfe:
This makes Runner.expand return the number of arguments actually consumed by
the format. printf then just iterates r.expand until all arguments are consumed
(given it consumes arguments at all).

```
printf "%d" 1 2 3
```

bash: `123`
gosh (before): `1`
gosh (after): `123`

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -95,7 +95,7 @@ func (r *Runner) builtinCode(pos syntax.Pos, name string, args []string) int {
 				r.outf(" ")
 			}
 			if expand {
-				arg = r.expand(arg, true)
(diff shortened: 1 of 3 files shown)


mvdan:
By the way - please remember to run `gofmt` on the files before committing. Otherwise my commits will change some of the lines you touched, making the git history a bit harder to understand.

ebfe:
> By the way - please remember to run `gofmt` on the files before
> committing. Otherwise my commits will change some of the lines you
> touched, making the git history a bit harder to understand.

Just noticed, sorry for that.

PR #191, opened 2017-12-03
interp: support read [-r] [name]...

ebfe:
Initial attempt at the `read` builtin. I'm not happy how this turned out. Most
of the complexity stems from the rules on when leading/intermediary/trailing
IFSs are stripped and when left completely untouched. Oh and escaping of
course. Comments?

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -4,6 +4,8 @@
 package interp
 
 import (
+	"bytes"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks for this - I had been dreading `read` for a while, among others like `getopts`. Let's polish it up once it's merged, which should be simple enough with the tests.

mvdan:
I did a quick commit cleaning up the code a bit. I didn't look into the actual implementation too much, as I don't understand how `read` works that well either. I might have a closer look tomorrow.

If you come up with ways to simplify or improve it, by all means, do send more patches.

PR #192, opened 2017-12-04
interp: simplify field splitting in the read builtin

ebfe:


interp/builtin.go
```diff
@@ -501,6 +436,61 @@ func (r *Runner) builtinCode(pos syntax.Pos, name string, args []string) int {
...
+		start, end int
+	}
+	var fpos []pos
+
+	runes := make([]rune, 0, len(s))
+	infield := false
+	esc := false
+	for _, c := range s {
+		if infield {
+			if r.ifsRune(c) && (raw || !esc) {
```
mvdan: You can rewrite both of these branches as `if infield == r.ifsRune(c) && (raw || !esc) { ... }`, I think

mvdan: I didn't think that through very much - the two bodies aren't the same.

PR #193, opened 2017-12-04
interp: slightly simplify readLine

mvdan:


mvdan:
@ebfe please take a look, since I'm not sure about this change.

ebfe:
Looks fine.

PR #195, opened 2017-12-10
interp: reject invalid identifiers in the read builtin

ebfe:
Previously read was able to create variables with invalid names. Check
that the names are valid identifiers before doing anything and error if
they are not. Also add a missing reslice in ifsFields().


mvdan:
You can use syntax.ValidName instead of reimplementing it, I think.

ebfe:
Oops, not sure how I missed that. Updated.

PR #197, opened 2017-12-11
interp: support getopts

ebfe:
Initial attempt at the getopts builtin. Tracking state across multiple getopts
invocations is a bit ugly (necessary for -abc shortcuts) and so is detecting
intervening assignments to OPTIND. State split into a separate getopts struct in
the hope it could be used to handle options to the builtins too.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -435,9 +435,48 @@ func (r *Runner) builtinCode(pos syntax.Pos, name string, args []string) int {
 
 		return 0
 
+	case "getopts":
+		if len(args) < 2 {
+			r.errf("getopts: usage: getopts optstring name [arg]\n")
+			return 2
+		}
+		optind, _ := strconv.Atoi(r.getVar("OPTIND"))
+		if optind-1 != r.optState.argidx {
+			if optind < 1 {
+				optind = 1
+			}
+			r.optState = getopts{argidx: optind - 1}
+		}
+		optstr := args[0]
+		name := args[1]
+		args = args[2:]
+		if len(args) == 0 {
+			args = r.Params
+		}
+		diagnostics := !strings.HasPrefix(optstr, ":")
+
+		opt, optarg, done := r.optState.Next(optstr, args)
+
+		r.setVarString(name, string(opt))
+		r.delVar("OPTARG")
+		switch {
+		case opt == '?' && diagnostics && !done:
+			r.errf("getopts: illegal option -- %q\n", optarg)
+		case opt == ':' && diagnostics:
+			r.errf("getopts: option requires an argument -- %q\n", optarg)
+		default:
+			if optarg != "" {
+				r.setVarString("OPTARG", optarg)
+			}
+		}
+		if optind-1 != r.optState.argidx {
+			r.setVarString("OPTIND", strconv.FormatInt(int64(r.optState.argidx+1), 10))
+		}
+
+		return oneIf(done)
+
 	default:
 		// "trap", "umask", "alias", "unalias", "fg", "bg",
-		// "getopts"
 		r.runErr(pos, "unhandled builtin: %s", name)
 	}
 	return 0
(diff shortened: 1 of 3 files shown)


mvdan:
Wow, jesus. I had a look at getopts the other day and I noped the fuck out. Props to you for having the patience.

interp/interp.go
```diff
@@ -83,6 +83,8 @@ type Runner struct {
 
 	dirStack []string
 
+	optState getopts // getopts
```
mvdan: Is the comment here useful?

interp/builtin.go
```diff
@@ -553,3 +595,45 @@ func (r *Runner) relPath(path string) string {
 	}
 	return filepath.Clean(path)
 }
+
+type getopts struct {
+	argidx  int
+	runeidx int
+}
+
+func (g *getopts) Next(optstr string, args []string) (opt rune, optarg string, done bool) {
```
mvdan: Perhaps `done == (optarg == "")`? Unless `string(opt)` or `optarg` can be empty string below.

interp/builtin.go
```diff
@@ -553,3 +595,45 @@ func (r *Runner) relPath(path string) string {
...
+	}
+
+	i := strings.IndexRune(optstr, opt)
+	if i == -1 {
+		// invalid option
+		return '?', string(opt), false
+	}
+
+	if i+1 < len(optstr) && optstr[i+1] == ':' {
+		if len(args) < g.argidx+1 {
```
mvdan: I'd prefer `if g.argidx >= len(args)`, as that is the common "if I can't use x as an index of y".

interp/builtin.go
```diff
@@ -435,9 +435,51 @@ func (r *Runner) builtinCode(pos syntax.Pos, name string, args []string) int {
...
+		default:
+			if optarg != "" {
+				r.setVarString("OPTARG", optarg)
+			}
+		}
+		if r.optState.argidx+1 != optind {
+			r.setVarString("OPTIND", strconv.FormatInt(int64(r.optState.argidx+1), 10))
+		}
+
+		if done {
```
mvdan: I believe you can do `return oneIf(done)`

interp/builtin.go
```diff
@@ -553,3 +595,45 @@ func (r *Runner) relPath(path string) string {
...
+	opt = opts[g.runeidx]
+	if g.runeidx+1 < len(opts) {
+		g.runeidx += 1
+	} else {
+		g.argidx += 1
+		g.runeidx = 0
+	}
+
+	i := strings.IndexRune(optstr, opt)
+	if i == -1 {
```
mvdan: I'd prefer `i < 0`, as that makes it clearer that we can never reach the code below with a negative `i`

interp/builtin.go
ebfe: That would be wrong in the case of an option with no arguments. But I agree this is ugly. Was thinking about a bufio.Scanner like API. getopts.Next(optstr string, args) bool and provide accessors for opt/optargs, so something like
```go
for opts.Next(optstr, args) {
       // use opts.Opt(), opts.Optarg()
}
```
would work.

interp/builtin.go
ebfe: done

interp/builtin.go
ebfe: done

interp/builtin.go
ebfe: done

interp/interp.go
ebfe: With struct named getopts probably not. Removed.

interp/builtin.go
mvdan: Ah, I see. I agree that something like `bufio.Scanner` would be easier to use.

However, at the moment there is a single use of this API, and it's not exported, so I'm not particularly concerned about going one way or another. I'll merge this for now, but feel free to rewrite it into the other style if you wish.

mvdan:
Appreciate it! Great code as usual :)

PR #201, opened 2017-12-20
interp: adapt new $PATH logic with $PATHEXT

mvdan:
For vanilla Windows, so that executing "foo" works with "foo.exe" both
in $PATH and in the current directory. Also other edge cases from
looking at os/exec.LookPath.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


mvdan:
@andreynering need your assistance with testing.

I know I should set up tests for this, but that'd be a lot of work and I believe that AppVeyor is slow and hard to deal with. I'll rely on you to let me know if anything is obviously broken, if that's ok.

mvdan:
Wait - I forgot part of the patch.

mvdan:
Done.

andreynering:
@mvdan It is still not working, and I'm debugging the issue.

Splitting with `:` is problematic, because on Windows paths can contain `:`: https://play.golang.org/p/HVJUg6LLLxt

I think fixing this will make it work. Maybe we need a custom split logic that ignore this pattern: `[A-z]:\/`.

andreynering:
Maybe it's too much work for you to be worth, but Microsoft distribute free VirtualBox machines with Windows for development: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

mvdan:
You mean splitting of `$PATH`? If so, we are pretty fucked - that workaround could still have unexpected issues, besides not being valid shell.

In any case, if we are to support plain Windows, I need to figure out a way to add regression tests for it. Or perhaps not support it at all - I'm starting to wonder if it's possible and/or worth everyone's time.

andreynering:
I think `PATH` can only have absolute paths, right? Then, I don't think it will cause problems, and this custom splitting logic would only work for Windows.

Also, it would not be hard to add unit tests for it. I can even try it myself if you want.

mvdan:
I am going to try to get the tests passing on Mac, and then on Windows. I figure that will make things easier, as it will let me rewrite and adapt the tests in two batches. I presume some will fail on Mac, and many more on Windows.

mvdan:
First step: https://ci.appveyor.com/project/mvdan/sh

mvdan:
Windows CI is set up - I'll continue this myself, as I no longer need someone else to test stuff.

andreynering:
Seems good.

PR #211, opened 2018-01-22
Open file read-only to check for shebang

### LINKED ISSUE #207: cmd/shfmt: should not fail when non-bash files are missing permissions
https://github.com/mvdan/sh/blob/master/cmd/shfmt/main.go#L200 tries to open all files for reading and writing (O_RDWR) if the `-w` flag is given. This seems to be used even for checking if it's a script at all.
If a file has no write permissions for the current user and it's not a script shfmt should not fail.

A fix would be to either open the files as O_RDONLY first and only if it's a script attempt to open for writing as well or the ignore any errors when opening files (possibly controlled by a flag).

dominikschulz:
If a file is only readable but not writeable and does not
contain a valid shebang shfmt should not fail.

This implementation always opens the file twice.
These two additional syscalls seem negligible compared to
the added complexity to only open it twice if the write flag
is set.

Fixes #207

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -195,11 +195,7 @@ func walk(path string, onError func(error)) {
 }
 
 func formatPath(path string, checkShebang bool) error {
-	openMode := os.O_RDONLY
-	if *write {
-		openMode = os.O_RDWR
-	}
-	f, err := os.OpenFile(path, openMode, 0)
+	f, err := os.Open(path)
 	if err != nil {
 		return err
 	}
@@ -240,10 +236,11 @@ func formatPath(path string, checkShebang bool) error {
 			}
 		}
 		if *write {
-			if err := f.Truncate(0); err != nil {
+			if err := f.Close(); err != nil {
 				return err
 			}
-			if _, err := f.Seek(0, io.SeekStart); err != nil {
+			f, err = os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, 0)
+			if err != nil {
 				return err
 			}
 			if _, err := f.Write(res); err != nil {


dominikschulz:
Thank you very much for your feedback. Of course we shouldn't open the file twice if it's not necessary and after you pointed me at it, I realized that it isn't that hard at all.

I've updated my PR to reflect this. Please let me know what you think.

cmd/shfmt/main.go
```diff
@@ -240,10 +236,15 @@ func formatPath(path string, checkShebang bool) error {
...
-			if err := f.Truncate(0); err != nil {
+			fi, err := f.Stat()
+			if err != nil {
+				return err
+			}
+			if err := f.Close(); err != nil {
 				return err
 			}
-			if _, err := f.Seek(0, io.SeekStart); err != nil {
+			f, err = os.OpenFile(path, os.O_WRONLY|os.O_TRUNC, fi.Mode())
```
mvdan: I'm fairly positive that you don't need to pass a real value for `perm` here. That is only used for `O_CREATE` - i.e., when opening a file for writing and the file doesn't exist.

For example, from `man 2 open` on linux:

> The  mode  argument  specifies  the file mode bits be applied when a new file is created.  This argument must be supplied when O_CREAT or O_TMPFILE is specified in flags; if neither O_CREAT nor O_TMPFILE is specified, then mode is ignored.

You could say "what if the file is deleted between the close and open calls", which is the kind of silliness you have to deal with when you want to open the same file twice. But I think that worrying about that technical possibility is not worth anyone's time. If the user is touching the files while the tool is running, lots of other things could go wrong anyway.

As an example - on the Go tree, doing `git grep 'OpenFile(' | grep WRONLY | grep -v CREATE`, I can see many `OpenFile(path, os.O_WRONLY, 0)` calls. I believe we can do the same.

dominikschulz: Good to know, will update my PR. I was actually checking if we need to pass to perms to the open call or not, but must have missed that section you quoted.

IMHO the "worst" that could happen due to opening the file twice is that the disk runs full between close and open and the content gets lost. But this wasn't better before.

mvdan:
Ah, GitHub doesn't notify me if you push commits so I missed this yesterday. Thanks for the patch!

PR #227, opened 2018-04-16
syntax: add DebugPrint

mvdan:
A simpler, smaller version of ast.Fprint but for our AST. Include an
example too, which also serves as its test.

Field filtering is left out, as it seems like an overkill feature for
what is essentially a debugging function.

The reflect code isn't very complex, as our AST doesn't contain certain
types like maps and arrays, nor do any node structs contain unexported
fields.


mvdan:
@myitcv have a look, if you have five minutes to spare :)

PR #231, opened 2018-04-23
README: clarify that string indexes must be quoted

### LINKED ISSUE #226: Formatting quirk in bash associative array keys
Hi there,

I ran into a small quirk on bash associative array keys that may or may not be worth fixing.

**Example:**
```sh
declare -A my_array=(
  [foo-bar]='blah blah blah'
)
```
gets formatted to this

```sh
declare -A my_array=(
  [foo - bar]='blah blah blah'
)
```

I should also note that this is not an issue if the user quotes the entire key in this situation.

```sh
# this works fine
declare -A my_array=(
  ['foo-bar']='blah blah blah'
)
```

Thanks again! :wave:

mvdan:
As suggested by Derek Sifford in the issue below. Also add an example
where a string index gets changed due to how we format arithmetic
expressions.

While at it, reword some other parts and re-wrap all text to 80
characters.

Fixes #226.


mvdan:
@dsifford please take a look

dsifford:
LGTM

PR #246, opened 2018-06-02
syntax: add an option to space redirect ops

### LINKED ISSUE #243: syntax: support for spaces after redirection operators
Hi @mvdan,

Thanks for this awesome tool, I haven't known any of other shell script formatter before! Just wondering if it's possible to add a space before and after the redirection and read symbol so it'll be easier to read?

Thank you for your effort again!

mvdan:
That is, to format them like:

	echo foo > file

This format is somewhat popular in the wild, and it is mentioned in some
prominent documents like 'man bash' and Classic Shell Scripting.

The option is very specific and doesn't clash with any other existing
options, so it shouldn't complicate the printer in the long run.

The Google Shell Style guide doesn't explicitly say how to format
redirections, so for now don't change its recommendation on the main
README file.

Don't change the default, as that would break backwards compatibility.
It's also not clear that this new format is vastly more popular than the
current one. For example, the POSIX Shell spec doesn't use spaces, and
the shfmt tool didn't have support for them.

Fixes #243.

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -33,6 +33,7 @@ var (
 	indent      = flag.Uint("i", 0, "")
 	binNext     = flag.Bool("bn", false, "")
 	caseIndent  = flag.Bool("ci", false, "")
+	spaceRedirs = flag.Bool("sr", false, "")
 	keepPadding = flag.Bool("kp", false, "")
 	minify      = flag.Bool("mn", false, "")
 
@@ -75,6 +76,7 @@ Printer options:
   -i uint   indent: 0 for tabs (default), >0 for number of spaces
   -bn       binary ops like && and | may start a line
   -ci       switch cases will be indented
+  -sr       redirect operators will be followed by a space
   -kp       keep column alignment paddings
   -mn       minify program to reduce its size (implies -s)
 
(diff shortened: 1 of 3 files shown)


mvdan:
I'm more interested if the solution actually works for the scripts in your projects :) have you given it a go?

PeterDaveHello:
@mvdan haven't tested as I barely know about go, trying to read some guides to help understand how to test it from src, or maybe you know some quick methods wanna share ;) ?

mvdan:
Once you have Go installed (e.g. version 1.10.2), you should be able to do `go get -u mvdan.cc/sh/cmd/shfmt`. That will install master, so then you can do the following to checkout the branch and install the newer version:

```
cd $GOPATH/src/mvdan.cc/sh
git checkout space-redirs
go install mvdan.cc/sh/cmd/shfmt
```

alok:
Doesn't seem to work. `a>b` becomes `a >b` with `shfmt -sr`. I definitely installed shfmt correctly, since `shfmt -h` shows the `-sr` option.

alok:
Seems to work. The only case that failed (and I'm not even sure this is legal syntax) was `a<<(b)` which is meant to be `a < <(b)`.

mvdan:
I'm not sure what you mean by failed - `a<<(b)` is indeed invalid syntax, yes. It reads as a heredoc, but `(b)` isn't a valid heredoc delimiter. Bash errors on that input too.

alok:
In that case, the new option looks like it works correctly.

alok:
Ran this on all my old scripts and it LGTM.

mvdan:
Thank you both! I won't release 2.5.0 just yet, so if you need this sooner, just run master or pin a commit from master.

PeterDaveHello:
Thanks a lot!

PR #252, opened 2018-06-16
interp: "pwd" and "$PWD" should be absolute paths

andreynering:


interp/interp.go
```diff
@@ -229,7 +229,9 @@ func (r *Runner) Reset() error {
 		}
 		r.Dir = dir
 	}
-	r.Vars["PWD"] = Variable{Value: StringVal(r.Dir)}
+	if e := r.changeDir(r.Dir); e > 0 {
+		return ExitCode(e)
+	}
```
mvdan: This is a good catch - with a supplied `Dir`, we never checked if it existed or if it was absolute.

However, note that in the common `r.Dir == ""` case, this extra work is unnecessary. `os.Getwd` always returns an absolute path. So put this under an else.

Also, CI failed - just clear `OLDPWD` here.

andreynering: Hey @mvdan,

Thanks for reviewing it.

I changed the approach and am now converting to path manually to absolute because the test revealed a problem with it (Seems that it was joining path to times, e.g. `./a/a` when `r.Dir` is just `a`)

mvdan: That makes sense, actually - you were doing `r.changeDir(r.Dir)`, which ends up doing `filepath.Join(r.Dir, r.Dir)`.

PR #253, opened 2018-06-23
interp: avoid redundant array copies

Quasilyte:
Every range over array that binds a value does a full copy of array being iterated.
The workaround is to either use `&x` instead `x` or do index-only range.

In this case, shellOptsTable (192 bytes) is copied before each changed range stmt.
Not it does only pointer copy. Compiler optimizes accesses, so there is no
drawbacks for this. (And there is close to no difference for sizes of < 48 byte.)

See: https://github.com/cristaloleg/go-advices

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -536,7 +536,7 @@ func (r *Runner) builtinCode(pos syntax.Pos, name string, args []string) int {
 				}
 				break
 			}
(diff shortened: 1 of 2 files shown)


Quasilyte:
This is minor issue and I'm not sure these functions are anywhere near hot paths (probably they don't).
Sending just in case.

PR #254, opened 2018-06-23
syntax: replace if-else chains with expr switch stmt

Quasilyte:
Usually this form is considered more idiomatic.
I've left if-else chains with init statements untouched
to less obvious changes.

Found with `elseif` check from go-critic linter.


mvdan:
Great, thanks. I presumed so, but you never know.

PR #255, opened 2018-06-23
interp,syntax: replace single case switches

Quasilyte:
Usually makes code simpler.
There are no TODO notes, so I've treated
these switches as refactoring artifacts.


interp/module.go
```diff
@@ -132,8 +132,7 @@ func DefaultOpen(ctx Ctxt, path string, flag int, perm os.FileMode) (io.ReadWrit
 
 func OpenDevImpls(next ModuleOpen) ModuleOpen {
 	return func(ctx Ctxt, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
-		switch ctx.UnixPath(path) {
-		case "/dev/null":
+		if ctx.UnixPath(path) == "/dev/null" {
```
mvdan: please leave this as is - it's likely we're going to add others here, like `/dev/urandom` and `/dev/zero`.

Quasilyte:
Done. (You may want to squash merge as I didn't `--amend` this commit.)

Issue #267, opened 2018-07-18
all: /v2 import paths require newer versions of Go


rmohr:
I run into the following issues if I do `go get` for latest master:

```bash
$ go get -u mvdan.cc/sh/cmd/shfmt
package mvdan.cc/sh/v2/fileutil: cannot find package "mvdan.cc/sh/v2/fileutil" in any of:
	/home/rmohr/.gimme/versions/go1.8.3.linux.amd64/src/mvdan.cc/sh/v2/fileutil (from $GOROOT)
	/home/rmohr/go/src/mvdan.cc/sh/v2/fileutil (from $GOPATH)
package mvdan.cc/sh/v2/syntax: cannot find package "mvdan.cc/sh/v2/syntax" in any of:
	/home/rmohr/.gimme/versions/go1.8.3.linux.amd64/src/mvdan.cc/sh/v2/syntax (from $GOROOT)
	/home/rmohr/go/src/mvdan.cc/sh/v2/syntax (from $GOPATH)
```

suyanlong:
me too with golang1.10

mvdan:
You need to use Go 1.9.7 or 1.10.3 or later, since that is when the minimal support for Go modules was backported. See https://golang.org/doc/devel/release.html#go1.10.minor for more details.

Also note that this software requires Go 1.9 or later, as it is clearly stated in the README. I'll add a note about the modules part too.

And, as a general recommendation, you should not depend on master. I try not to break it, but I will make mistakes from time to time. You should be using the latest release, which is v2.5.0 at the moment. It doesn't include this v2 import path either, as I did the switch to Go modules after the release.

rmohr:
> Also note that this software requires Go 1.9 or later, as it is clearly stated in the README. I'll add a note about the modules part too.

Yes, I did it with 1.10 too, just posted the 1.8 output because it was the same.

> And, as a general recommendation, you should not depend on master. I try not to break it, but I will make mistakes from time to time. You should be using the latest release, which is v2.5.0 at the moment. It doesn't include this v2 import path either, as I did the switch to Go modules after the release.

Yes, we were actually using v2.3. It was just that we were using "go get" to fetch the sources and then checkout out v2.3 from it. We are now cloning the repo with git.

rmohr:
Here with the original version:

```
go version go1.10 linux/amd64
package mvdan.cc/sh/v2/fileutil: cannot find package "mvdan.cc/sh/v2/fileutil" in any of:
[...]

PR #268, opened 2018-07-18
Remove `V2` path

suyanlong:


mvdan:
This was done on purpose, and should not be reverted - see my reply to #267.

PR #276, opened 2018-07-24
README: Add another Dockerized shfmt / Docker repo

PeterDaveHello:
This Docker repository releases with corresponding upstream version tag, updates more frequently, and also brings basic shell utils provided `busybox` which could be helpful, especially for those Docker native CI environment like GitLab CI (So that we don't need to override the entrypoint).

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,7 @@ See the [_js](_js) directory for more information.
 
 ### Related projects
 
-* [dockerised-shfmt] - A docker image of `shfmt`
-* [dockerized-shfmt] - Another docker image of `shfmt` with versioning and basic shell
+* Docker images - by [jamesmstone][dockerized-jamesmstone], [PeterDaveHello][dockerized-peterdavehello]
 * [format-shell] - Atom plugin for `shfmt`
 * [micro] - Editor with a built-in plugin for `shfmt`
 * [shell-format] - VS Code plugin for `shfmt`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,6 +93,7 @@ See the [_js](_js) directory for more information.
 ### Related projects
 
 * [dockerised-shfmt] - A docker image of `shfmt`
+* [dockerised-shfmt] - Another docker image of `shfmt` with versioning and basic shell
```
mvdan: Please rephrase this, like:

```
* Docker images - by [jamesmstone][dockerized-jamesmstone], [PeterDaveHello][dockerized-peterdavehello]
```

README.md
```diff
@@ -102,6 +103,7 @@ See the [_js](_js) directory for more information.
 [bash]: https://www.gnu.org/software/bash/
 [crux]: https://github.com/6c37/crux-ports-git/tree/3.3/shfmt
 [dockerised-shfmt]: https://hub.docker.com/r/jamesmstone/shfmt/
+[dockerized-shfmt]: https://github.com/PeterDaveHello/dockerized-shfmt
```
mvdan: The change proposed above will also fix these links. You can't have two named links with the same name.

README.md
PeterDaveHello: Done, thanks.

README.md
PeterDaveHello: I tried to use dockeri**z**ed-shfmt than dockeri**s**ed-shfmt here :smile: Now revised.

PR #296, opened 2018-10-01
CI: add goreleaser config

docwhat:
This should make releasing easier.

The documentation is at https://goreleaser.com/

You can set this up so that Travis or Appveyor uploads the GitHub
releases for you.

I also added sections for Docker Hub and Homebrew.  Delete them if you
don't want to support them; though I'd like both. :-)

Docker images are useful for CI and Homebrew is handy on my workstation.

There are plenty of other packaging formats, include `.deb`, `.rpm`, snaps, etc.  I don't have a lot of experience with them, so I didn't set them up.


mvdan:
`shfmt` is already on homebrew-core, so I don't see the advantage of putting it there: https://formulae.brew.sh/formula/shfmt

I also don't see a big advantage in having my own docker image. Plenty of other people have their own images (see the README), so you can pick from any of them. A docker image is also not a necessity here, as we provide fully static binaries for every platform.

As for releasing on GitHub itself, that takes about five minutes of my time every couple of months. I think automating that, especially when it adds a new tool to learn and maintain, is over-optimizing somewhat :)

I'll leave this PR open for now, in case I'm missing something here, but my first impression is that this is unnecessary complexity.

docwhat:
Ah, yeah, you already have an official Homebrew package.

I would really like you to push out the docker images though.  I know it is "easy" but if you have an official image then everyone who wants to use `shfmt` in a docker image won't have to do it themselves (and maintain it).

Triggering docker image builds when you release a new image is more work than having you just automatically build them for us.

The docker image is great for CI. :-)

As for GitHub releases, it does more than just create a release... it creates a Changelog as well.  You can see it on my project: <https://github.com/docwhat/go-importd/releases>  I started using goreleaser as of version 0.1.5.

I have my project set up so I just create a tag (e.g., `v1.0.0.`) and push it up and Travis does the rest for me.  It did require adding some encrypted variables for dockerhub and github, which is more annoying than I would like.

However, the fact that I can tweak the release by editing the [well documented](https://goreleaser.com/customization/) `.goreleaser.yml` and test it with `--snapshot` makes it very easy.

mvdan:
> I would really like you to push out the docker images though.

I may decide to publish my own Docker image at some point, but automation isn't the issue here. I could simply have Travis or Docker Hub automate the builds for me for each tag.

> it creates a Changelog as well

What I can see is a list of commits, which isn't particularly useful to me. Each release has dozens if not hundreds of commits. If you look at my releases page, I spend five minutes crafting a useful changelog with the noteworthy changes in each release.

So, all in all, thanks for the pull request but I don't think this makes sense for me. In the future, it would be good if you filed an issue before a PR, and I would have told you I didn't need the tool before you posted the PR :)

PR #301, opened 2018-10-18
CI: reject non-gofmted code

ebfe:


mvdan:
Any reason why you want this in particular? Have I forgotten to run `gofmt`? :)

I haven't done this because this isn't trivial. `gofmt` changes at every Go release, so you could end up with Go 1.10 and 1.11 disagreeing over the format.

However, in this particular case it's a simple fix. Simply check that we're on the latest Go version tested on CI (should be 1.11) and then check the format.

ebfe:
> Any reason why you want this in particular? Have I forgotten to run `gofmt`? :)

No, but I have (https://github.com/mvdan/sh/pull/190) ;). Forgot to submit this.
 
> I haven't done this because this isn't trivial. `gofmt` changes at every Go release, so you could end up with Go 1.10 and 1.11 disagreeing over the format.
> However, in this particular case it's a simple fix. Simply check that we're on the latest Go version tested on CI (should be 1.11) and then check the format.

I've added a variable to the build matrix to address this. Only enabling
this in one build configuration should be enough and it avoids the "there
is no diff installed"-problem when enabling windows builds.

PR #316, opened 2018-10-28
Update .travis.yml

rpsene:
Add an entry to enable building on ppc64le (IBM POWER Architecture)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ matrix:
       env: CHECK_GOFMT=1
     - os: osx
       go: 1.11.x
+    - os: linux-ppc64le
+      go: 1.11.x
 
 go_import_path: mvdan.cc/sh
 


mvdan:
Thanks for the patch! Is there a specific reason to test this on ppc64le? Don't get me wrong, testing on more environments is better, but I wonder if there's a specific reason to include POWER and not others like ARM or MIPS. Have I broken POWER recently, or is there a prominent user that's interested in it being well supported?

I'm just curious because the code isn't particularly portable across operating systems (hence why I test on Mac and Windows), but the code has zero arch-dependent code, assumint its dependencies are fully portable.

I'm also curious as to where this `linux-ppc64le` value comes from. I've tried to look up online if there are any other architectures that I could use, but I can't even find `ppc64le` being documented anywhere. As far as I can see from the logs, this looks like a VM environment, since sudo seems to be installed. Where did you learn about this from?

rpsene:
@mvdan we ported Travis to work on IBM Power. Part of Travis worker uses this repository during the build process, which is currently failing when using an automation to download a released package for ppc64le. When we decided to make Travis work on Power, we chose linux-ppc64le as the keyword to indicate you want to build on Power. You may not find it in Travis docs yet, it is still in beta. More on ppc64le at https://www.ibm.com/developerworks/library/l-power-little-endian-faq-trs/index.html

mvdan:
> Part of Travis worker uses this repository during the build process, which is currently failing when using an automation to download a released package for ppc64le. 

Gotcha. The binaries available on the releases page has nothing to do with what's tested with CI, though. I could have twice the number of binaries there, but I only build for the most common platforms.

It shouldn't matter anyway; building this program with Go should be trivial and take under ten seconds, as there are no dependencies outside of the standard library.

mvdan:
Timing out. As said above, I'm not sure I understand the issue. Even if the issue was the lack of ppc64le binaries, its addition on Travis wouldn't help that.

PR #355, opened 2019-02-02
Force to uppercase, paths contains in PATH environment variables

goldfix:
on Windows:
when PATH env. variable contains paths with first letter start with lowercase (drive name), does not resolve and does not find file to execute.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -1196,7 +1196,7 @@ func splitList(path string) []string {
 	// join "C", "/foo" into "C:/foo"
 	var fixed []string
 	for i := 0; i < len(list); i++ {
-		s := list[i]
+		s := strings.ToUpper(list[i]) //on Windows: useful when %PATH% contains path that starts with lowercase disk name
 		switch {
 		case len(s) != 1, s[0] < 'A', s[0] > 'Z':
 			// not a disk name


mvdan:
Thanks! Do you have an example of how this could happen, i.e. a way to reproduce this issue? I'd like to add a test too, since we run the tests on Windows via Travis.

goldfix:
Hi Daniel,

maybe this is useful.

Via windows command line:
`mkdir c:\tmp\test_path_folder`    #create temporary folder
`echo @echo off>c:\tmp\test_path_folder\test.cmd`    #create an "executable" file
`set PATH=%PATH%;c:\tmp\test_path_folder\`      #set PATH with lowercase drive disk
`gosh.exe -c "test.cmd"`    #Error: "test.cmd": executable file not found in $PATH - exit status 127

but if set:
`set PATH=%PATH%;C:\tmp\test_path_folder\`      #set PATH with uppercase drive disk
`gosh.exe -c "test.cmd"`      #It's ok

mvdan:
Thanks! I'll take a look at writing a test for this.

mvdan:
I've implemented this in a slightly simpler way, with a proper test, and a descriptive commit message. It was faster to do this myself than to ask that of you; hope that's okay with you.

goldfix:
Perfect! 

Thanks Daniel.

p

PR #364, opened 2019-03-09
Fix Params

bilus:
This PR fixes two serious issues with `Params` passed to `runner.New`:

1. Passing `Params` causes a panic due to an untimely call to `updateExpandOpts`. This seems to be only there because `set` called from script requires it.
2. `Reset` called on first call to `Run` resets shell options.


mvdan:
I've sent https://github.com/mvdan/sh/pull/366, which I think is a simpler fix; I hope that's okay. I also credited you in the commit message.

mvdan:
#366 is merged, so I'm closing this for now. Feel free to send follow-up issues or PRs if I missed anything. Thanks!

PR #366, opened 2019-03-09
interp: fix using Params as an option to New

### LINKED ISSUE #365: interp: Params isn't supported as an option to New
First of all, great project. We're tentatively planning to use it in an open project we'll be launching in two weeks (an automation tool with a twist) as syntax for our Makefile-like tasks. There are plenty of opportunities there as opposed to using bash directly, including better error messages and even debugging!

While integrating it with our tool, I noticed a problem with `Params`. While I was working on a fix, tests I added uncovered another issue. 

PR: https://github.com/mvdan/sh/pull/364

I hope you'll find it not entirely worthless and can merge it, it's vital for us here, would hate to have to create a fork. :)

mvdan:
As correctly pointed out by Marcin Bilski, using the exported function
in that way would lead to a panic. And, even after working around the
panic, any shell flags used would be overwritten by the first implicit
call to Reset.

To fix the crash, move the updateExpandOpts method call to the end of
the "set" builtin. After all, it's only needed there; the Run function
already ends up calling the method later via fillExpandConfig.

To keep the shell options from being deleted in Reset, simply whitelist
them from the list of fields to clear. We do that for Params, so it
follows that the same should happen for the opts field.

Finally, add a couple of tests covering both scenarios to
TestRunnerOpts.

Fixes #365.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -262,7 +262,6 @@ func Params(args ...string) func(*Runner) error {
 			args = args[1:]
 		}
 		r.Params = args
-		r.updateExpandOpts()
 		return nil
 	}
 }
@@ -501,6 +500,7 @@ func (r *Runner) Reset() {
 		Exec:        r.Exec,
 		Open:        r.Open,
 		KillTimeout: r.KillTimeout,
+		opts:        r.opts,
 
 		// emptied below, to reuse the space
 		Vars:      r.Vars,
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -70,6 +70,7 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			r.errf("set: %v\n", err)
 			return 2
 		}
+		r.updateExpandOpts()
 	case "shift":
 		n := 1
 		switch len(args) {
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -2743,6 +2743,16 @@ func TestRunnerOpts(t *testing.T) {
 			"[[ $PWD == foo ]]",
 			"exit status 1",
 		},
+		{
+			opts(Params("foo")),
+			"echo $@",
+			"foo\n",
+		},
+		{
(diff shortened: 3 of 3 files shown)


mvdan:
@bilus could you please review this PR and see if it fixes the two bugs you found? Thanks!

bilus:
Are you planning to backport it to v2? If not, how stable is v3 for production use *right now*? We want to launch next week internally (i.e. to ~100 developers) with a public release a week after.

interp/interp.go
```diff
@@ -501,6 +500,7 @@ func (r *Runner) Reset() {
 		Exec:        r.Exec,
 		Open:        r.Open,
 		KillTimeout: r.KillTimeout,
+		opts:        r.opts,
```
bilus: I thought of that but it *think* leaks shell options `set` in a script from one `Run` to another. Is this the intended behaviour?

bilus:
> @bilus could you please review this PR and see if it fixes the two bugs you found? Thanks!

All bugs are captured in tests I added but I'll test it, probably today.

mvdan:
This seems like a bad bug in master too, so I'll probably backport. I probably won't do a v2 release right away though, as the vast majority of my user base is for `cmd/shfmt`, which is unaffected by the interpreter. You'll of course still be able to pin a commit from master.

I'm fairly confident that my tests should be equivalent to yours, so I'll go ahead and merge. If it turns out I missed something, please feel free to send more issues and PRs.

interp/interp.go
mvdan: Yes, I've been thinking the same overnight. This is intended for now, because `Runner.opts` should be in the same camp as `Runner.Params`. They can both be set by the `Params` function, and by the `set` builtin.

Like you say, this isn't ideal, because `Reset` should leave constructor options intact, while undoing changes done by scripts. Still, it's better if both fields behave consistently for now; we can always refine the API to be saner in v3. That kind of change would likely be too invasive to backport to v2.

PR #368, opened 2019-03-12
Rewrite if-else-if-else chain as a switch

sosiska:
This commit rewrites if-else-if-else chain as a switch.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch

### DIFF
--- a/expand/param.go
+++ b/expand/param.go
@@ -92,23 +92,24 @@ func (cfg *Config) paramExp(pe *syntax.ParamExp) (string, error) {
 		str = strconv.Itoa(n)
 	case pe.Excl:
 		var strs []string
-		if pe.Names != 0 {
+		switch {
+		case pe.Names != 0:
 			strs = cfg.namesByPrefix(pe.Param.Value)
-		} else if orig.Kind == NameRef {
+		case orig.Kind == NameRef:
 			strs = append(strs, orig.Str)
-		} else if vr.Kind == Indexed {
+		case vr.Kind == Indexed:
 			for i, e := range vr.List {
 				if e != "" {
 					strs = append(strs, strconv.Itoa(i))
 				}
 			}
-		} else if vr.Kind == Associative {
+		case vr.Kind == Associative:
 			for k := range vr.Map {
 				strs = append(strs, k)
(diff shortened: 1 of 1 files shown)


syntax/braces.go
```diff
@@ -107,11 +107,12 @@ func SplitBraces(word *Word) *Word {
 				broken := false
 				for i, elem := range br.Elems[:2] {
 					val := elem.Lit()
-					if _, err := strconv.Atoi(val); err == nil {
-					} else if len(val) == 1 &&
-						'a' <= val[0] && val[0] <= 'z' {
+					switch _, err := strconv.Atoi(val); {
```
mvdan: I don't think this change is worth it; the if-elseif chain only had three elements, and this new define-switch is more complex than needed.

syntax/parser.go
```diff
@@ -2046,7 +2046,8 @@ func (p *Parser) wordIter(ftok string, fpos Pos) *WordIter {
 		return wi
 	}
 	p.got(_Newl)
-	if pos, ok := p.gotRsrv("in"); ok {
+	switch pos, ok := p.gotRsrv("in"); {
```
mvdan: Like in the previous case, I think using a define-switch here is overkill for just three cases.

syntax/braces.go
sosiska: Maybe then return the if-else, but remove the empty block?
```go
if err == nil
```
?

mvdan: Maybe - feel free to take a stab at that.

sosiska: OK, it was a dumb question. It means if err == nil do nothing/

mvdan: well, if there was no error, there's nothing else to do. if there was an error, we move on to the other cases.

sosiska: Done in e336613e7883ab4a30afe31f6d84c28752fa2f75.

syntax/parser.go
sosiska: Done in e336613e7883ab4a30afe31f6d84c28752fa2f75.

PR #378, opened 2019-04-08
Preserve script parameters when invoking set.

bilus:
The code is far from pretty but I want to make a minimal change since you said you're working on it on v3. This is a blocker for us working on https://oya.sh

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -66,7 +66,13 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 		r.setErr(ShellExitStatus(r.exit))
 		return 0 // the command's exit status does not matter
 	case "set":
-		if err := Params(args...)(r); err != nil {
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks for the PR. Is there an open or closed issue about this bug? It doesn't seem to me like the same issue I fixed in v3.

mvdan:
Hmm, this is a bug in both v3 (now master) and v2. I think the underlying issue is in Params, so a nicer fix might be possible.

mvdan:
I've found a nicer fix, which will work when using `Params` as an option directly too. I've also added more test cases. I'll backport into the v2 branch shortly; you can pull the latest commit from that without the need for me to tag `v2.6.5`.

bilus:
Thanks, perfect!

mvdan:
The backport is ee3cf905f3a4dd7a6fbfb5899f21304c64a92233.

PR #388, opened 2019-06-12
Support color output for the added/deleted lines in diff mode

mingrammer:
This PR supports color output for the added/deleted lines in diff mode.

Current:

Screen Shot 2019-06-13 at 12 07 31 AM

When PR is merged:

Screen Shot 2019-06-13 at 12 07 09 AM

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"regexp"
 
+	"golang.org/x/crypto/ssh/terminal"
 	"mvdan.cc/sh/v3/fileutil"
 	"mvdan.cc/sh/v3/syntax"
 )
@@ -26,7 +27,6 @@ var (
 	simple = flag.Bool("s", false, "")
 	find   = flag.Bool("f", false, "")
 	diff   = flag.Bool("d", false, "")
-	color  = flag.Bool("c", false, "")
 
 	langStr = flag.String("ln", "", "")
 	posix   = flag.Bool("p", false, "")
@@ -49,12 +49,25 @@ var (
 	in  io.Reader = os.Stdin
 	out io.Writer = os.Stdout
 
+	color       bool
 	ansiFgRed   = "\u001b[31m"
 	ansiFgGreen = "\u001b[32m"
 	ansiReset   = "\u001b[0m"
 
 	version = "v3.0.0-alpha1"
 )
+
+func init() {
+	if term, ok := os.LookupEnv("TERM"); ok && term == "dumb" {
+		color = false
+		return
+	}
+	if f, ok := out.(*os.File); ok && terminal.IsTerminal(int(f.Fd())) {
+		color = true
+		return
+	}
+	color = false
+}
 
 func main() {
 	flag.Usage = func() {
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks for the PR. However, I disagree with always using colored output. Tools like `diff` will only do so when stdout is a terminal, for example. Printing color codes always isn't a good idea; for example, some people avoid them via `TERM=dumb`.

I'll take a look at an alternative solution.

mingrammer:
Thank you for feedback. So how about using [fatih/color](https://github.com/fatih/color)? This package works well even if `TERM=dumb` set. (If set, it does not add the color codes around the text) See [this](https://github.com/fatih/color/blob/master/color.go#L20)

As you see, the author of this library has stopped maintaining, but it is already "done". (no need further maintenance)

mvdan:
I'm still not convinced. That would be an extra dependency, which itself also pulls in another two dependencies, all just so we could have a bit of logic to color a few lines.

But I do see the use of colored diffs. I'm ok with merging something like this, as long as colors are only enabled when:

* `TERM` is not `dumb`
* `out` is an `*os.File`, and `IsTerminal` returns true

I'll also leave a code review shortly.

cmd/shfmt/main.go
```diff
@@ -322,16 +326,34 @@ func diffBytes(b1, b2 []byte, path string) ([]byte, error) {
 		// as diff will return non-zero if the files differ.
 		return nil, err
 	}
+
+	output := make([][]byte, 0)
```
mvdan: use a `bytes.Buffer` instead, and then the functions can use `fmt.Fprint(w, ...)`. this also means we won't need to use `bytes.Join`.

cmd/shfmt/main.go
```diff
@@ -48,6 +48,10 @@ var (
 	in  io.Reader = os.Stdin
 	out io.Writer = os.Stdout
 
+	ansiFgRed   = []byte("\u001b[31m")
+	ansiFgGreen = []byte("\u001b[32m")
+	ansiReset   = []byte("\u001b[0m")
```
mvdan: once you switch to a buffer, these can be simply strings

cmd/shfmt/main_test.go
```diff
@@ -60,13 +61,13 @@ func TestStdin(t *testing.T) {
...
 @@ -1,3 +1,2 @@
-- foo
-+foo
+%s
+%s
  bar
--
-`
+%s
+`, markDeleted([]byte("- foo")), markAdded([]byte("+foo")), markDeleted([]byte("-")))
```
mvdan: Now that we'll have diff with and without colors, you could make `color` a global bool like `diff`, and have two subtests:

`Diff`, which is the old test as-is
`DiffColored`, which sets `color = true`, and ensures the escape codes are present.

Also, don't use `fmt.Sprintf` for the expected output. Just write the entire string as a single literal, escape codes and all.

mingrammer:
I fixed codes by reflecting the reviews.

cmd/shfmt/main.go
```diff
@@ -26,6 +26,7 @@ var (
 	simple = flag.Bool("s", false, "")
 	find   = flag.Bool("f", false, "")
 	diff   = flag.Bool("d", false, "")
+	color  = flag.Bool("c", false, "")
```
mvdan: sorry, but no flag - this should be on by default on a terminal, and it can be turned off with TERM=dumb

mingrammer: Oh, sorry I missed that comment. I fixed it.

PR #399, opened 2019-08-16
Add modd to README.md

cortesi:
Modd is a flexible developer tool that runs processes and responds to filesystem changes, which embeds sh as its built-in shell.

Many thanks for sh - embedding a shell in modd has made the cross-platform story for the tool much nicer.


cortesi:
Sure. That better?

PR #409, opened 2019-09-06
Fix home dir (~) expansion on Windows

andreynering:
This worked on Windows, but only on environments, like Git Bash, that sets the $HOME environment variable to improve compatibility with Unix tools.

While using Commnand Prompt or PowerShell, the real user path is available at $USERPROFILE or $HOMEDRIVE + $HOMEPATH.

See homedir implementation for reference:

https://github.com/mitchellh/go-homedir/blob/af06845cf3004701891bf4fdb884bfe4920b3727/homedir.go#L148

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/user"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -624,11 +623,11 @@ func (cfg *Config) expandUser(field string) (prefix, rest string) {
 		}
 	}
 
-	u, err := user.Lookup(name)
+	home, err := os.UserHomeDir()
 	if err != nil {
 		return "", field
 	}
-	return u.HomeDir, rest
+	return home, rest
 }
 
 func findAllIndex(pattern, name string, n int) [][]int {
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -8,14 +8,13 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/user"
 	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
 
 	"mvdan.cc/sh/v3/syntax"
-
-	"github.com/mitchellh/go-homedir"
 )
 
 // A Config specifies details about how shell expansion should be performed. The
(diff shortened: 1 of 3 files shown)


mvdan:
All the CI environments are broken?

Also, please don't add an external dependency for this. Use https://golang.org/pkg/os/#UserHomeDir; master only supports 1.12 and up now.

andreynering:
Interesting, all tests passes for me locally on macOS.

Perhaps it's a Linux specific issue?

mvdan:
Mac is failing there too, so I doubt it.

mvdan:
Also, I haven't updated CI to use 1.12 and up yet, but I'll do that today or tomorrow :)

andreynering:
@mvdan homedir is a pretty small library that doesn't use C, unlike `os/user`. Anyway, updated to use the standard library.

mvdan:
Note that I mentioned `os.UserHomeDir`, which has no cgo at all. It basically does the better version of what my code did here.

I suspect that the best mechanism would be to check $HOME first, and UserHomeDir as a fallback. This way, platforms like Windows would try both.

mvdan:
I also wonder why `user.Lookup` didn't work for your scenario. Were you building a binary without CGo?

andreynering:
> I also wonder why `user.Lookup` didn't work for your scenario. Were you building a binary without CGo?

Note that the real cause of this bug is the fact that the function returned if `name == ""`, even if `$HOME` was blank. The fallback code wasn't running in this scenario.

andreynering:
And thanks, I didn't know `os.UserHomeDir` was a thing.

mvdan:
Ah, sorry, I've been misreading this patch. The expand package isn't buggy per se. If anything, its call to `os/user` should be removed entirely. It's up to the caller to ensure that `$HOME` is set up properly.

And this is what the `interp` package does. In `Runner.Reset`, if `$HOME` isn't set, it uses `os/user.Current().HomeDir`. That's what should be swapped with `os.UserHomeDir`, which would make it work in cases where CGo is not enabled.

But again, this should all work on plain Windows, unless you were building with `CGO_ENABLED=0`. I'm all for making it work without CGo, though.

mvdan:
(In the future, filing a bug might be the best way to discuss subtle bugs and fixes like these; otherwise, with a patch I have to guess what you're trying to fix and why)

andreynering:
For some reason, this bug is happening even without `CGO_ENABLED=0`. There might be something else...

mvdan:
We should be discussing all of this on an issue, with clear reproduction steps :) Then we could also write a regression test that actually ensures we fix the bug for good.

andreynering:
Sorry, I was just trying to have a bug fixed here 🙂 

Opening an issue in a minute...

PR #415, opened 2019-09-21
interp,syntax: use named type for options

kaey:
This makes option funcs group nicely under their type in godoc.

I think binary options should also take `v bool` argument for consistency. I didn't use `syntax` package and can't make an informed decision whether it's needed. Can implement in another PR.


syntax/printer.go
```diff
@@ -39,29 +49,39 @@ func SpaceRedirects(p *Printer) { p.spaceRedirects = true }
...
 		tabWriter: new(tabwriter.Writer),
 	}
-	for _, opt := range options {
+	for _, opt := range opts {
 		opt(p)
 	}
 	return p
 }
 
+func (p *Printer) SetOption(opts ...PrinterOption) {
```
mvdan: not a big fan of `SetOption`; the option type is already pretty trivial to apply. why the extra API?

syntax/printer.go
```diff
@@ -39,29 +49,39 @@ func SpaceRedirects(p *Printer) { p.spaceRedirects = true }
 // Note that this feature is best-effort and will only keep the
 // alignment stable, so it may need some human help the first time it is
 // run.
-func KeepPadding(p *Printer) {
-	p.keepPadding = true
-	p.cols.Writer = p.bufWriter.(*bufio.Writer)
-	p.bufWriter = &p.cols
+func KeepPadding() PrinterOption {
```
mvdan: Not convinced that it's worth swapping these to return a function. I assume this is for consistency and to make the godoc prettier?

Did you check if godoc groups all PrinterOptions, even if some of them are like `func KeepPadding(p *Printer) {`? If it doesn't, then I'd argue that's something to improve there. I don't see a problem per se in omitting parameters and a function return value when they're not needed.

This would also mean churn for the existing `syntax` and `interp` users, which further makes me think this is probably not something we want to do.

syntax/printer.go
kaey: Implicit docs and clear api for users.
Without this you would need to add explicit example on how to apply option to existing printer. And also answer your newly hired juniors why syntax for applying opts to new printer and existing differ.

```
[kaey@kpc sh]$ go doc ./syntax printer
package syntax // import "mvdan.cc/sh/v3/syntax"

type Printer struct {
        // Has unexported fields.
}
    Printer holds the internal state of the printing mechanism of a program.

func NewPrinter(opts ...PrinterOption) *Printer
func (p *Printer) Print(w io.Writer, node Node) error
func (p *Printer) SetOption(opts ...PrinterOption)


[kaey@kpc sh]$ go doc ./syntax printeroption
package syntax // import "mvdan.cc/sh/v3/syntax"

type PrinterOption func(*Printer)
    PrinterOption is a function which can be passed to NewPrinter or
    Printer.SetOption() to alter its behaviour.

func BinaryNextLine() PrinterOption
func Indent(spaces uint) PrinterOption
func KeepPadding() PrinterOption
func Minify() PrinterOption
func SpaceRedirects() PrinterOption
func SwitchCaseIndent() PrinterOption
```

mvdan: Gotcha. I think an example would be fine. With the addition of SetOption, there are now *three* ways to set up options for a new printer, and that's definitely bad design. One alternative would be to not accept any options in New, but in my opinion that's worse.

> And also answer your newly hired juniors why syntax for applying opts to new printer and existing differ.

That's just how it is if we want to allow New to take options. Juniors would most likely not be applying options to existing printers, so I dont' think this point is important. As said before, we can add an example if needed.

syntax/printer.go
kaey: >I assume this is for consistency and to make the godoc prettier?

Yes

>If it doesn't, then I'd argue that's something to improve there.

1. As I mentioned in issue, there is nothing godoc can do to group it without explicit types.
2. As I mentioned in commit message, these funcs should take bool argument
```
func BinaryNextLine(v bool) PrinterOption
func Indent(spaces uint) PrinterOption
func KeepPadding(v bool) PrinterOption
func Minify(v bool) PrinterOption
func SpaceRedirects(v bool) PrinterOption
func SwitchCaseIndent(v bool) PrinterOption
```
This way they become togglable. While I don't know if it's any useful for Printer, it definitely can be useful for Runner (one can imagine altering runner config with pseudocomands, like :set opt=foo when it's running interactively).

>This would also mean churn for the existing syntax and interp users, which further makes me think this is probably not something we want to do.

Since v3 is not stable yet why not take chance to improve docs and ease onboarding for all new users? Churn is unfortunate, yes, but this specific case can be quickly fixed with `sed s/syntax.BinaryNextLine/syntax.BinaryNextLine()/g`

One more example why I don't like top-level functions used as values:
take a look this popular package https://godoc.org/github.com/pkg/profile
I'm pretty sure it's impossible to quickly understand how to use it without taking a look at examples and I think this is a big enough problem. Explicit typing creates implicit documentation which needs zero maintenance. This is always a good thing.

syntax/printer.go
kaey: Third way is really hard to discover and there is no need for users to use it.
Also there is another design where it's not possible at all - move config into separate unexported struct and define option like this `type PrinterOption func(*printerConfig)`. But I don't see the need to do this here.

>One alternative would be to not accept any options in New

This would make the most common usage more clunky, I don't see any need to do this.

mvdan: > But I don't see the need to do this here.

> This would make the most common usage more clunky,

We agree on those two points, at least :)

I still don't think SetOption is necessary. I see the point about discoverability on godoc, but we can accomplish that with a bit of docs and an example too. An extra API, particularly when it doesn't enable any new features, seems like a bad option to me. Please just revert; that change isn't necessary to use named type options.

syntax/printer.go
mvdan: > This way they become togglable.

That's a good point, actually. I didn't think about disabling the options initially, because most of the time they're applied on a new object, where they default to false.

Please add the bool arguments and LGTM. Don't call them `v bool` though; perhaps `enabled bool`.

syntax/printer.go
kaey: Done

syntax/printer.go
kaey: Done

syntax/printer.go
```diff
@@ -39,24 +50,37 @@ func SpaceRedirects(p *Printer) { p.spaceRedirects = true }
...
+func NewPrinter(opts ...PrinterOption) *Printer {
+	bw := bufio.NewWriter(nil)
+
 	p := &Printer{
-		bufWriter: bufio.NewWriter(nil),
+		bufWriter: bw,
 		tabWriter: new(tabwriter.Writer),
+		cols: colCounter{
+			Writer: bw,
+		},
```
mvdan: why this change? it's not mentioned at all in the commit message, and I don't see why it's necessary

kaey: This is to make KeepPadding option togglable, so that user can call it with false or true in any order any number of times.

kaey: Previously setting KeepPadding multiple times would result in a panic.

mvdan: If that's the case, please file a separate issue for that, and let's fix that in a separate PR with a test.

kaey: It's also possible to remove writer swapping logic and always write to colCounter irregardless of KeepPadding setting. This would simplify code slightly, but also slightly regress performance.

mvdan: I think we should discuss that in a separate issue+PR; this thread and patch are big enough as it is.

kaey: I've changed this back and added TODO comment.

PR #422, opened 2019-09-26
cmd/shfmt: add ignore file option

### LINKED ISSUE #288: cmd/shfmt: add a way to skip custom directories when walking directories
When recursing it would be nice to be able to ignore certain directories or files (for example, `node_modules`).

Either a `--exclude=node_modules`, or `--ignore=node_modules`, or a `.shfmtignore`-type functionality would be nice.

soullivaneuh:
Fixes #288

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 
 	"github.com/pkg/diff"
+	ignore "github.com/sabhiram/go-git-ignore"
 	"golang.org/x/crypto/ssh/terminal"
 
 	"mvdan.cc/sh/v3/fileutil"
@@ -30,8 +31,9 @@ var (
 	find    = flag.Bool("f", false, "")
 	diffOut = flag.Bool("d", false, "")
 
-	langStr = flag.String("ln", "", "")
-	posix   = flag.Bool("p", false, "")
+	langStr    = flag.String("ln", "", "")
+	posix      = flag.Bool("p", false, "")
+	ignoreFile = flag.String("if", ".shfmtignore", "")
 
 	indent      = flag.Uint("i", 0, "")
 	binNext     = flag.Bool("bn", false, "")
@@ -78,6 +80,7 @@ Parser options:
 
   -ln str   language variant to parse (bash/posix/mksh, default "bash")
   -p        shorthand for -ln=posix
+  -if       specify path of ignore file (default ".shfmtignore")
 
 Printer options:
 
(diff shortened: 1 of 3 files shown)


soullivaneuh:
The failing test does not seem to be related: https://github.com/mvdan/sh/pull/422/checks?check_run_id=236977427#step:4:408

cmd/shfmt/main.go
```diff
@@ -30,8 +31,9 @@ var (
 	find    = flag.Bool("f", false, "")
 	diffOut = flag.Bool("d", false, "")
 
-	langStr = flag.String("ln", "", "")
-	posix   = flag.Bool("p", false, "")
+	langStr    = flag.String("ln", "", "")
+	posix      = flag.Bool("p", false, "")
+	ignoreFile = flag.String("if", ".shfmtignore", "")
```
mvdan: The default doesn't make a lot of sense to me; the point of gitignore files is generally that they work in nested directories too.

Also, why is a flag necessary? If the file exists, the tool should use it.

soullivaneuh: The flag allow use to specify an another ignore file model such as `.gitignore`:

```
shfmt --ignore-path .gitignore
```

Ref: https://github.com/mvdan/sh/issues/288#issuecomment-535381642

It also solve what you said in: https://github.com/mvdan/sh/issues/288#issuecomment-535918319

> it adds a custom -if flag and a custom .shfmtignore file on disk.

It add a custom flag, but not necessary a new ignore file, depending of your choice.

This is the more flexible way and it is working great. But I can also make it less restrictive by just using `.gitignore` directly and without flag. But it will remove flexibility and will not work with other vcs.

I recommend you to take a look to the other lint tools, many of them use this method. This is quite handful, especially for CI.

soullivaneuh: > the point of gitignore files is generally that they work in nested directories too.

You may also have `.gitignore` files on nested directory. Also, I set up the code like this because shfmt can have multiple path arguments:

```
shfmt /path/to/project1 /path/to/project2
```

So the flag is like a "model" file to use.

fpuga: I also like this approach as it seems to be the approach that are taking other formatters like stylelint, prettier, ...

For the commit maybe the flag could be `-ip` instead of `-if` to try to mimic `--ignore-path` that seems to be the more common agreement.

soullivaneuh: I was not sure about the name. I can change it but I would have @mvdan final review for this first.

mvdan:
I fixed #288 with `.editorconfig`, as I laid out in that thread two months ago. If you disagree with that, or want to propose an alternative, please do so in that thread. I'm closing this PR for now, because I think the drawbacks of this patch (adding a flag and a tool-specific dot-file) aren't worth it.

PR #430, opened 2019-10-08
cmd/gosh: remove usage of global variables

kaey:
Func interactive() depends on runner.Std* which will be unexported
in a follow-up PR. To remove that dependency we pass runner and
std* directly to functions instead of using global vars.

### DIFF
--- a/cmd/gosh/main.go
+++ b/cmd/gosh/main.go
@@ -17,82 +17,82 @@ import (
 	"mvdan.cc/sh/v3/syntax"
 )
 
-var (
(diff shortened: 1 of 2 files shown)


kaey:
fuzz failure looks unrelated
```
Please set env variable FUZZIT_API_KEY or pass --api-key. API Key for you account: https://app.fuzzit.dev/settings
```

mvdan:
Agreed that the failure is unrelated. Looks like some issue with GitHub Actions as I haven't touched it.

PR #431, opened 2019-10-08
interp: use simple funcs instead of middlewares in Open/ExecModule

### LINKED ISSUE #419: interp: move OpenDevImpls code to DefaultOpen
OpenDevImpls name is confusing with regards to what it does,
`>/dev/null` is very common, making it work on windows by default is a +.
If user would want to implement /dev/null themselves, they would need to either handle it before calling DefaultOpen or copy-paste DefaultOpen, it's small anyway.

### LINKED ISSUE #421: interp: don't accept multiple mods in With*Modules
Current example suggests using WithExecModules like this https://github.com/mvdan/sh/blob/master/interp/example_test.go#L38
There are a few issues with it
1. Much glue code - when you want to implement 10 commands, you make a constructor and it looks like this
```
		gocmd := func(cmd string, f func(mc interp.ModuleCtx, args []string) error) func(next interp.ExecModule) interp.ExecModule {
			return func(next interp.ExecModule) interp.ExecModule {
				return func(ctx context.Context, args []string) error {
					if args[0] == cmd {
						mc, _ := interp.FromModuleContext(ctx)
						if err := f(mc, args); err != nil {
							return err
						}
					}
					return next(ctx, args)
				}
			}
		}
```
And also makes panic traces worse (lose func name) unless you implement it as package-level function.

2. Weird stack traces - say you add 2 commands - `join`, `foo`, DefaultExec, stack-trace shows like `join` calls `foo` which calls `DefaultExec`

3. DefaultExec has different func signature which means it requires wrapping. Currently it's actually always called as the last module, so I don't know why export it at all,
but see https://github.com/mvdan/sh/issues/420

I propose to change With*Module to just accept single *Module, this actually makes code simpler and fixes stack traces.
```
func exec(ctx context.Context, args []string) error {
	if logTimings {
		start := time.Now()
		defer func() {
			log.Printf("command %v finished in %v", args, time.Since(start))
		}()
	}

	mc, _ := interp.FromModuleContext(ctx)
	switch args[0] {
	case "join":
		return gojoin(ctx, mc, args)
	case "foo:
		return gofoo(ctx, mc, args)
	}

	return DefaultExec(ctx, args)
}
```
This is not HTTP, I don't think there are usecases for tons of middlewares (they're still possible, user will have to write wrapping themselves)

kaey:
Middleware style api is overcomplicated for users and require a lot of glue code,
makes path handling unclear on windows and produces weird stack traces.
Switch to simple functions for now. If user wants to compose their handlers,
they can implement whatever interface is comfortable to use for them themselves.

Renames:
	WithExecModules -> ExecModule
	WithOpenModules -> OpenModule
	OpenModule -> OpenModuleFunc
	ExecModule -> ExecModuleFunc

Removed OpenDevImpls - it's unclear what signature it should have.
It can be readded later, if needed.

Switched tests and examples to new api.

Fixes #421
Fixes #419

### DIFF
--- a/interp/module.go
+++ b/interp/module.go
@@ -50,14 +50,14 @@ type ModuleCtx struct {
 	KillTimeout time.Duration
 }
 
-// ExecModule is the module responsible for executing a simple command. It is
+// ExecModuleFunc is the module responsible for executing a simple command. It is
 // executed for all CallExpr nodes where the first argument is neither a
 // declared function nor a builtin.
 //
 // Use a return error of type ExitStatus to set the exit status. A nil error has
 // the same effect as ExitStatus(0). If the error is of any other type, the
 // interpreter will come to a stop.
-type ExecModule func(ctx context.Context, args []string) error
+type ExecModuleFunc func(ctx context.Context, args []string) error
 
 func DefaultExec(ctx context.Context, args []string) error {
 	mc, _ := FromModuleContext(ctx)
@@ -256,7 +256,7 @@ func pathExts(env expand.Environ) []string {
 	return exts
 }
 
-// OpenModule is the module responsible for opening a file. It is
+// OpenModuleFunc is the module responsible for opening a file. It is
 // executed for all files that are opened directly by the shell, such as
 // in redirects. Files opened by executed programs are not included.
 //
@@ -266,7 +266,7 @@ func pathExts(env expand.Environ) []string {
 // Use a return error of type *os.PathError to have the error printed to
 // stderr and the exit status set to 1. If the error is of any other type, the
 // interpreter will come to a stop.
-type OpenModule func(ctx context.Context, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error)
+type OpenModuleFunc func(ctx context.Context, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error)
 
 func DefaultOpen(ctx context.Context, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
 	mc, _ := FromModuleContext(ctx)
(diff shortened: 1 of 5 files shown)


interp/module.go
```diff
@@ -275,21 +275,3 @@ func DefaultOpen(ctx context.Context, path string, flag int, perm os.FileMode) (
 	}
 	return os.OpenFile(path, flag, perm)
 }
-
-func OpenDevImpls(next OpenModule) OpenModule {
```
andreynering: Hey guys, I have some feedback regarding this change. Please let me know of your opinions.

Is there any particular reason why this function was removed? While it's easy to copy it to my projects, I think it's useful to have it here so new users can find and use it. This is a common use case IMO. I noticed that you even had to copy it to the tests files to keep it green.

If you decide to proceed with the removal, it would be nice to at least mention it in the CHANGELOG. IIRC this was here since v2.x.

/cc @kaey @mvdan

mvdan: Have you seen #419? It contains all the context, I think.

I don't have a strong opinion here, to be honest. I initially added it with the idea that more devices would be added with time, such as `/dev/zero`, `/dev/urandom`, `/dev/stdin`, and so on.

If people would find the above useful, I think it's worthwhile keeping or re-adding the feature. But as it was, barely above six lines of code, I don't think there's a strong argument for keeping it there.

mvdan: I do agree that this belongs in the changelog if we decide to remove it, though.

andreynering: Got it.

Another suggestion if you proceed with the removal: along with the CHANGELOG mention, an example on the documentation or README would be helpful. I think that people that needs Windows support will want to know how to support `/dev/null`.

kaey: This function cannot work generally well for all cases due to differences in windows path handling (forward slash/backward slash/mixed).
For simple cases you can get away with 
```
if path == "/dev/null" && runtime.GOOS == "windows" {
    path = "NUL"
}
```

No need for builtin devnull implementation.

mvdan: I agree with @kaey; let's leave this out for now. I'll amend the changelog.

PR #441, opened 2019-10-17
interp: add a simple Interactive method

mvdan:
Mimicking Parser.Interactive, this method implements the basics of a
full interactive shell. That's useful for many user-facing use cases,
and also as a debugging tool.

The function has zero options because the list of options it could take
is never-ending. We could start with PS1 and PS2, but we could also add
PS3, PS4, a context.Context, Parser options, et cetera.

Instead of tripling the amount of code and adding lots to the API,
suggest that advanced users simply copy the twenty lines to add their
own options and logic.

### DIFF
--- a/cmd/gosh/main.go
+++ b/cmd/gosh/main.go
@@ -42,7 +42,7 @@ func runAll() error {
 	}
 	if flag.NArg() == 0 {
 		if terminal.IsTerminal(int(os.Stdin.Fd())) {
-			return runInteractive(r, os.Stdin, os.Stdout, os.Stderr)
+			return r.Interactive()
 		}
 		return run(r, os.Stdin, "")
 	}
--- a/cmd/gosh/main_test.go
+++ b/cmd/gosh/main_test.go
@@ -165,6 +165,8 @@ var interactiveTests = []struct {
 	},
 }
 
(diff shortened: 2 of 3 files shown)


mvdan:
@kaey thoughts?

mvdan:
I'll be removing the copy in `cmd/gofmt` and moving the tests, if this API change seems sensible. Multiple people have asked to be able to create an interactive shell easily in plain Go, and this method is non-trivial to implement right.

kaey:
Seems like `shell` pkg is a better place for it.

mvdan:
Hm, maybe - but that forces us to repeat the stdin/stdout/stderr parameters.

kaey:
funcs in shell do not accept `Runner` anyway so you can define it like this
```
func Interactive() error {
    return runInteractive(os.Stdin, os.Stdout, os.Stderr)
}

func runInteractive(blabla...) error {
    r, err := interp.New(blabla...)

    everything else
}
```

mvdan:
You're right. I'll do that :)

mvdan:
I'm leaving this for v3.1.

mvdan:
If/when I get back to this, I'll just open a new PR.

PR #465, opened 2019-12-29
Add support for pre-commit

ArturKlauser:
[pre-commit](https://pre-commit.com/) is infrastructure to easily set up running any set of tools as git pre-commits on a repo. A tool only needs to provide a hook description (.pre-commit-hook.yaml) in its repo to become usable by the pre-commit infrastructure.

Here I'm adding this hook description, which makes shfmt easily usable by pre-commit (see added example in README.md). I'm also turning pre-commit loose on this repo and configure it to run automatically on the Github Actions CI pipeline on every push and pull request.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+# See https://pre-commit.com for more information
+-   id: shfmt
+    name: shfmt shell scripts
+    description: Format shell scripts with shfmt running in Docker
+    language: docker_image
+    entry: mvdan/shfmt:v3.0.0
+    types: [shell]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+on: [create]
+
+name: Check release tag consistency
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1  # no history, just HEAD
+    - name: Check pre-commit hook version tag
+      run: |
+        is_tag="$(echo ${GITHUB_REF} | grep -c '^refs/tags/')" || true
+        if [ "$is_tag" != '0' ]; then
+          # A tag was created - let's check.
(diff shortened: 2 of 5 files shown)


mvdan:
I'm conflicted about this PR. I see the value in pre-commit hooks, but this adds nearly a hundred lines of code for me to maintain. I'm not sure I see the point of all this machinery, when in reality running `shfmt` as a pre-commit hook is pretty trivial.

In the past, I've welcomed "downstream" contributions like packaging shfmt for Linux distros, but I haven't been involved simply because there is too much work and I can't keep up with the 20+ ways to distribute software. I simply publish git tags with binaries, and a Docker image, since that last bit is very little work for me and does seem like a very popular request.

I think pre-commit falls under the kind of downstream work that I'd prefer others to maintain. If someone must maintain these files, it should be possible for it to happen in a separate git repository. We can then link that from https://github.com/mvdan/sh#related-projects.

ArturKlauser:
Up to you. If you're not interested feel free to close the PR.

mvdan:
Thanks for understanding. I'll close this for now, but I'm still happy to add links to other repos or projects in the README.

asottile:
The only necessary file is `.pre-commit-hooks.yaml` -- would you be more amenable to a PR with just that file?

Not sure why the other stuff was proposed here :S

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+# See https://pre-commit.com for more information
+-   id: shfmt
+    name: shfmt shell scripts
+    description: Format shell scripts with shfmt running in Docker
+    language: docker_image
+    entry: mvdan/shfmt:v3.0.0
```
asottile: this should be `language: go` and `entry: shfmt`

asottile:
(with my recommendations and only shipping `.pre-commit-hooks.yaml` there should be zero maintenance as the config format is stable)

mvdan:
I mean, sure, I can add a short yaml file with some simple metadata. I'm not sure why that's necessary or useful, though, or why it couldn't be in another repo.

PR #467, opened 2019-12-30
Minimize Docker image

ArturKlauser:
This is a proposal for a minimized Docker image that only contains the shfmt binary. It reduces the size a little by not packaging up busybox. Up to you if you want that.
```
$ docker images
REPOSITORY     TAG         IMAGE ID            CREATED             SIZE
shfmt          latest      fbf022530b4b        33 minutes ago      3.02MB
shfmt          3.0.0       4c876e317d99        38 minutes ago      4.48MB
```
Also added more usage instructions to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,14 @@ To build a Docker image, checkout a specific version of the repository and run:
 
 	docker build -t my:tag -f cmd/shfmt/Dockerfile .
 
+To use the Docker image, run:
+
+	docker run --rm -v $PWD:/mnt -w /mnt my:tag <shfmt arguments>
+
+E.g.:
+
+	docker run --rm -v $PWD:/mnt -w /mnt shfmt:latest -i 2 -l .
+
 ### Related projects
 
 * Alternative docker images - by [jamesmstone][dockerized-jamesmstone], [PeterDaveHello][dockerized-peterdavehello]
--- a/cmd/shfmt/Dockerfile
+++ b/cmd/shfmt/Dockerfile
@@ -1,9 +1,14 @@
-FROM golang:1.13.5-alpine3.10
+FROM golang:1.13.5-alpine3.10 as build
 
 WORKDIR /src
 COPY . .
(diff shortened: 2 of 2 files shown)


mvdan:
We had briefly discussed this in #68; in particular, I laid out what I'd do in https://github.com/mvdan/sh/issues/68#issuecomment-566073825 a couple of weeks ago.

This PR as-is is probably not a good idea, because plenty of users need a Docker image with shfmt as well as a shell installed. For example, that's a requirement for Gitlab CI. A bit silly, I know, but that's how the world works.

If you want to have a try at the duplicate images (scratch and alpine), I'll be happy to review. If not, please open an issue, and I'll probably get to it sometime in January.

ArturKlauser:
I see that you're currently using Docker Hub autobuilds. Unfortunately, they are not very configurable, e.g. AFAIK you can't specify a --target for the build, which would be the handy way to build two images from a single docker file. Are you open to switch the docker image build and publish to a different CI? E.g. Github Actions, which you are already using, should be able to handle that.

mvdan:
> AFAIK you can't specify a --target for the build, which would be the handy way to build two images from a single docker file.

Ah, that's a neat trick. I was imagining that maintaining two Dockerfiles was the only way. I assume we'd still have scratch be the default target, or is there no way to do that?

> E.g. Github Actions, which you are already using, should be able to handle that.

Yes, that's fine. Really, the only reason I did it via Docker Hub is because I liked how it shows you the Dockerfile directly, as well as how it built each of the images. Is there a way to show the Dockerfile from builds pushed from our CI here?

If you are happy to do most of the work, I can do the bits I assume you won't be able to, like turning off docker hub auto-builds and adding the right secret to CI.

mvdan:
Do you want to focus on #468 and close this, since the changes here are included there? Or do you prefer keeping the changes as two commits?

ArturKlauser:
Doing everything in #468 SGTM.

PR #468, opened 2019-12-30
Add multi-architecture Docker image build

ArturKlauser:
This is a proposal related to #467. It builds two docker images, the busybox image as is currently built as well as a minimal image containing only the shfmt binary.

In addition, the docker images are built as multi-architecture images which is a recent feature supported by docker's buildx extension. See how my [test images](https://hub.docker.com/repository/docker/arturklauser/shfmt/tags?page=1) report availability for multiple architectures. Each client pulling from Docker Hub automatically pulls only the appropriate architecture version for itself, so there is no additional overhead for the client. Clients don't need experimental features enabled for this (only the builder need them).

The docker build and deployment is hooked into the existing Github Actions CI/CD pipeline, conditioned on all other builds and tests passing. You can see a [sample run](https://github.com/ArturKlauser/sh/commit/2e029ed023138eadd102a99d1766f9ce77b13c35/checks?check_suite_id=379078248) on my fork.
To activate this there are two requirements:
  * Add the DOCKER_PASSWORD [secret to Github](https://github.com/mvdan/sh/settings/secrets)
  * Deactivate the [autobuild on Docker Hub](https://hub.docker.com/repository/docker/mvdan/shfmt/builds/edit), as it would conflict with the Github CI/CD build.

The Github Action CI job is currently only hooked up to pushes on the master branch. To release the versioned docker images it'd also need to be hooked up to tag creations. I'd be happy to do that if you have any interest in this PR at all.


build_multi_arch_docker.sh
```diff
@@ -0,0 +1,97 @@
...
+#   (source ./build_multi_arch_docker.sh; build-multi-arch-docker::main)
+# Where the environment variables are:
+#   DOCKER_USER: Docker Hub user name to which to push images.
+#   DOCKER_PASSWORD: Docker Hub password for above user.
+#   DOCKER_BASE: Docker image base name (user/image).
+#   TAGS: A set of image tags to which to deploy. Appended to DOCKER_BASE.
+#   DOCKER_PLATFORMS: Docker buildx platforms (see `docker buildx ls`).
+# Example:
+#   DOCKER_USER=foo DOCKER_PASSWORD=bar DOCKER_BASE=foo/shfmt TAGS=latest \
+#   (source ./build_multi_arch_docker.sh; build-multi-arch-docker::main)
```
mvdan: Is this source+namespacing method some form of standard? It seems kinda awkward and complex to me.

`./script action` seems much simpler and cleaner to me than `(source script && script::action)`.

build_multi_arch_docker.sh
```diff
@@ -0,0 +1,97 @@
...
+
+function build_multi_arch_docker::install_docker_buildx() {
```
mvdan: This command makes heavy use of sudo, and assumes a Debian-based distro. I think it belongs in the CI script directly, not here. We don't want to encourage users to modify their system as part of a "build" script, even less so when it's part of the recommended "main" command.

build_multi_arch_docker.sh
```diff
@@ -0,0 +1,97 @@
...
+	# Instantiate docker buildx builder with multi-architecture support.
+	export DOCKER_CLI_EXPERIMENTAL=enabled
+	docker buildx create --name mybuilder
+	docker buildx use mybuilder
+	# Start up buildx and verify that all is OK.
+	docker buildx inspect --bootstrap
+}
+
+# Log in to Docker Hub for deployment.
+function build_multi_arch_docker::login_to_docker_hub() {
```
mvdan: similarly, this really just belongs in the CI script directly.

build_multi_arch_docker.sh
mvdan: Also, did you check if the Docker version in CI is too old? Because it might not be.

.github/workflows/test.yml
```diff
@@ -50,3 +50,49 @@ jobs:
       if: github.event_name == 'push' && github.ref == 'refs/heads/master'
       env:
         FUZZIT_API_KEY: ${{ secrets.FUZZIT_API_KEY }}
+
+  docker:
```
mvdan: I prefer this to be part of a `deploy.yaml` workflow, as it's not really related to test, and only runs on pushes.

cmd/shfmt/Dockerfile
```diff
@@ -1,9 +1,18 @@
-FROM golang:1.13.5-alpine3.10
+FROM golang:1.13.5-alpine3.10 as build
 
 WORKDIR /src
 COPY . .
 RUN CGO_ENABLED=0 go build -ldflags '-w -s -extldflags "-static"' ./cmd/shfmt
 
-FROM busybox:1.31.1-musl
-COPY --from=0 /src/shfmt /bin/shfmt
+FROM busybox:1.31.1-musl AS busybox
```
mvdan: I think if we start duplicating images, we might as well make this a proper distro - alpine linux. Just pick the latest tag they've published, and fix all the names elsewhere.

cmd/shfmt/Dockerfile
```diff
@@ -1,9 +1,18 @@
...
 ENTRYPOINT ["/bin/shfmt"]
+
+FROM scratch
+COPY --from=build /src/shfmt /bin/shfmt
+ENTRYPOINT ["/bin/shfmt"]
+
+# Usage: 
+#  docker run --rm -v $PWD:/mnt -w /mnt mvdan/shfmt:<version> <shfmt arguments>
+# E.g.:
+#  docker run --rm -v $PWD:/mnt -w /mnt mvdan/shfmt:latest -i 2 -l .
```
mvdan: This is in the README, so I'd delete the comment

build_multi_arch_docker.sh
```diff
@@ -0,0 +1,97 @@
...
+		build_multi_arch_docker::buildx -t "$DOCKER_BASE-busybox:$tag" \
+			--target busybox
+	done
+}
+
+# Test all pushed docker images on all platforms.
+function build_multi_arch_docker::test_all() {
+	printf '%s\n' "#!/bin/sh" "echo 'hello world'" >myscript
+
+	for platform in ${DOCKER_PLATFORMS//,/ }; do
```
mvdan: why go through the trouble of using commas as separators? spaces would be much simpler :)

README.md
```diff
@@ -126,6 +126,18 @@ To build a Docker image, checkout a specific version of the repository and run:
 
 	docker build -t my:tag -f cmd/shfmt/Dockerfile .
 
+Alternatively, if you want an image that includes busybox, run:
```
mvdan: > Alternatively, if you want an image that includes busybox, add `--target busybox`.

README.md
```diff
@@ -126,6 +126,18 @@ To build a Docker image, checkout a specific version of the repository and run:
...
+
+	docker build -t my:tag -f cmd/shfmt/Dockerfile --target busybox .
+
+To use the Docker image, run:
+
+	docker run --rm -v $PWD:/mnt -w /mnt my:tag <shfmt arguments>
+
+E.g.:
+
+	docker run --rm -v $PWD:/mnt -w /mnt shfmt:latest -i 2 -l .
```
mvdan: I think you can remove this second example. I don't want the README to be too long, and I think it's clear enough.

build_multi_arch_docker.sh
ArturKlauser: It sounds like you'd prefer everything that is in build_multi_arch_docker.sh inside the ci.yml file? That's fine with me.

.github/workflows/test.yml
ArturKlauser: One hitch with this approach in Github Actions though: AFAIK jobs in one foo.yml can'd depend on jobs in another bar.yml, so the deployment couldn't be conditional on the tests passing.

build_multi_arch_docker.sh
ArturKlauser: The `docker buildx build --platform` need commas, so we have to change it from one to the other somewhere either way.

build_multi_arch_docker.sh
mvdan: Yeah. The only thing that's really directly useful to users is the "build" function, but that seems like a one-liner anyway.

.github/workflows/test.yml
mvdan: Ah, that's a good point. SGTM

build_multi_arch_docker.sh
mvdan: Ah, gotcha.

build_multi_arch_docker.sh
ArturKlauser: > Also, did you check if the Docker version in CI is too old? Because it might not be.

AFAICT the installed version of docker does not support buildx, even when experimental mode is turned on.
It is apparently from a different package, not docker-ce:
```
dpkg -S /usr/bin/docker
moby-cli: /usr/bin/docker
```
and that doesn't cut it:
```
export DOCKER_CLI_EXPERIMENTAL=enabled
docker buildx
docker: 'buildx' is not a docker command.
```

mvdan: Gotcha, thanks for checking

build_multi_arch_docker.sh
ArturKlauser: build_multi_arch_docker.sh is now folded into test.yml

.github/workflows/test.yml
ArturKlauser: I took you SGTM as meaning you're OK with having it in test.yml, conditioned on the tests passing. If I misunderstood let me know.

cmd/shfmt/Dockerfile
ArturKlauser: done

README.md
ArturKlauser: done

README.md
ArturKlauser: done

ArturKlauser:
Ready for review again.
Please trigger a rerun of the CI pipeline. MacOS has failed (as it occasionally does).

.github/workflows/test.yml
```diff
@@ -50,3 +50,117 @@ jobs:
...
+    # Only deploy if previous stages pass.
+    needs: [test, test-linux-alpine, fuzz]
+    runs-on: ubuntu-latest
+    if: >
+      github.event_name == 'push' && 
+        (github.ref == 'refs/heads/master' ||
+         startsWith(github.ref, 'refs/tags/'))
+    env:
+      # Export environment variables for all stages.
+      DOCKER_CLI_EXPERIMENTAL: enabled  # for 'docker buildx'
```
mvdan: nit: please use single spaces for comments. as far as I know, the double-space format is common in python, but not particularly in yaml or Go-related stuff

.github/workflows/test.yml
```diff
@@ -50,3 +50,117 @@ jobs:
...
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1  # no history, just HEAD
+    - name: Set up image tags
+      run: |
+        set -vx
+        # Export environment variable for later stages.
+        if echo "$GITHUB_REF" | grep -q '^refs/heads/'; then
+          # Pushes to (master) branch - deploy 'latest'.
+          echo "::set-env name=TAGS::latest"
```
mvdan: isn't TAGS always just one word? it's probably easiest to make it singular, and remove the loops further below. if we really need multiple tags in the future, we can re-add the looping.

.github/workflows/test.yml
```diff
@@ -50,3 +50,117 @@ jobs:
...
+            -f cmd/shfmt/Dockerfile \
+            "$@" \
+            .
+        }
+
+        for tag in $TAGS; do
+          buildx -t "$DOCKER_BASE:$tag"
+          buildx -t "$DOCKER_BASE-alpine:$tag" --target alpine
+        done
+    - name: Test multi-architecture Docker images
```
mvdan: I'm not sure what the point of the "test" step is, since it happens after the image is already pushed. If CI fails because of these tests, I assume we shouldn't be pushing at all.

I do think the test is useful for the simplest sanity checks, but it seems more like an informative test at this stage. If that's what you were going for for now, that's fine - but perhaps we can add a TODO about making `docker push` conditional on this step.

.github/workflows/test.yml
ArturKlauser: Done

.github/workflows/test.yml
ArturKlauser: Done

.github/workflows/test.yml
ArturKlauser: You are right that the test is past docker push and as such doesn't prevent bad images to get to Docker Hub. It does test if the just pushed images can be pulled and used correctly though, so that's something.

Unfortunately there is no way that I'm aware of at this time to keep multi-architecture docker images in the local docker image cache. You *have to* push the multi-architecture `buildx build` image to Docker Hub in the same command (the `--push` flag). Whenever you have a local version of such an image, you always only have a single-architecture version corresponding to your host architecture. I.e. you also can't pull + retag + push a multi-architecture image without having narrowed down the pushed image to a single architecture.

ArturKlauser:
Should be all fixed now. Please take another look.
Note that if you want to try it on a non-master branch, you'll have to change test.yml
```
(github.ref == 'refs/heads/master' ||
```
`master` to your branch name in order for the new docker deployment code to run.

mvdan:
Yep, thanks! I come back from holidays on Monday, so I'll merge this coming week.

.github/workflows/test.yml
oliv3r: > nit: please use single spaces for comments. as far as I know, the double-space format is common in python, but not particularly in yaml or Go-related stuff

actually, yamllint complains about single space comments, doesn't it

cmd/shfmt/Dockerfile
```diff
@@ -1,9 +1,13 @@
-FROM golang:1.13.5-alpine3.10
+FROM golang:1.13.5-alpine3.11 as build
 
 WORKDIR /src
 COPY . .
 RUN CGO_ENABLED=0 go build -ldflags '-w -s -extldflags "-static"' ./cmd/shfmt
 
-FROM busybox:1.31.1-musl
-COPY --from=0 /src/shfmt /bin/shfmt
+FROM alpine:3.11 AS alpine
```
oliv3r: why add the 'AS' keyword if your not reusing it?

cmd/shfmt/Dockerfile
```diff
@@ -1,9 +1,13 @@
-FROM golang:1.13.5-alpine3.10
+FROM golang:1.13.5-alpine3.11 as build
```
oliv3r: i'd uppercase the 'AS' as it is a keyword (you did below in the alpine case)

README.md
```diff
@@ -126,6 +126,12 @@ To build a Docker image, checkout a specific version of the repository and run:
 
 	docker build -t my:tag -f cmd/shfmt/Dockerfile .
 
+This creates an image that only includes shfmt. Alternatively, if you want an
+image that includes alpine, add `--target alpine`.
```
oliv3r: is this still relevant, as there's only one target?

mvdan:
@oliv3r regarding your comments about the `alpine` target, it's because we want to offer two docker image flavors - as a scratch image that's as small and simple as possible, and an alpine one that has a shell and a package manager.

cmd/shfmt/Dockerfile
ArturKlauser: Done

.github/workflows/test.yml
ArturKlauser: Yes it does, it issues a warning.
But then, it complains about a bunch of things in this file, mainly wrong indentation of '-' lines, and I didn't want to go there.

cmd/shfmt/Dockerfile
ArturKlauser: Because you can specify it from the docker build command line with `--target alpine` and the CI actually does that when building the docker images.

cmd/shfmt/Dockerfile
ArturKlauser: Yup, my bad. I'll fix it.

README.md
ArturKlauser: Sure it's relevant. Why do you say there is only one target? You can build either without specifying a target, in which case you get the minimal *scratch* image, or you can build with `--target alpine`, in which case you get the alpine image.

cmd/shfmt/Dockerfile
oliv3r: learned something new today :) Didn't know you could have multiple targets in a single file, that's pretty cool and saves me from needing multiple dockerfiles!

README.md
oliv3r: wasn't aware of the target thing :) and missed the 'scratch' part (as that's a new thing and mdvan said something about distro's and just sticking with alpine)

from your reaction I gather a) we'll have at least 2 images, alpine and 'scratch' (makes perfect sense) and in that case, yes this is relevant :) sorry for the noise.

.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v1
     - name: Test as root, without cgo, and with busybox
-      run: docker run -v=$PWD:/pwd golang:1.13-alpine3.10 sh -c 'export CGO_ENABLED=0; cd /pwd && go test ./...'
+      run: docker run -v=$PWD:/pwd golang:1.13-alpine3.11 sh -c 'export CGO_ENABLED=0; cd /pwd && go test ./...'
```
oliv3r: While I know this is yaml, I still personally prefer to quote and brace shell stuffs.
e.g. -v="${PWD}:/pwd"
 actually, isn't $(pwd) preferred over $PWD?

in any case, also i think it's better to just set the 'workdir'.

```yaml
run: docker run --volume "$(pwd):/workdir" --workdir "/workdir" golang:1.13-alpine3.11 sh -c 'export CGO_ENABLED=0; go test ./...'
```

hmm, i just noticed you only bumped the version here :) leaving this for the future then

ArturKlauser: In general I agree with you. Hadolint is a nice tool that can check these sorts of things, and some checker should be in one's automated pre-commit checks if there are Dockerfiles in the repo. But as you say I only bump the alpine version here so I don't think it's appropriate to start muddling with everything else here too.

mvdan:
Thanks for your patience. I only just got home yesterday after a lot of travel, and this required about thirty minutes of work with proper internet. Here's the result of the pre-merge testing: https://github.com/mvdan/sh/runs/396544036

And here's the multi-arch image result: https://hub.docker.com/layers/mvdan/shfmt/latest/images/sha256-c3a15e777aa2c13063ae2817d7e6fa091b762c5495995234df58538eb1babdf8

Thanks again so much for this. I hope I can ping you in the future if this breaks or needs maintenance :) Hopefully the Docker folks will make the build setup much simpler.

ArturKlauser:
> I hope I can ping you in the future if this breaks or needs maintenance :)

Certainly, any time. Please ping me directly if you need to, since I don't scan Github notifications all that frequently and haven't figured out how to make sure that all of the important ones get send out to me by email as well. I've also written up a [post](https://nexus.eddiesinentropy.net/2020/01/12/Building-Multi-architecture-Docker-Images-With-Buildx/) which might give you some background on docker buildx if you're interested.

ArturKlauser:
BTW, you might want to copy&paste the Github README.md to DockerHub's Readme in the [Overview page of shfmt-alpine](https://hub.docker.com/r/mvdan/shfmt-alpine).

PR #472, opened 2020-01-09
cmd/shfmt: Make the docker container consistent

oliv3r:
Docker (the org) likes to have its containers be consistent [0] and by
adding a simple entrypoint script we become compliant. This would in the
future allow it to be possible to become an 'official' container.

In any case, it also makes the application behave as expected from a
users point of view.

e.g. running the container only arguments, invokes shfmt. Adding `shfmt`
as the first argument, also still works as expected.

More importantly, it allows for starting the container with a shell by
using `/bin/sh` as an argument.

The last point is important as it allows CI's such as gitlab to use the
container 'as is' and run shfmt inside of it.

[0] https://github.com/docker-library/official-images#consistency

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>

### DIFF
--- a/cmd/shfmt/docker-entrypoint.sh
+++ b/cmd/shfmt/docker-entrypoint.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# Copyright (C) 2019 Olliver Schinagl <oliver@schinagl.nl>
+#
+# A beginning user should be able to docker run image bash (or sh) without
+# needing to learn about --entrypoint
+# https://github.com/docker-library/official-images#consistency
+
+set -eu
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -gt 0 ] && \
+   [ "${1#-}" = "${1}" ] && \
+   command -v "${1}" > "/dev/null" 2>&1; then
+	exec "${@}"
+else
+	# else default to run the command
+	exec /bin/shfmt "${@}"
+fi
+
+exit 0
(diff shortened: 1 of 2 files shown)


mvdan:
Huh, I wasn't aware of this consistency rule, thanks for bringing it up.

Let's merge after https://github.com/mvdan/sh/pull/468 though, since that contributor has spent quite a bit of time on it with multiple reviews and I don't want to give them merge conflicts.

oliv3r:
I don't mind rebasing after his request is merged; just note I just added 3 comments to his PR :)

Having me do the rebasing is probably easier. Also if he wants he can cherry-pick my patch and add it to his PR as well if he wants.

cmd/shfmt/docker-entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+# A beginning user should be able to docker run image bash (or sh) without
+# needing to learn about --entrypoint
+# https://github.com/docker-library/official-images#consistency
+
+set -eu
+
+# run command if it is not starting with a "-" and is an executable in PATH
+if [ "${#}" -gt 0 ] && \
+   [ "${1#-}" = "${1}" ] && \
+   command -v "${1}"; then
```
ArturKlauser: Shouldn't that add `>/dev/null 2>&1` to  avoid the output?

oliv3r: > Shouldn't that add `>/dev/null 2>&1` to avoid the output?

Hah, you are right, I don't understand where I have lost it, as I just checked a different repo of mine, and sure enough > "/dev/null" 2>&1 right there.

So sorry, my bad!

mvdan:
@oliv3r can you please rebase, so we can merge?

mvdan:
@ArturKlauser ended up rebasing this as #490, so I'm closing this one. Thanks!

Issue #477, opened 2020-01-15
cmd/shfmt: add ability to skip unknown (non-shell) files


tbcs:
If scripts in a git repository don't have a `.sh` extension and are spread across various directories it's cumbersome to use `shfmt` in a pre-commit hook. It would be useful if `shfmt` could skip non-shell files when invoked from a git hook with the list of files being changed (as it is the case with [pre-commit](https://pre-commit.com/)).

This is similar to #288, but different, because the desired behavior here would be that not explicit configuration is required. Rather, `shfmt` should "magically" skip non-shell files.

mvdan:
Hm. I really don't want to add more options, particularly when the use case is a bit niche, and when we already have very similar flags like `-f`.

Here's another idea. What about allowing `-f` to work on files too? For example:

```
$ shfmt -f no-shebang shell-shebang extension.sh
shell-shebang
extension.sh
```

That way, you could do something like `shfmt -w $(shfmt -f $FILES)`. It does invoke the tool twice, but I doubt the overhead in this case matters much, and it allows the tool to remain simpler and more composable.

tbcs:
> Here's another idea. What about allowing -f to work on files too?

What do you mean by "allowing"? Because the `-f` flag "works" on files as of the current (3.0.1) version as well, it's just that the output in nonsensical: it would list any file specified as part of the result set, regardless of its detected type (it happily echoes the name of a `.jar` file).

If you meant that the behavior of `-f` could be changed to the one you described then yes, I think that would be a good thing to do! It would still require a wrapper script for a generic "reformat all shell scripts" pre-commit hook, but at least it would not need to scan the whole repository for scripts just to discard (most likely) the majority of them.

tbcs:
@mvdan: please see #485

mvdan:
I could have sworn I replied to your comment a week ago, but I assume it got lost.

Yes, I do mean changing the behavior of `-f`. Requiring a bit of extra glue is fine, I think. The tool is general purpose, and not just for pre-commit scripts.

PR #478, opened 2020-01-15
shfmt: add flag for skipping unknown (non-shell) files

tbcs:
When invoked with flag `-su` skip files which have not been detected as
shell scripts instead of returning an error.


tbcs:
There are no tests in this PR, it's only meant to facilitate discussion about #477.

Issue #481, opened 2020-01-18
Syncing forked repo fails fuzzing test


ArturKlauser:
I've noticed for a while that whenever I sync my forked repo `git checkout master; git pull upstream master; git push origin master`, one of the fuzzing test running on the Github Actions CI fails on my repo.

```
fuzz continuous jobs
##[error]Process completed with exit code 1.
Run export PATH="$PATH:$(go env GOPATH)/bin"; cd _fuzz/it && ./fuzz-ci fuzzing
+ go install github.com/dvyukov/go-fuzz/go-fuzz github.com/dvyukov/go-fuzz/go-fuzz-build github.com/fuzzitdev/fuzzit/v2
+ export GOFUZZ111MODULE=on
+ GOFUZZ111MODULE=on
+ go-fuzz-build -libfuzzer -o fuzzit-program.a .
+ clang -fsanitize=fuzzer fuzzit-program.a -o fuzzit-program
+ fuzzit create job --type fuzzing mvdan/sh-syntax-parser-printer fuzzit-program
2020/01/18 09:26:42 FUZZIT_API_KEY is not set. continue with public auth...
2020/01/18 09:26:42 Creating job...
2020/01/18 09:26:42 FUZZIT_API_KEY is not set. continue with public auth...
2020/01/18 09:26:42 Uploading fuzzer...
2020/01/18 09:26:43 401 Unauthorized
##[error]Process completed with exit code 1.
```
It seems specific to both
* being user *mvdan* (mvdan/sh-syntax-parser-printer)
* having a FUZZIT_API_KEY

I'm proposing to constrain running of this test to your repo, i.e. not running it in forks, in order to have a cleaner sync experience for contributors.
I'll send a PR for that.

PR #482, opened 2020-01-18
Fix fuzzing failure for forked repos

ArturKlauser:
Restrict CI test 'fuzz continuous jobs' to original repo user mvdan as poposed in #481.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,10 @@ jobs:
       run: export PATH="$PATH:$(go env GOPATH)/bin"; cd _fuzz/it && ./fuzz-ci local-regression
     - name: fuzz continuous job
       run: export PATH="$PATH:$(go env GOPATH)/bin"; cd _fuzz/it && ./fuzz-ci fuzzing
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -47,7 +47,10 @@ jobs:
       run: export PATH="$PATH:$(go env GOPATH)/bin"; cd _fuzz/it && ./fuzz-ci local-regression
     - name: fuzz continuous job
       run: export PATH="$PATH:$(go env GOPATH)/bin"; cd _fuzz/it && ./fuzz-ci fuzzing
-      if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+      if: >
+        github.event_name == 'push' &&
+        github.ref == 'refs/heads/master' &&
+        github.actor == 'mvdan'
```
mvdan: please use `github.repository == 'mvdan/sh'` instead. otherwise, if I was to give another person push access and they pushed to master, this wouldn't trigger.

ArturKlauser: Done

PR #485, opened 2020-01-21
shfmt: fix behavior of -f flag for file arguments

tbcs:
Ensure that invoking shfmt -f on files yields the same result as when
invoked with the directory containing the files. In other words, the
output of `shfmt -f` should only contain files which have been detected
as shell files. This changes shfmt's behavior whereby previously, given
the following file tree:

    bin
    ├── README.md
    └── script.sh

running `shfmt -f bin` would output:

    bin/script.sh

but running `shfmt -f bin/script.sh bin/README.md` would output:

    bin/script.sh
    bin/README.md

As of this commit the output for both cases is consistent.

### DIFF
--- a/cmd/shfmt/main_test.go
+++ b/cmd/shfmt/main_test.go
@@ -40,6 +40,7 @@ type action uint
 
 const (
 	None action = iota
+	Skip
 	Modify
 	Error
 )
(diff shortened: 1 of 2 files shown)


cmd/shfmt/main_test.go
```diff
@@ -70,9 +71,9 @@ var walkTests = []struct {
 	{None, false, "ext.other", " foo"},
 	{None, false, "ext-shebang.other", "#!/bin/sh\n foo"},
 	{None, false, "shebang-nospace", "#!/bin/envsh\n foo"},
-	{None, false, filepath.Join(".git", "ext.sh"), " foo"},
-	{None, false, filepath.Join(".svn", "ext.sh"), " foo"},
-	{None, false, filepath.Join(".hg", "ext.sh"), " foo"},
+	{Skip, false, filepath.Join("foo", ".git", "ext.sh"), " foo"},
```
mvdan: why the change with `"foo"`? that seems unnecessary

tbcs: That was just by mistake, removed now.

PR #490, opened 2020-01-25
cmd/shfmt: Make the docker container consistent

ArturKlauser:
Docker (the org) likes to have its containers be consistent [0] and by
adding a simple entrypoint script we become compliant. This would in the
future allow it to be possible to become an 'official' container.

In any case, it also makes the application behave as expected from a
users point of view.

e.g. running the container only arguments, invokes shfmt. Adding `shfmt`
as the first argument, also still works as expected.

More importantly, it allows for starting the container with a shell by
using `/bin/sh` as an argument.

The last point is important as it allows CI's such as gitlab to use the
container 'as is' and run shfmt inside of it.

[0] https://github.com/docker-library/official-images#consistency

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>


ArturKlauser:
@mvdan this is PR #472 from @oliv3r, just rebased so you can merge.

PR #491, opened 2020-01-25
Add support for '-' filename to mean stdin

ArturKlauser:
Proposal to treat '-' on the command line as stdin, like it has been done forever by many Unix command line tools.

This came up when I was trying to run the minimal docker image with stdin and needed to override the default docker image command line argument of '-h'. I.e. `cat my.sh | docker run -i --rm shfmt` currently doesn't work since the image executes 'shfmt -h'. One option is to override some command line argument with it's default value, e.g. `cat my.sh | docker run -i --rm shfmt -i 0`. But IMHO a cleaner option would be to support the '-' filename argument, i.e. `cat my.sh | docker -i --rm shfmt -`.

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -67,9 +67,9 @@ func main1() int {
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, `usage: shfmt [flags] [path ...]
 
-If path is a single dash ('-') or absent, standard input will be used. If a
-given path is a directory, it will be recursively searched for shell files -
-both by filename extension and by shebang.
+If the only argument is a dash ('-') or no arguments are given, standard input
+will be used. If a given path is a directory, it will be recursively searched
+for shell files - both by filename extension and by shebang.
 
   -version  show version and exit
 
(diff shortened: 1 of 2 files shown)


mvdan:
Hm. Why did we put `CMD ["-h"]` in there? Is that part of the Docker consistency rules document?

I get that `-` is pretty standard, and I'm not particularly opposed to it. But I want to understand why we want to make the help flag a default. `shfmt` on its own does not show its help when given no arguments, so I can't see a reason why the Docker image should be any different - that could confuse the users.

ArturKlauser:
In the [docker consistency reference](https://github.com/docker-library/official-images#consistency) @oliv3r gave in #472, point 3) insinuates that '-help' would be an appropriate default argument for the case of "image behaves as executable" and I remember having run into such images before. We can take that out if you think it's too confusing. As a novice user of such a docker container I find it kind of helpful. In any case, I'd consider that issue separate from this PR as such.

cmd/shfmt/main.go
```diff
@@ -67,9 +67,9 @@ func main1() int {
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, `usage: shfmt [flags] [path ...]
 
-If no arguments are given, standard input will be used. If a given path
-is a directory, it will be recursively searched for shell files - both
-by filename extension and by shebang.
+If path is a single dash ('-') or absent, standard input will be used. If a
```
mvdan: > If the only argument is a dash ('-') or no arguments are given,

I don't want to imply that we support weird cases like `shfmt - -` or `shfmt - foo.sh`. Some of them we might, but I'd rather leave that undocumented, because it doesn't seem like a good idea to encourage.

cmd/shfmt/main.go
```diff
@@ -155,27 +155,31 @@ Utilities:
...
-		return 1
+		args = []string{"-"}
 	}
 	status := 0
-	for _, path := range flag.Args() {
-		walk(path, func(err error) {
-			if err != errChangedWithDiff {
-				fmt.Fprintln(os.Stderr, err)
+	for _, path := range args {
+		if path == "-" {
```
mvdan: as per my comment above, don't do this in a loop; you can just do:

```
if len(args) == 0 || (len(args) == 1 && args[0] == "-") {
    // stdin
    return ...
}
// file loop
```

Otherwise, we have to think about difficult scenarios like `shfmt - -` or `shfmt - /dev/stdin`. Do we want the tool to block forever? Error in some way? I don't think it's worth keeping those as options.

cmd/shfmt/main.go
ArturKlauser: Done

cmd/shfmt/main.go
ArturKlauser: OK, I use that logic now but just went back to your original code for most part, since larger changes are not necessary in that case.

PR #492, opened 2020-01-30
Add DockerHub description upon deployment

ArturKlauser:
It irked me that moving the docker image build from DockerHub auto-build to Github Actions had lost the automatic update of the info (aka. repository description) on DockerHub's web UI. So this PR puts it back, using the DockerHub API.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,8 +67,9 @@ jobs:
     env:
       # Export environment variables for all stages.
       DOCKER_CLI_EXPERIMENTAL: enabled # for 'docker buildx'
-      DOCKER_BASE: mvdan/shfmt
+      DOCKER_USER: mvdan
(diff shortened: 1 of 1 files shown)


ArturKlauser:
I wanted it for one of my own repos anyway ;)

ArturKlauser:
Oh-oh, it looks like it didn't work on your setup. Seems like the DOCKER_PASSWORD you set in Github Actions is some sort of access token, not the actual password? DockerHub doesn't seem to allow updating the repository description with that. Not sure if you can choose the permissions for that token somewhere on DockerHub UI (or where ever you created that token).

From Github Actions log:
(node:17914) UnhandledPromiseRejectionWarning: Error: {"detail":"access to the resource is forbidden with personal access token"}

mvdan:
Yeah, I did set up a token. Not too happy about having to use my real password, but it's a generated one anyway, so I'll change it.

PR #497, opened 2020-02-07
Set runner exitShell to true before calling the exec handler for exec builtin

cclerget:
Set runner exitShell to true before calling the exec handler for exec builtin, this way exec handlers could retrieve this information via runner.Exited() and take appropriate actions.

Not sure if this is the best way to handle that but in a first time it would allow to retrieve this information without dealing with statements and parsing to know if it's an exec builtin or a sub-process.

### DIFF
--- a/interp/handler_test.go
+++ b/interp/handler_test.go
@@ -38,6 +38,17 @@ func blacklistNondevOpen(ctx context.Context, path string, flags int, mode os.Fi
 	return testOpenHandler(ctx, path, flags, mode)
 }
 
+// runnerCtx allows us to give handler functions access to the Runner, if needed.
+var runnerCtx = new(int)
+
+func execBuiltin(ctx context.Context, args []string) error {
+	runner, ok := ctx.Value(runnerCtx).(*Runner)
+	if ok && runner.Exited() {
+		return fmt.Errorf("exec builtin: %s", args[0])
+	}
+	return nil
+}
+
 var modCases = []struct {
 	name string
 	exec ExecHandlerFunc
@@ -82,6 +93,12 @@ var modCases = []struct {
 		want: "blacklisted: malicious",
 	},
 	{
+		name: "ExecBuiltin",
+		exec: execBuiltin,
+		src:  "exec /bin/sh",
+		want: "exec builtin: /bin/sh",
+	},
+	{
 		name: "OpenForbidNonDev",
 		open: blacklistNondevOpen,
 		src:  "echo foo >/dev/null; echo bar >/tmp/x",
@@ -106,7 +123,7 @@ func TestRunnerHandlers(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			ctx := context.Background()
+			ctx := context.WithValue(context.Background(), runnerCtx, r)
 			if err := r.Run(ctx, file); err != nil {
 				cb.WriteString(err.Error())
 			}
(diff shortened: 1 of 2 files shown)


mvdan:
I'm not sure I follow - could you provide a real example? The `ExecHandlerFunc` interface only has a context with some data, and that doesn't include the `*Runner`, so I'm not sure how you'd call `Runner.Exited`.

mvdan:
Having said the above, I'm not opposed to the change. I'd like a test to go along with the change, though, as otherwise it might break again in a future refactor.

cclerget:
By embedding runner in the context :

```
package main

import (
    "context"
    "os"
    "log"
    "fmt"

    "mvdan.cc/sh/v3/interp"
    "mvdan.cc/sh/v3/syntax"
)

var runnerCtx = struct{}{}

var DefaultExec = interp.DefaultExecHandler(2)

func ExecHandler() interp.ExecHandlerFunc {
    return func(ctx context.Context, args []string) error {
        runner := ctx.Value(&runnerCtx).(*interp.Runner)
        if runner.Exited() {
            fmt.Println("exec builtin")
        }
        return DefaultExec(ctx, args)
    }
}

func main() {
    script := "/tmp/test.sh"

    f, err := os.Open(script)
    if err != nil {
        log.Fatal(err)
    }
    defer f.Close()

    runner, _ := interp.New(
        interp.StdIO(os.Stdin, os.Stdout, os.Stderr),
        interp.ExecHandler(ExecHandler()),
    )
    parser := syntax.NewParser()

    ctx := context.WithValue(context.Background(), &runnerCtx, runner)

    sf := func(st *syntax.Stmt) bool {
        runner.Run(ctx, st)
        return true
    }

    parser.Stmts(f, sf)
}
```

> Having said the above, I'm not opposed to the change. I'd like a test to go along with the change, though, as otherwise it might break again in a future refactor.

Sure

mvdan:
Thanks - your use case seems fine :) Adding that as a standalone test shouldn't be too hard.

cclerget:
@mvdan Thanks ! Test added

interp/handler_test.go
```diff
@@ -81,6 +93,12 @@ var modCases = []struct {
 		src:  "{ malicious; true; } & { malicious; true; } & wait",
 		want: "blacklisted: malicious",
 	},
+	{
+		name: "ExecBuiltin",
+		exec: execBuiltin(),
```
mvdan: remove the `()`, and then the func doesn't have to return another func.

interp/handler_test.go
```diff
@@ -38,6 +38,18 @@ func blacklistNondevOpen(ctx context.Context, path string, flags int, mode os.Fi
 	return testOpenHandler(ctx, path, flags, mode)
 }
 
+var runnerCtx = struct{}{}
```
mvdan: use `&struct{}{}`, so that you can just do `runnerCtx` later instead of `&runnerCtx`

mvdan: Also add a short comment, like:

```
// runnerCtx allows us to give handler functions access to the Runner, if needed.
```

cclerget:
Should be good to go and commit squashed

interp/handler_test.go
theclapp: re: `&struct{}{}` — does it matter that that might not be unique?  See [here](https://dave.cheney.net/2014/03/25/the-empty-struct):

> Interestingly, the address of two struct{} values may be the same.
> 
> ```go
> var a, b struct{}
> fmt.Println(&a == &b) // true
> ```

mvdan: Ah, right. @cclerget please use a non-empty allocated value, like `= new(int)`

cclerget: Fixed

mvdan:
Thanks! I've squashed to just adapt the commit message to the format I use, but otherwise I've left the commit as it was.

PR #501, opened 2020-02-12
   release: Also build and push alpine containers

oliv3r:
We currently have two containers to choose from during the build stage,
an alpine container and a scratch container.
    
The scratch container is there for size, where we do not want or need
anything other then just the raw binary.
    
The alpine container exists for when more is needed, such as a shell or
apk to add packages.
    
CI pipelines need the later, due to the requirement of needing a shell
among things. To cater this need, also build the alpine based
container and push it with the -alpine tag, much like other
multi-target docker containers do.

Whilst here, also clean-up the script a little.


release-docker.sh
```diff
@@ -1,14 +1,25 @@
...
 
-docker build -t mvdan/shfmt:$tag -f cmd/shfmt/Dockerfile .
+docker build --tag "mvdan/shfmt:${tag}" \
+             --file "cmd/shfmt/Dockerfile" \
+             "./"
+
+docker build --tag "mvdan/shfmt:${tag}-alpine" \
+             --target "alpine" \
+             --file "cmd/shfmt/Dockerfile" \
+             "."
```
mvdan: follow `shfmt` :)

release-docker.sh
```diff
@@ -1,14 +1,25 @@
-#!/bin/bash
+#!/bin/sh
 
-tag=$(git describe --exact-match 2>/dev/null)
-if [[ $? != 0 ]]; then
-	tag=latest
+set -eu
+
+if ! tag="$(git describe --exact-match 2> "/dev/null")"; then
```
mvdan: hm, I prefer the older form, as it's more obvious. I don't like expressions which are also statements.

In a way, it reminds me of Go's `if foo := bar; expr... {`, where the definition is only part of the nested scope. So I find it pretty confusing that it actually modifies the parent scope.

release-docker.sh
```diff
@@ -1,14 +1,25 @@
...
-	echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+if [ -n "${DOCKER_PASSWORD:-}" ]; then
+	echo "${DOCKER_PASSWORD}" | docker login --password-stdin \
+	                                         --username "${DOCKER_USERNAME}"
 fi
 
-docker push mvdan/shfmt:$tag
+docker push "mvdan/shfmt:${tag}" "mvdan/shfmt:${tag}-alpine"
+
+exit 0
```
mvdan: I've never seen this pattern - why?

mvdan:
Also CC @ArturKlauser who wrote this code initially. I don't have strict shell patterns, other than shfmt and being portable. In that sense, using POSIX Shell sounds good.

ArturKlauser:
Actually, you're mistaken, that's not the code I wrote. This one comes from https://github.com/mvdan/sh/commit/a3dcc70e39089b6bedc25682c0420f91c2e779de which might have been in use a while ago before you switched to DockerHub auto-build and before we switched to do the docker builds as part of the GitHub Actions CI. In fact, if you'd run this script now, it'd clobber your multi-architecture docker containers built by CI with single-architecture containers built by this script. Maybe it's time to retire this script all together?

mvdan:
You're right! Complete brain fart on my part, for some reason I thought this was still being used as part of CI/CD. Yes, we should delete it. @oliv3r you can do it as part of this PR, or if not, I'll do it in master.

oliv3r:
I removed the script, but I'm not sure how this autobuild thing is supposed to work. I've looked into the .github-workflow and it seems like it should be building and pushing something, but appearantly, this does not happen. Maybe the target is wrong? now it's "$DOCKER_BASE-alpine:$TAG" which I think requires this to be its own repository, so we better put the -alipne postfix after the $TAG.

release-docker.sh
oliv3r: I'll reply, even though the file has been deleted now ;) Just because :p
You are right, I didn't, I had forgotten and my OCD wanted to make it look pretty. I'll try to remember to use your style here :)

release-docker.sh
oliv3r: I think that's purely personal, as what you are familiar with. But if you prefer the earlier, that's fine as well. I like this style better as it doesn't rely on having to manually deal with `$?`

release-docker.sh
oliv3r: I think it's in the google bash rules.

It basically states, always have a clean and reliable  exit point. For example, if someone manages to append code to your script due some leak in the system, the script would not execute this rogue code. Also `set -eu` ensures if you reach this point, exiting is safe.

mvdan:
You're right. It's been pushing to https://hub.docker.com/r/mvdan/shfmt-alpine/tags, which is an entirely new repo I didn't intend to create. Let's move that to a separate issue: https://github.com/mvdan/sh/issues/504

PR #503, opened 2020-02-12
interp: make "type" builtin respect the current $PATH

### LINKED ISSUE #502: interp: "type" does not respect $PATH if it's set within the shell
actual:
```sh
% gosh
$ PATH=/
$ type ping
ping is /sbin/ping
```

expected — something more like:
```sh
% PATH=/ ./gosh
$ type ping
type: ping: not found
```

Edit: It's actually a problem with "type"; see below.

theclapp:
Fixes #502.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -25,6 +25,15 @@ import (
 	"mvdan.cc/sh/v3/syntax"
 )
 
+// Some program which should be in $PATH.  Needs to run before runTests is
+// initialized (so an init function wouldn't work), because runTest uses it.
+var pathProg = func() string {
+	if runtime.GOOS == "windows" {
+		return "cmd"
+	}
+	return "sh"
+}()
+
 func parse(tb testing.TB, parser *syntax.Parser, src string) *syntax.File {
 	if parser == nil {
 		parser = syntax.NewParser()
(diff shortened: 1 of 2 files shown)


theclapp:
Regarding the locale change:

```go
-	os.Setenv("LANGUAGE", "en_US.UTF8")
-	os.Setenv("LC_ALL", "en_US.UTF8")
+	os.Setenv("LANGUAGE", "en_US.UTF-8")
+	os.Setenv("LC_ALL", "en_US.UTF-8")
```

Without this change, I got failures galore.  This didn't used to fail.  I think it does now because I upgraded bash to 5.0.16 locally.

I looked at trying to install that locale, or copy the existing one, and came up (mostly) empty.  All the UTF8 locales on my system are `...UTF-8` not `...UTF8`.  Copying the one to the other failed:

```sh
# in /usr/share/locale
% sudo cp -rp en_US.UTF-8 en_US.UTF8
cp: cannot create directory 'en_US.UTF8': Operation not permitted
```

Not sure what you can or should do about that.  I assume the current "UTF8" code works for you and everybody else.  🤷‍♂

theclapp:
Fixes #502.

interp/interp_test.go
```diff
@@ -99,8 +108,8 @@ func TestMain(m *testing.M) {
 	}
 	os.Setenv("GOSH_PROG", prog)
 
-	os.Setenv("LANGUAGE", "en_US.UTF8")
-	os.Setenv("LC_ALL", "en_US.UTF8")
+	os.Setenv("LANGUAGE", "en_US.UTF-8")
+	os.Setenv("LC_ALL", "en_US.UTF-8")
```
mvdan: I agree this is a good change. Not sure why the old version worked. I think it was a typo of mine.

interp/interp_test.go
```diff
@@ -25,6 +25,15 @@ import (
 	"mvdan.cc/sh/v3/syntax"
 )
 
+// Some program which should be in $PATH.  Needs to run before runTests is
+// initialized, below.
```
mvdan: explain why - I assume it's because we want to use it in "expected output" string literals

theclapp:
Done.  PTAL.

PR #510, opened 2020-02-18
Fix issue with readonly and naked export.

cclerget:
Doing the following should not return a readonly variable error

```
readonly foo=bar
export foo
```


cclerget:
Sure will do, thanks !

PR #547, opened 2020-04-21
Link to Emacs shfmt package

purcell:
I wrote an Emacs package to integrate `shfmt`, which may interest some of your users. Thanks!


purcell:
> Thanks! Please keep the "related projects" list sorted.

Sorry, hadn't noticed that was how it was organised. Will update now.

Issue #576, opened 2020-05-30
interp: process substitution named pipes seem to be flaky on Darwin


mvdan:
Split from https://github.com/mvdan/sh/issues/513, since this one can be reproduced by @theclapp on his Mac machine.

It doesn't seem to be the same issue that we're seeing with pseudo-TTYs on CI, which we haven't been able to reproduce ourselves. These failed tests run without any sort of TTY.

Some examples:

```
    --- FAIL: TestRunnerRun/817 (0.01s)
        interp_test.go:2668: wrong output in "cat <(cat <(echo nested))":
            want: "nested\n"
            got:  "file exists\n"

    --- FAIL: TestRunnerRun/817 (0.00s)
        interp_test.go:2668: wrong output in "cat <(cat <(echo nested))":
            want: "nested\n"
            got:  ""

    --- FAIL: TestRunnerRun/817 (0.05s)
        interp_test.go:2667: wrong output in "echo foo bar > >(sed 's/o/e/g')":
            want: "fee bar\n"
            got:  ""

     --- FAIL: TestRunnerRun/814 (0.00s)
        interp_test.go:2684: wrong output in "sed 's/o/e/g' <(echo foo bar)":
            want: "fee bar\n"
            got:  "open /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/sh-interp-133253605: no such file or directory"
FAIL
```

None of these are reproducible on Linux with stress testing, apparently. Thoughts:

* Is our code racy? Right now, we assume that it's OK for the named pipe writer to delete the file when done writing, under the assumption that the reader goroutine has already opened the pipe.
* Following the above, we assume that writes to the named pipe won't succeed until the reads have happened. This might be a wrong assumption, given the "no such file or directory" error above.
* On Linux, `unix.Mkfifo` calls `unix.Mknod`, which calls the `mknod` syscall. On Darwin, `unix.Mkfifo` calls the `mkfifo` syscall via libc directly, presumably since the `mknod` syscall requires root privileges.
* Are Linux and Darwin named pipe semantics any different? E.g. buffering, or deleting the file while a reader is still reading.

mvdan:
Another one:

```
 --- FAIL: TestRunnerRun (0.03s)
    --- FAIL: TestRunnerRun/822 (0.00s)
##[error]        interp_test.go:2701: wrong output in "echo foo bar | tee >(sed 's/o/e/g') >/dev/null":
            want: "fee bar\n"
[...]

PR #580, opened 2020-06-11
syntax: replace Arithmetic parser with a recursive-descent parser

### LINKED ISSUE #579: syntax: Arithmetic expression precedence is wrong
Hi!

I ran into a parsing error. Say you have an expression like `(( 2 != 0 == 2 ))`. This gets parsed like this in sh:

```
*syntax.File {
.  Name: "/dev/stdin"
.  StmtList: syntax.StmtList {
.  .  Stmts: []*syntax.Stmt (len = 1) {
.  .  .  0: *syntax.Stmt {
.  .  .  .  Comments: []syntax.Comment (len = 0) {}
.  .  .  .  Cmd: *syntax.ArithmCmd {
.  .  .  .  .  Left: 1:1
.  .  .  .  .  Right: 1:16
.  .  .  .  .  Unsigned: false
.  .  .  .  .  X: *syntax.BinaryArithm {
.  .  .  .  .  .  OpPos: 1:6
.  .  .  .  .  .  Op: 0x28
.  .  .  .  .  .  X: *syntax.Word {
.  .  .  .  .  .  .  Parts: []syntax.WordPart (len = 1) {
.  .  .  .  .  .  .  .  0: *syntax.Lit {
.  .  .  .  .  .  .  .  .  ValuePos: 1:4
.  .  .  .  .  .  .  .  .  ValueEnd: 1:5
.  .  .  .  .  .  .  .  .  Value: "2"
.  .  .  .  .  .  .  .  }
.  .  .  .  .  .  .  }
.  .  .  .  .  .  }
.  .  .  .  .  .  Y: *syntax.BinaryArithm {
.  .  .  .  .  .  .  OpPos: 1:11
.  .  .  .  .  .  .  Op: 0x27
.  .  .  .  .  .  .  X: *syntax.Word {
.  .  .  .  .  .  .  .  Parts: []syntax.WordPart (len = 1) {
.  .  .  .  .  .  .  .  .  0: *syntax.Lit {
.  .  .  .  .  .  .  .  .  .  ValuePos: 1:9
.  .  .  .  .  .  .  .  .  .  ValueEnd: 1:10
.  .  .  .  .  .  .  .  .  .  Value: "0"
.  .  .  .  .  .  .  .  .  }
.  .  .  .  .  .  .  .  }
.  .  .  .  .  .  .  }
.  .  .  .  .  .  .  Y: *syntax.Word {
.  .  .  .  .  .  .  .  Parts: []syntax.WordPart (len = 1) {
.  .  .  .  .  .  .  .  .  0: *syntax.Lit {
.  .  .  .  .  .  .  .  .  .  ValuePos: 1:14
.  .  .  .  .  .  .  .  .  .  ValueEnd: 1:15
.  .  .  .  .  .  .  .  .  .  Value: "2"
.  .  .  .  .  .  .  .  .  }
.  .  .  .  .  .  .  .  }
.  .  .  .  .  .  .  }
.  .  .  .  .  .  }
.  .  .  .  .  }
.  .  .  .  }
.  .  .  .  Position: 1:1
.  .  .  .  Semicolon: 0:0
.  .  .  .  Negated: false
.  .  .  .  Background: false
.  .  .  .  Coprocess: false
.  .  .  .  Redirs: []*syntax.Redirect (len = 0) {}
.  .  .  }
.  .  }
.  .  Last: []syntax.Comment (len = 0) {}
.  }
}
```

So as `(( 2 != (0 == 2) ))`. This is wrong, because it should actually be parsed as `(( (2 !=) 0 == 2 ))`. This is a problem because `==` and `!=` are not [associative](https://en.wikipedia.org/wiki/Associative_property).

Here's an example of how this leads to different behavior in bash and gosh:
```
$ bash
$ (( 2 != 0 == 2 ))
$ echo $?
1

$ gosh
$ (( 2 != 0 == 2 ))
$ echo $?
0
```

bouk:
Here I'm replacing the parser with a recursive-descent parser that's
modeled after the parser used in Bash. All the methods correspond to
methods in Bash's expr.c.

This fixes operator precedence problems like #579

### TODO

- [x] Fix all the error message tests

Fixes #579

### DIFF
--- a/syntax/parser_arithm.go
+++ b/syntax/parser_arithm.go
@@ -1,0 +1,484 @@
+package syntax
+
+import ()
+
+// compact specifies whether we allow spaces between expressions.
+// This is true for let
+func (p *Parser) arithmExpr(compact bool) ArithmExpr {
+	return p.arithmExprComma(compact)
+}
+
+// These function names are inspired by Bash's expr.c
+
+func (p *Parser) arithmExprComma(compact bool) ArithmExpr {
+	value := p.arithmExprAssign(compact)
+	for BinAritOperator(p.tok) == Comma {
+		if compact && p.spaced {
+			return value
+		}
+		pos := p.pos
+		tok := p.tok
+		p.nextArithOp(compact)
+		y := p.arithmExprAssign(compact)
+		if y == nil {
+			p.followErrExp(pos, tok.String())
+		}
+		value = &BinaryArithm{
+			OpPos: pos,
+			Op:    BinAritOperator(tok),
+			X:     value,
+			Y:     y,
+		}
+	}
+	return value
+}
+
+func (p *Parser) arithmExprAssign(compact bool) ArithmExpr {
+	value := p.arithmExprCond(compact)
+	switch BinAritOperator(p.tok) {
+	case AddAssgn, SubAssgn, MulAssgn, QuoAssgn, RemAssgn, AndAssgn,
+		OrAssgn, XorAssgn, ShlAssgn, ShrAssgn, Assgn:
+		if !isArithName(value) {
+			p.posErr(p.pos, "%s must follow a name", p.tok.String())
+		}
+		pos := p.pos
+		tok := p.tok
+		p.nextArithOp(compact)
+		y := p.arithmExprAssign(compact)
+		if y == nil {
+			p.followErrExp(pos, tok.String())
+		}
+		return &BinaryArithm{
+			OpPos: pos,
+			Op:    BinAritOperator(tok),
+			X:     value,
+			Y:     y,
+		}
+	}
+	return value
+}
+
+func (p *Parser) arithmExprCond(compact bool) ArithmExpr {
+	value := p.arithmExprLor(compact)
+	if BinAritOperator(p.tok) == TernQuest {
+		if compact && p.spaced {
+			return value
+		}
+		questPos := p.pos
+		p.nextArithOp(compact)
+		if BinAritOperator(p.tok) == TernColon {
+			p.followErrExp(questPos, TernQuest.String())
+		}
+		trueExpr := p.arithmExpr(compact)
+		if trueExpr == nil {
+			p.followErrExp(questPos, TernQuest.String())
+		}
+		if BinAritOperator(p.tok) != TernColon {
+			p.posErr(p.pos, "ternary operator missing : after ?")
+		}
+		colonPos := p.pos
+		p.nextArithOp(compact)
+		falseExpr := p.arithmExprCond(compact)
+		if falseExpr == nil {
+			p.followErrExp(colonPos, TernColon.String())
+		}
+		return &BinaryArithm{
+			OpPos: questPos,
+			Op:    BinAritOperator(TernQuest),
+			X:     value,
+			Y: &BinaryArithm{
+				OpPos: colonPos,
+				Op:    BinAritOperator(TernColon),
+				X:     trueExpr,
+				Y:     falseExpr,
+			},
+		}
+	}
+
+	return value
+}
+
+func (p *Parser) arithmExprLor(compact bool) ArithmExpr {
+	value := p.arithmExprLand(compact)
+	for BinAritOperator(p.tok) == OrArit {
+		if compact && p.spaced {
+			return value
+		}
+		op := p.tok
+		pos := p.pos
+		p.nextArithOp(compact)
+		y := p.arithmExprLand(compact)
+		if y == nil {
+			p.followErrExp(pos, op.String())
+		}
+		value = &BinaryArithm{
+			OpPos: pos,
+			Op:    BinAritOperator(op),
+			X:     value,
+			Y:     y,
+		}
+	}
+	return value
+}
+
+func (p *Parser) arithmExprLand(compact bool) ArithmExpr {
+	value := p.arithmExprBor(compact)
+	for BinAritOperator(p.tok) == AndArit {
+		if compact && p.spaced {
+			return value
+		}
+		op := p.tok
+		pos := p.pos
+		p.nextArithOp(compact)
+		y := p.arithmExprBor(compact)
+		if y == nil {
+			p.followErrExp(pos, op.String())
+		}
+		value = &BinaryArithm{
+			OpPos: pos,
+			Op:    BinAritOperator(op),
+			X:     value,
+			Y:     y,
+		}
+	}
+	return value
+}
+
+func (p *Parser) arithmExprBor(compact bool) ArithmExpr {
+	value := p.arithmExprBxor(compact)
+	for BinAritOperator(p.tok) == Or {
+		if compact && p.spaced {
+			return value
+		}
+		op := p.tok
+		pos := p.pos
+		p.nextArithOp(compact)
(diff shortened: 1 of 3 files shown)


bouk:
Basically all the code is new, only a little bit of code in arithmExpr0 was copied over.

So I'm afraid creating the new file in a separate commit would not help much!

mvdan:
Gotcha, thanks.

syntax/parser_arithm.go
```diff
@@ -0,0 +1,482 @@
...
+			OpPos: pos,
+			Op:    BinAritOperator(op),
+			X:     value,
+			Y:     y,
+		}
+	}
+	return value
+}
+
+func (p *Parser) arithmExpr2(compact bool) ArithmExpr {
```
mvdan, line 311: A lot of these funcs copy the same ~20 lines and that's making me a bit uneasy. It's really easy to make mistakes when copying code around, or forgetting to update some of the copies when doing fixes.

Any thoughts on how we could deduplicate the code? I realise it would be a bit more inefficient to have one function with more branches, but honestly, I prefer a nearly-insignificant performance hit that brings an improvement in maintainability.

I also don't find the 1,2,3,4 function names super helpful :) I get that we also used numbers before, but to my mind those were isolated and didn't mean repeating code to this degree.

bouk: I agree those names are super unhelpful, but I was inspired by bash: https://github.com/bminor/bash/blob/9e49d343e3cd7e20dad1b86ebfb764e8027596a7/expr.c#L780

I wanted to have the structure be identical to bash's, so the precedence rules are exactly the same. All the duplication is a bit weird, but I'm not sure how to improve it while keeping the same structure. I can try to improve it though, maybe I can add some helpers or something

mvdan: Here's an idea - a higher-level function that takes a list of operators to check against, and a function value as the "next" level. like:

```
type nextFunc func(compact bool, ops []token, next nextFunc) ArithmExpr

func arithmLevel(compact bool, ops []token, next nextFunc) ...
```

Of course, this will have a small perf hit, but I think it's worthwhile to avoid the ~5x copy-paste. An alternative would be to lay out these "levels" in a table, similar to what I do in master with the switch and level integers. Then, each table entry can tell you what operators need to be handled, and what the "next" level or function will be.

mvdan: In other words, I appreciate that recursive descent is easier to write, but it's also pretty hard to maintain when there are so many levels of precedence involved. This is why I coded it up with a table in master. Arguably, that was buggy, but I still think it should be possible to do it that way and not add another 200+ lines of mostly-repeated code.

bouk: :+1: That's what I was working towards 😄 

I'll DRY it up

bouk: I've refactored it, saved 200 lines of code! much better I think. Now all the functions look different at least

syntax/parser_arithm.go
```diff
@@ -0,0 +1,318 @@
...
+	return p.arithmExprBinary(compact, p.arithmExprPower, Mul, Quo, Rem)
+}
+
+func (p *Parser) arithmExprPower(compact bool) ArithmExpr {
+	// Power is different from the other binary operators because it's right-associative
+	value := p.arithmExpr1(compact)
+	if BinAritOperator(p.tok) == Pow {
```
mvdan, line 130: if you invert this comparison, you can then unindent most of the code. the same applies to other funcs.

syntax/parser_arithm.go
```diff
@@ -0,0 +1,318 @@
...
+		return true
+	}
+	for !compact && p.got(_Newl) {
+	}
+	return false
+}
+
+func (p *Parser) nextArithOp(compact bool) {
+	pos := p.pos
+	tok := p.tok
```
mvdan, line 250: you add quite a lot of code to deal with saving and reusing previous tokens and positions. as far as I can tell, the previous code didn't do that, so I'm a bit confused.

syntax/parser_arithm.go
```diff
@@ -0,0 +1,318 @@
...
+
+// nextArith consumes a token.
+// It returns true if compact and the token was followed by spaces
+func (p *Parser) nextArith(compact bool) bool {
+	p.next()
+	if compact && p.spaced {
+		return true
+	}
+	for !compact && p.got(_Newl) {
+	}
```
mvdan, line 244: if I recall correctly, you should only care about a single _Newl token, since they are deduplicated. so you could do:

```
if !compact {
    p.got(_Newl)
}
```

More lines, but also simpler, and it's best to avoid loops in general as they just add more ways for the parser to get stuck or burn cpu forever.

syntax/parser_arithm.go
```diff
@@ -0,0 +1,318 @@
...
+	value := nextOp(compact)
+	for {
+		var foundOp BinAritOperator
+		for _, op := range operators {
+			if p.tok == token(op) {
+				foundOp = op
+				break
+			}
+		}
+		if foundOp == 0 {
```
mvdan, line 267: please use illegalTok instead, which is a nicer way to say 0

syntax/parser_arithm.go
```diff
@@ -0,0 +1,318 @@
...
+			OpPos: pos,
+			Op:    BinAritOperator(tok),
+			X:     value,
+			Y:     y,
+		}
+	}
+	return value
+}
+
+func (p *Parser) arithmExprCond(compact bool) ArithmExpr {
```
mvdan, line 44: nit: perhaps replace Cond with Tern, since that's probably less ambiguous to the reader

syntax/parser_arithm.go
```diff
@@ -0,0 +1,318 @@
...
+
+func (p *Parser) arithmExpr4(compact bool) ArithmExpr {
+	return p.arithmExprBinary(compact, p.arithmExprShift, Lss, Gtr, Leq, Geq)
+}
+
+func (p *Parser) arithmExprShift(compact bool) ArithmExpr {
+	return p.arithmExprBinary(compact, p.arithmExpr3, Shl, Shr)
+}
+
+func (p *Parser) arithmExpr3(compact bool) ArithmExpr {
```
mvdan, line 119: why not replace all the numbers with names, while we're at it? for example, 3 could be AddSub, 4 could be Compare, 5 could be Equal, and 2 could be MulDiv.

I get that the numbers are inherited from bash, and that they help explain the order, but jumping between funcs that are numbered isn't generally helpful. If we want to clarify the order somewhere, I would do it with a small table in comments near the top of the file.

bouk: Yup you're right, I'll make that change

syntax/parser_arithm.go
bouk: ah didn't know they're deduplicated by the lexer, will do

bouk:
Tests are passing now ✅

mvdan:
Don't worry about that mac failure, that's #576.

Thanks for your work! I have some more thoughts, but I'll just continue in master.

mvdan:
Sigh, pressed the wrong button, I meant to squash-merge. I'll have to force-push to master.

PR #581, opened 2020-06-15
syntax: Record token at which parser stops for default case.

Hamza-Q:
Printing the token where the parser stops can help when debugging why a shell command fails to get parsed, so I'd like to upstream a patch to do so. 

Let me know your thoughts on this.

Thanks,
Hamza

### DIFF
--- a/syntax/parser.go
+++ b/syntax/parser.go
@@ -2274,7 +2274,7 @@ loop:
 			}
 			fallthrough
 		default:
-			p.curErr("a command can only contain words and redirects")
+			p.curErr("a command can only contain words and redirects. This command stopped at token: '%s'", p.tok)
 		}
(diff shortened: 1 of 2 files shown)


syntax/parser.go
```diff
@@ -2274,7 +2274,7 @@ loop:
 			}
 			fallthrough
 		default:
-			p.curErr("a command can only contain words and redirects")
+			p.curErr("a command can only contain words and redirects. This command stopped at token: '%s'", p.tok)
```
mvdan, line 2277: error messages are a single sentence. how about:

    "a command can only contain words and redirects; encountered %s", p.tok)

Hamza-Q: Sure, updated

Issue #261, opened 2018-07-02
syntax: add option to configure how many statements in a list can fit in a single line


tamalsaha:
Thanks for this tool. In our shell script, we have lines like 

```
 	local cmd="docker build -t $DOCKER_REGISTRY/$IMG:$TAG ."
	echo $cmd; $cmd
```

We would like to not break lines like `echo $cmd; $cmd` into 2 separate lines. I did not see any flag to support this. Can this be supported?

mvdan:
There was a similar suggestion here earlier: https://github.com/mvdan/sh/issues/250

The suggestion there was to add an option to never split statements. Yours seems to be related to minimum length. Do you have a strong opinion either way?

I'm not terribly convinced by either suggestion. The earlier one is all-or-nothing, and this one is based on a heuristic. Neither are bad ideas per se, but I only want to add new features and flags (which I'll have to maintain for a long time) if I'm sure they will be simple and useful.

Here's another point - would this apply to other statements like blocks and subshells? Or should this only apply within lists of statements? For example, suppose this line was under the threshold:

    foo() { bar; baz; }

Should it be turned into the following?

    foo() {
        bar
        baz
    }

/cc @dcsobral

tamalsaha:
Thanks for the quick response.

> The suggestion there was to add an option to never split statements.

I don't think that makes sense, since I want to optionally disable line break.

Regarding the block statement, we have number of cases like

```
kubectl config current-context || { echo "Set a context (kubectl use-context <context>) out of the following:"; echo; kubectl config get-contexts; exit 1; }
```
Again one heuristic will be to a not break if # of statements is <= 2 .

I have seen linters from other languages support per line directive via comments. So, one generic way to do this can be:

```
 	local cmd="docker build -t $DOCKER_REGISTRY/$IMG:$TAG ."
	echo $cmd; $cmd # NOWRAP
```
This kind of looks ugly to me. But I can live with this. May be I would move to the right most column by auto formatting.

mvdan:
> Again one heuristic will be to a not break if # of statements is <= 2.

I'm not sure I follow. Do you want that line you pasted to be broken into many, or not? When you said min length, I presumed you meant a number of columns, not a number of statements.

> I have seen that linters from other languages usually support per line directive via comments.

I'm really not a fan of directives, so I doubt I'll ever add those.

The problem, then, is teaching the tool when it is fine to cram many statements into a single line.

tamalsaha:
[...]

PR #585, opened 2020-07-08
Added an option to disable line splitting

patrickvane:
Added option `-ns` which will disable line splitting. This option is for people who just want the formatter to fix indention and spacing, people who don't want their whole code to be restructured (by having newlines added to it and by having semicolons be removed/replaced with newlines).

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -48,6 +48,7 @@ var (
 	spaceRedirs = flag.Bool("sr", false, "")
 	keepPadding = flag.Bool("kp", false, "")
 	funcNext    = flag.Bool("fn", false, "")
+	noSplitting = flag.Bool("ns", false, "")
 
 	toJSON = flag.Bool("tojson", false, "")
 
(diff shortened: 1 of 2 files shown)


mvdan:
It's unfortunate that you raised a PR without first discussing this feature, because I generally don't accept new features or flags without any discussion first.

In this particular case, this has been discussed multiple times before, and I think https://github.com/mvdan/sh/issues/261 would be a better way to go about this.

patrickvane:
Hi, thanks for replying. I made this feature because I needed it myself, I figured might as well share it with others in case they need it too. I was looking at issue #261 and it had already been years (that issue is from 2018?) so I figured it's likely never going to happen, hence I made a simple fix/option myself. Either way, thanks for replying and good luck. On Jul 9, 2020 17:39, Daniel Martí <notifications@github.com> wrote:
It's unfortunate that you raised a PR without first discussing this feature, because I generally don't accept new features or flags without any discussion first.
In this particular case, this has been discussed multiple times before, and I think #261 would be a better way to go about this.

mvdan:
The project is in active development. There just are a lot of issues to work on, and making the formatter more lax is not at the top of my TODO list :) Thanks for understanding.

PR #586, opened 2020-07-09
Restore unified diff on -d flag

lvjp:
Between the `v2.6.4` and `v3.0.0` there is a regression in the diff output.

Before it was done with the command `diff -u` but now with the package `github.com/pkg/diff` without the [WithContextSize](https://pkg.go.dev/github.com/pkg/diff@v0.0.0-20190915220856-5883b2e1a98f?tab=doc#EditScript.WithContextSize) option output a full diff which is a regression and boring.

I just add a `WithContext(3)` to restore the former behavior.


mvdan:
Thanks! Makes sense.

PR #594, opened 2020-07-29
Preserve indentation on line start with -kp and space indent

### LINKED ISSUE #471: syntax: KeepPadding should obey extra indentation when indenting with spaces
We currently have the option to choose between indentation format, tabs vs spaces. However, it is very common (when using tabs) to only use tabs for indentation, but use spaces to align.

The most obvious spot is when aligning multiple if terms on multiple lines.

To give an example.
```console
	image="some long string" \
	      "spanning multiple lines"

	if [ -n "${image}" ] && \
	   docker build \
	          --hostname "$(hostname)"
	          --rm \
	          --interactive \
	          --tty \
	          --workdir "${workdir}" \
	          "${image}" "$(pwd)"; then
	fi
```
(note that github doesn't use a monospace font for rendered code, but not when editing/entering text *sigh*)
Here we have one indentation level (the level of the if) and alignment of the arguments.

The reason why we want to do this, is that regardless of the tabstop setting, things are always properly aligned. Using spaces only fixes this of course, but tab users don't want to ;)

I tried using -kp but that only messes with the arguments (hostname and workdir) and aligns those.

When [searching](https://duckduckgo.com/html?q=tab%20to%20indent%2C%20space%20to%20align) for it we get plenty of results.

I think vim calls it [smart-tabs](https://vim.fandom.com/wiki/Indent_with_tabs,_align_with_spaces#Smart_tabs)?

Since you probably don't want a new option for this, this would either be needed to be auto-detected (how to know when a space is for alignment vs just a stray whitespace following an indentation tab) or -i could be changed from uint to sint and -1 would be tabs with alignment :) This would 'break' if a user insist on having INT_MAX + 1 - UINT_MAX number of spaces, but since having anything more then 128 spaces shouldn't even be considered I think we're good.

maximbaz:
I went through a few scripts that caused me trouble, and with this patch they all seem to behave fine.

Reflected in tests all the gotchas that I stumbled upon, namely my initial example, a simple indented line, comments.

If you can think of more edge cases, please let me know 🙂 
If I find some examples where this logic fails, I will return with another PR 😄 

Fixes #471

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -831,9 +831,8 @@ func (p *Printer) wordJoin(ws []*Word) {
 				anyNewline = true
 			}
 			p.bslashNewl()
-		} else {
-			p.spacePad(w.Pos())
 		}
+		p.spacePad(w.Pos())
 		p.word(w)
 	}
 	if anyNewline {
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -269,13 +269,12 @@ func (p *Printer) space() {
 }
 
 func (p *Printer) spacePad(pos Pos) {
-	if p.cols.lineStart && (!p.keepPadding || p.indentSpaces == 0) {
-		// Never add padding at the start of a line unless keepPadding
-		// is set and we are indenting with spaces, since this may
-		// result in broken indentation or mixing of spaces and tabs.
+	if p.cols.lineStart && p.indentSpaces == 0 {
+		// Never add padding at the start of a line unless we are indenting
+		// with spaces, since this may result in mixing of spaces and tabs.
 		return
 	}
-	if p.wantSpace && !p.cols.lineStart {
+	if p.wantSpace {
 		p.WriteByte(' ')
 		p.wantSpace = false
 	}
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -269,16 +269,17 @@ func (p *Printer) space() {
 }
 
 func (p *Printer) spacePad(pos Pos) {
-	if p.cols.lineStart {
-		// Never add padding at the start of a line, since this may
+	if p.cols.lineStart && (!p.keepPadding || p.indentSpaces == 0) {
+		// Never add padding at the start of a line unless keepPadding
+		// is set and we are indenting with spaces, since this may
 		// result in broken indentation or mixing of spaces and tabs.
 		return
 	}
-	if p.wantSpace {
+	if p.wantSpace && !p.cols.lineStart {
 		p.WriteByte(' ')
 		p.wantSpace = false
 	}
-	for !p.cols.lineStart && p.cols.column > 0 && p.cols.column < int(pos.col) {
+	for p.cols.column > 0 && p.cols.column < int(pos.col) {
 		p.WriteByte(' ')
 	}
 }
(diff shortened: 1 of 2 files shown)


mvdan:
Please ignore that CI failure. Mac has been a constant source of headaches for the CI here.

syntax/printer_test.go
```diff
@@ -844,6 +857,24 @@ func TestPrintKeepPadding(t *testing.T) {
...
+		samePrint("array=('one'\n        # 'two'\n        'three')"),
+		samePrint("    abc=123"),
+	}
+	parser := NewParser(KeepComments(true))
+	printer := NewPrinter(KeepPadding(true), Indent(2))
+	for i, tc := range tests {
+		t.Run(fmt.Sprintf("%03d", i), func(t *testing.T) {
+			// ensure that Reset does properly reset colCounter
+			printer.WriteByte('x')
+			printer.Reset(nil)
```
mvdan, line 872: I don't think you need to repeat this Reset check here :)

syntax/printer.go
```diff
@@ -269,16 +269,17 @@ func (p *Printer) space() {
 }
 
 func (p *Printer) spacePad(pos Pos) {
-	if p.cols.lineStart {
-		// Never add padding at the start of a line, since this may
+	if p.cols.lineStart && (!p.keepPadding || p.indentSpaces == 0) {
+		// Never add padding at the start of a line unless keepPadding
+		// is set and we are indenting with spaces, since this may
 		// result in broken indentation or mixing of spaces and tabs.
```
mvdan, line 275: The only result we're avoiding here is mixing spaces and tabs now, so you should remove the bit about "broken indentation". When I wrote that, I did indeed mean obeying the user's choice of space indentation, as opposed to enforcing the equivalent of tabs.

syntax/printer.go
```diff
@@ -269,16 +269,17 @@ func (p *Printer) space() {
 }
 
 func (p *Printer) spacePad(pos Pos) {
-	if p.cols.lineStart {
-		// Never add padding at the start of a line, since this may
+	if p.cols.lineStart && (!p.keepPadding || p.indentSpaces == 0) {
```
mvdan, line 272: I *think* you can avoid the keepPadding check here, because we didn't have it before. I think that, when keepPadding is false, the column counter isn't running, so p.cols.column==0 and the rest of the logic doesn't trigger.

syntax/printer.go
```diff
@@ -269,16 +269,17 @@ func (p *Printer) space() {
...
+		// Never add padding at the start of a line unless keepPadding
+		// is set and we are indenting with spaces, since this may
 		// result in broken indentation or mixing of spaces and tabs.
 		return
 	}
-	if p.wantSpace {
+	if p.wantSpace && !p.cols.lineStart {
```
mvdan, line 278: hmm, this feels backwards; if anything calls spacePad at the start of a line, then wantSpace should be false. If you don't know how to fix that, that's fine, you can leave a TODO here and I'll look into it at a later point.

maximbaz: No you are right, this additional check was useless

maximbaz:
>  If you only want to fix those too, that would be great. If you prefer to keep this PR as-is, then make it an "Updates", not a "Fixes".

I added the tests you suggested and made sure that the code covers them, but I'm really not sure that I in fact covered _all_ scenarios of #471. If you are also not sure that we indeed covered everything, I would be totally fine with us replacing "Fixes" to "Updates" (btw you can just edit my first message before merge).

mvdan:
Thanks - this looks great :)

PR #599, opened 2020-08-18
Update walk_test.go

comalice:
Add missing paren.


mvdan:
Sorry, but I don't get this. There isn't any missing parenthesis. The builds break with your change, as you can see above.

comalice:
Oh! `go` syntax confused me here. That's an anonymous function declaration.

PR #603, opened 2020-08-25
feat: compatibility with browser usage

JounQin:


_js/go.mod
```diff
@@ -8,5 +8,6 @@ replace mvdan.cc/sh/v3 => ../
 
 require (
 	github.com/gopherjs/gopherjs v0.0.0-00010101000000-000000000000
+	golang.org/dl v0.0.0-20200811212135-d149fc5456ff // indirect
```
mvdan, line 11: undo this, and all the go.sum changes.

_js/build
```diff
@@ -1,12 +1,13 @@
 #!/bin/sh
 
-if ! go version | grep -q go1.12; then
+if ! command -v "go1.12" >/dev/null 2>&1; then
 	echo "Go 1.12.x required to build."
+	echo "run \`go get golang.org/dl/go1.12 && go1.12 download\` first"
```
mvdan, line 5: undo this as well. you can set up your PATH accordingly instead, like:

```
mkdir -p some/bin
ln -sf $(which go1.12) some/bin/go
export PATH="$PWD/some/bin:$PATH"
# run your command here
```

Most people will likely use the 1.12 wrapper, but we shouldn't add a hard dependency on it.

_js/index.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
mvdan, line 1: I'm not sure I understand what this file does. Please document it.

_js/main.go
```diff
@@ -14,7 +14,19 @@ import (
...
+		exps = js.Module.Get("exports")
+	} else {
+		exps = js.MakeWrapper(map[string]interface{}{})
+		originalGlobalSh := js.Global.Get("sh")
+		exps.Set("noConflict", func() *js.Object {
+			js.Global.Set("sh", originalGlobalSh)
+			return exps
+		})
+		js.Global.Set("sh", exps)
+	}
```
mvdan, line 29: I also don't understand what any of this does, sorry - please document it.

mvdan (changes requested):
I can imagine what you're trying to do here, but without any comments and no commit message, I just don't understand much. Please expand via comments.

_js/build
JounQin: The main go package `Requires Go 1.14 or later`, so I imagined that we have to install multiple go versions, but if you don't like it, I can revert it.

_js/go.mod
JounQin: OK, it should be related to `go get golang.org/dl/go1.12`.

_js/index.html
JounQin: Testing browser usage, or should I rename this file?

_js/main.go
JounQin: In this condition, `module` is not available, so we need to assign `window.sh` to get access to exported `sh` object outside.

_js/build
mvdan: Yes, you will need to install multiple versions unfortunately, but what I'm saying is that you can do that outside the script pretty easily without having to hard-code it into the script :)

JounQin:
> I can imagine what you're trying to do here, but without any comments and no commit message, I just don't understand much. Please expand via comments.

@mvdan `js.Module.Get('exports')` means `module.exports` but it is not available on browser, so we need to check whether `module` be available. On browser, we need to set the reference to `window.sh`(or something else).

mvdan:
Gotcha, the explanations above make sense. Just leave them as comments in the code, because I want them in git.

JounQin:
@mvdan Done.

_js/build
JounQin: It seems redundant to change go versions between folders at a same project personally, but of course it's fine to keep it as-is.

mvdan (approved):
Thank you!

PR #605, opened 2020-09-10
repair linkrot: mirbsd.org no longer reliably HTTPS-accessible

LucasLarson:
and while `http://mirbsd.de/mksh` “[is the canonical homepage URI” “via a redirection](http://www.mirbsd.org/mksh.htm)”, that address is also pointing to mirbsd.org over HTTPS, which is no longer accessible without disabling secure settings.

cc @mirabilos

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ repo configuration like:
 [homebrew]: https://formulae.brew.sh/formula/shfmt
 [macports]: https://ports.macports.org/port/shfmt/summary
 [micro]: https://micro-editor.github.io/
-[mksh]: https://www.mirbsd.org/mksh.htm
+[mksh]: http://www.mirbsd.org/mksh.htm
 [modd]: https://github.com/cortesi/modd
 [mvdan-sh]: https://www.npmjs.com/package/mvdan-sh
 [nixos]: https://github.com/NixOS/nixpkgs/blob/HEAD/pkgs/tools/text/shfmt/default.nix


mvdan:
Damn, that's pretty bad. Thanks for noticing. I assume this wasn't a problem before, because browsers still accepted older TLS versions. It's worrying that in 2020 they don't suppot TLS 1.2.

mvdan (approved):

mirabilos:
Lucas Larson dixit:

>mirbsd.org over HTTPS, which is no longer accessible

mirbsd.org indeed doesn’t work, but www.mirbsd.org is still
accessible, and nothing changed on the server.

bye,
//mirabilos
-- 
>> Why don't you use JavaScript? I also don't like enabling JavaScript in
> Because I use lynx as browser.
+1
	-- Octavio Alvarez, me and ⡍⠁⠗⠊⠕ (Mario Lang) on debian-devel

mvdan:
It does not work on the current Firefox version:

> Error code: SSL_ERROR_UNSUPPORTED_VERSION

> This website might not support the TLS 1.2 protocol, which is the minimum version supported by Firefox. Enabling TLS 1.0 and TLS 1.1 might allow this connection to succeed.

> TLS 1.0 and TLS 1.1 will be permanently disabled in a future release.

mvdan:
Suffice to say, re-enabling TLS 1.0 or 1.1 would be a temporary workaround at best, and not a good idea in any case.

mirabilos:
Daniel Martí dixit:

>It does not work on the current Firefox version:
>
>> Error code: SSL_ERROR_UNSUPPORTED_VERSION

Oh, indeed, that just came in yesterday with the switch
to a new ESR. But I use lynx more anyway…

Well, can’t be helped at the moment. It’s a secured-ish
TLSv1.0, and unless you’re a bank or something, more isn’t
needed, but meh. There’s no login, the same content is
also delivered over http, so there’s your workaround.

bye,
//mirabilos
-- 
<cnuke> den AGP stecker anfeilen, damit er in den slot aufm 440BX board passt…
oder netzteile, an die man auch den monitor angeschlossen hat und die dann für
ein elektrisch aufgeladenes gehäuse gesorgt haben […] für lacher gut auf jeder
LAN party │ <nvb> damals, als der pizzateig noch auf dem monior "gegangen" ist

mvdan:
Up to the site admin :) But navigating that site is going to be difficult or at least confusing for anyone with an updated browser. I agree that an HTTP link is OK for now, but hopefully there's a long-term plan to support TLS 1.2 and 1.3.

mirabilos:
Daniel Martí dixit:

>agree that an HTTP link is OK for now, but hopefully there's a
>long-term plan to support TLS 1.2 and 1.3.

Yeah, some day I get the time and effort needed to port LibreSSL
to MirBSD and drop all the unlicenced parts from it…

bye,
//mirabilos
-- 
This space for rent.

https://paypal.me/mirabilos to support my work.

PR #607, opened 2020-09-16
interp: $@ and $* always exist

### LINKED ISSUE #606: interp: $@ and $* always exist
The interpreter had a bug where the empty expansion is not removed in a
call when a shift operation caused it to become empty. This is due to
shift's behavior of setting interp.Runner.Params to nil when no more
positional parameters are left.

Unfortunately, looking up $@ and $* returns a value with with List set
to interp.Runner.Params, meaning the List field can be nil when the
variable claims to be set. This breaks expand's expectation of List type
existence meaning non-zero values are available, as seen in
`quotedElemFields`.

twilly:
The interpreter had a bug where the empty expansion is not removed in a
call when a shift operation caused it to become empty. This is due to
shift's behavior of setting interp.Runner.Params to nil when no more
positional parameters are left.

Unfortunately, looking up $@ and $* returns a value with with List set
to interp.Runner.Params, meaning the List field can be nil when the
variable claims to be set. This breaks expand's expectation of List type
existence meaning non-zero values are available, as seen in
`quotedElemFields`.

This changes interp's environment lookup to always return a non-nil
[]string for $@ and $*, satisfying expand's expectations.

Fixes #606.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -297,6 +297,7 @@ var runTests = []runTest{
 	{`set -- b c; echo a"$@"d`, "ab cd\n"},
 	{`count() { echo $#; }; set --; count "$@"`, "0\n"},
 	{`count() { echo $#; }; set -- ""; count "$@"`, "1\n"},
+	{`count() { echo $#; }; set -- ""; shift; count "$@"`, "0\n"},
 	{`count() { echo $#; }; a=(); count "${a[@]}"`, "0\n"},
 	{`count() { echo $#; }; a=(""); count "${a[@]}"`, "1\n"},
 	{`echo $1 $3; set -- a b c; echo $1 $3`, "\na c\n"},
--- a/interp/vars.go
+++ b/interp/vars.go
@@ -70,7 +70,13 @@ func (r *Runner) lookupVar(name string) expand.Variable {
 	case "#":
 		vr.Kind, vr.Str = expand.String, strconv.Itoa(len(r.Params))
 	case "@", "*":
-		vr.Kind, vr.List = expand.Indexed, r.Params
+		vr.Kind = expand.Indexed
+		if r.Params == nil {
+			// r.Params may be nil but positional parameters always exist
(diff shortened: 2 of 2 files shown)


mvdan (approved):
Thanks for the thorough investigation and patch!

twilly:
> Thanks for the thorough investigation and patch!

And thank you for maintaining such an awesome open source project. It's a very neat tool. :)

mvdan:
I'm curious, what are you using the interpreter for?

twilly:
> I'm curious, what are you using the interpreter for?

I've got an internal CLI tool that's ~1k lines of shell calling `git` and `jq` underneath. It's fairly well written, including good test coverage, and fits a <noun> <verb> CLI style (much like git). Verbs are functions with the main function little more than a big switch statement. (Perhaps you can see why I ran into this bug - shifting with/to an empty $@/$* is common in this CLI pattern.)

This internal tool works well but it's finally getting to the point where it'd be better off as a pure Go program. The conversion strategy is pretty simple: have the main program be in Go, embed and unpack the shell at runtime, and start implementing verbs until the program is pure Go. Such a strategy seems easy enough, but there's a catch: these verbs are composed. Implementing a single verb may require implementing several.

I could get around this by creating my own FFI using `jq` to pass JSON encoded messages between the shell and Go function implementations, or I could use the mvdan/sh's parsing and AST interpretation to skip this whole dance. Not only do I no longer depend on the host's, often extremely old, shell, but I have more control over how the shell executes. The interpreter may be experimental, but I've got test coverage and control over its executing environments. I've got the freedom to explore this crazy idea of mvdan/sh as a FFI. :)

mvdan:
Sounds cool! Happy to receive further PRs for the interpreter, or issues with any ideas you might have. Since the v3 release the interpreter is no longer experimental (at least feature and API wise), but it is of course a work in progress, as implementing all of Bash could take years. But it should work fine for most common code.

PR #613, opened 2020-09-26
shfmt: Don't add parentheses to functions, if they were not in the original source

jansorg:
Bash supports `foo() {}`, `function foo {}`, and `function foo() {}`.
Currently, shfmt always prints parentheses, i.e. `function foo {}` is always printed as `function foo() {}`.

This patch changes shfmt to only add parentheses, if they were in the original source.
Reasons:
- imo this may be unexpected for some users of Bash
- ksh only supports `foo() {}` and `function foo {}`, but not `function foo() {}`. This patch slightly increases compatibility with this and hopefully allows to run shfmt on more .ksh file (which is useful for [BashSupport Pro](https://www.bashsupport.com))

As this is my first PR for shfmt, this may not match the expectations or requirements :) Please let me know what should be changed or if this isn't how shfmt is supposed to work. Thansk!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   - Add `-filename` to give a name to standard input
 - **syntax**
   - Rewrite arithmetic parsing to fix operator precedence
+  - Don't add parentheses to `function f {...}` declarations for ksh support
 
 ## [3.1.2] - 2020-06-26
 
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -713,16 +713,18 @@ var fileTests = []testCase{
 			"foo()\n{\na\nb\n}",
 		},
 		common: &FuncDecl{
-			Name: lit("foo"),
-			Body: stmt(block(litStmt("a"), litStmt("b"))),
+			Parens: true,
+			Name:   lit("foo"),
+			Body:   stmt(block(litStmt("a"), litStmt("b"))),
 		},
 	},
 	{
 		Strs: []string{"foo() { a; }\nbar", "foo() {\na\n}; bar"},
 		common: []Command{
 			&FuncDecl{
-				Name: lit("foo"),
-				Body: stmt(block(litStmt("a"))),
+				Parens: true,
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - **cmd/shfmt**
   - Add `-filename` to give a name to standard input
+  - Don't add parentheses to functions, if they were not in the original source
```
mvdan, line 7: the commit message and changelog should be about `syntax` instead, since that's where the change is being made (shfmt is just one of the consumers)

mvdan (commented):
My initial reaction was going to be "this hurts consistency"; if a shell script is all Bash, for example, there's little reason why some funcs should use one form and others another. But we don't enforce that kind of consistency between `foo() {...}` and `function foo() {...}` either, so we're not losing any ground here, really.

I also wasn't aware that this was impeding ksh support. If you think it would be valuable to explicitly support ksh, raise an issue and we can talk about that. It would probably not be much work, given that we already support bash and mksh.

Otherwise the change looks good! I can squash-merge later so that your PR body remains in the commit message in master.

CHANGELOG.md
mvdan: Also, I would make this a little shorter and clearer:

```
- Don't add parentheses to `function {...}` declarations for ksh support
```

jansorg: Thanks!
As this is mostly about ksh support, adding "official" support for that would be good. I'll create a new issue for that. 
It's up to you if you want to merge this nontheless. I don't mind if this PR is dropped instead of ksh support.
Let me know if you want to merge this and I'll clean it up.

mvdan: I think merging this PR is fine anyway :) please do finish it up and I'll merge.

jansorg: @mvdan I've updated changelog and commit message.

mvdan: Thanks!

mvdan (approved):

PR #618, opened 2020-10-21
Add Mega-Linter in related projects

nvuillam:
If you have a logo / icon, please notify me and i'll add it in your linter related page :)
https://github.com/nvuillam/mega-linter/blob/master/docs/descriptors/bash_shfmt.md#readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ repo configuration like:
 - Alternative docker images - by [jamesmstone][dockerized-jamesmstone], [PeterDaveHello][dockerized-peterdavehello]
 - [format-shell] - Atom plugin for `shfmt`
 - [intellij-shellcript] - Intellij Jetbrains `shell script` plugin includes `shfmt`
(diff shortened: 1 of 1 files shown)


mvdan:
Thanks for your pull request! I think I'm going to decline this one, because I don't want the "related projects" section to include any project related to this repository at any level - that would quickly turn into dozens of links, which wouldn't be useful.

I think the cutoff point would be:

* Any project popular enough which integrates with this project (like micro)
* Any project whose main purpose is to wrap this project (like vim-shfmt)

So I don't think any linter or formatter aggregator which includes shfmt should be in the list. I went through the existing list again and I don't think any existing items should be removed, but I will make a commit to clarify the purpose of the section.

nvuillam:
@mvdan > No problem, your repo -> your choice :)
Just thought it could help to have your linter in CI, so when users does not install it locally, they can't escape the linting anyway haha ^^

PR #627, opened 2020-11-02
WIP: interp: add default variable test

zimbatm:
If a variable is accessed, but the pattern also provides a default
value, `set -u` should not throw an error.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3046,6 +3046,11 @@ func TestRunnerOpts(t *testing.T) {
 			"foo\nunset: unbound variable\nexit status 1",
 		},
 		{
+			opts(Params("-u", "--", "foo")),
+			"echo $@; echo ${unset:-default}",
+			"foo\ndefault\n",
+		},
+		{
 			opts(Params("foo")),
 			"set >/dev/null; echo $@",
 			"foo\n",


zimbatm:
Sorry, this is a failing test only for now. I think that the solution would need for this part of the code to look ahead: https://github.com/zimbatm/sh/blob/7f541b79ae97d518ef31adfbc28545c269767247/interp/vars.go#L126-L130

mvdan:
Ah, you're right that we're doing this at the wrong level. `set -u` in this case should not complain, because the default should kick in before the `nounset` check, but we do it the opposite way.

This fix will be a bit non-trivial because shell expansion is implemented in the `expand` package, separately from the `interp` package. It's configured via the `ecfg` runner field.

We already pass along a shell option to its config via `GlobStar`, so we probably need to do the same here like `NoUnset`.

mvdan:
If the above is confusing, I'm happy to help further over on the #shell channel on gophers.slack.com. You can join here: https://invite.slack.golangbridge.org/

Alternatively, I'm happy to take this over if you don't have the time.

zimbatm:
Thanks for the invite! For now, I want to keep prodding the project for more Bash incompatibilities. I will eventually join the channel and dig more into the code as I do so.

Issue #631, opened 2020-11-13
Unable to pull container with podman


fauust:
Hi, this is not an issue.

I am moving most of my docker usage to podman (https://podman.io/).

For most of the cases it is relatively transparent but in the case of shfmt container, I am unable to pull the image:
```console
❯ podman pull mvdan/shfmt:v3.2.0
Trying to pull docker.io/mvdan/shfmt:v3.2.0...
  unsupported docker v2s2 media type: "application/vnd.oci.image.layer.v1.tar+gzip"
```

I am not sure where the incompatibility comes from (container itself so build process or docker registry). Maybe there is a workaround?

If you could provide a podman compatible container, that would be very appreciated. If I can help in that task, please let me know.

mvdan:
I don't know. I build multi-architecture images with buildkit, so maybe that's too new/experimental for podman.

fauust:
No idea, no urgency either, thanks for your response.
Here is an issue that may give information https://github.com/containers/buildah/issues/2266

mvdan:
If someone wants to investigate this, I'll happily accept a pull request. Otherwise, I'll likely close this issue in a couple of weeks since I don't intend to investigate what needs to be done to support podman. Maintaining the necessary details to build multi-architecture images with buildkit (see CI) is already plenty of time taken by publishing to Docker.

fauust:
I'll investigate this, can you confirm that the build/push is all done into github actions?

mvdan:
Yes, the code is all [here](https://github.com/mvdan/sh/blob/d0927061ffa5d137ca779488192ee87254e68052/.github/workflows/test.yml#L39).

You can ignore the last step. It just updates the docker image's README with the README in this repo.

Similarly, the previous large step just tests that `docker run` works with the images as expected.

`Install Docker buildx` is also a lot of code to just... install a recent enough Docker version that has buildx. This was done when GitHub Actions used Ubuntu 18.04, which was too old. They've introduced 20.04 now, so maybe that step is not necessary anymore. I haven't dug into it yet.

The author of a lot of this code is @ArturKlauser, who might be kind enough to give tips or help in some other way.

fauust:
So, what I have found out so far is that I am not able to reproduce the problem by building a very simple container locally (but I am able to reproduce the problem with shfmt container build locally):

- create a basic Dockerfile containing the following:

```Dockerfile
FROM alpine:3.12.1
RUN apk update && apk add bash
```
- start the local registry:

```console
[...]

PR #638, opened 2020-11-28
Use official Docker GitHub actions

fauust:
Use the following actions instead of manual Docker buildx setup:
- https://github.com/docker/login-action
- https://github.com/docker/setup-qemu-action
- https://github.com/docker/build-push-action

Notes:
- It would be good to merge "Build and push" and "Build and push
  (alpine)" steps but I have no solution to propose for now;
- there are some linting changes (using prettier).

WIP because CI needs to be tested against upstream repo:
- create a DOCKER_USER secret so that forks can test the CI (push);
- I am using token and 2FA on DockerHub so the update description step
  could not be tested.

Regarding the last step (Update DockerHub description), we may consider using this https://github.com/marketplace/actions/docker-hub-description.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,9 +48,10 @@ jobs:
       github.repository == 'mvdan/sh'
     env:
       # Export environment variables for all stages.
-      DOCKER_CLI_EXPERIMENTAL: enabled # for 'docker buildx'
-      DOCKER_USER: mvdan
-      DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
+      DOCKER_USER: ${{ secrets.DOCKER_USER }}
+      # Pushing READMEs to Dockerhub currently only works with username/password
+      # and not with personal access tokens (Step: Update DockerHub description)
+      DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
       DOCKER_REPO: shfmt
       # We use all platforms for which FROM images in our Dockerfile are
       # available.
@@ -69,11 +70,29 @@ jobs:
       # Get "https://proxy.golang.org/...": local error: tls: unexpected message
       # Get "https://proxy.golang.org/...": x509: certificate signed by unknown authority
     steps:
+    - name: Check GitHub settings
+      run: |
+        missing=()
+        [[ -n "${{ secrets.DOCKER_USER }}" ]] || missing+=(DOCKER_USER)
+        [[ -n "${{ secrets.DOCKER_PASSWORD }}" ]] || missing+=(DOCKER_PASSWORD)
+        for i in "${missing[@]}"; do
+          echo "Missing github secret: $i"
+        done
+        (( ${#missing[@]} == 0 )) || exit 1
     - name: Checkout code
       uses: actions/checkout@v2
       with:
         fetch-depth: 0 # also fetch tags for 'git describe'
-    - name: Set up image tag
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Login to DockerHub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_USER }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+    - name: Set up env vars
       run: |
         set -vx
         # Export environment variable for later stages.
@@ -84,18 +103,27 @@ jobs:
           # Pushes tag - deploy tag name.
           echo "TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
         fi
-        echo "DOCKER_BASE=${DOCKER_USER}/${DOCKER_REPO}" >> $GITHUB_ENV
-    - name: Install Docker buildx
+        echo "DOCKER_BASE=${{ secrets.DOCKER_USER }}/${{ env.DOCKER_REPO }}" >> $GITHUB_ENV
+        echo "DOCKER_BUILD_PLATFORMS=${DOCKER_PLATFORMS// /,}" >> $GITHUB_ENV
+    - name: Build and push
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./cmd/shfmt/Dockerfile
+        platforms: ${{ env.DOCKER_BUILD_PLATFORMS }}
+        push: true
+        tags: ${{ env.DOCKER_BASE }}:${{ env.TAG }}
+    - name: Build and push (alpine)
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./cmd/shfmt/Dockerfile
+        platforms: ${{ env.DOCKER_BUILD_PLATFORMS }}
+        push: true
+        tags: ${{ env.DOCKER_BASE }}:${{ env.TAG }}-alpine
+        target: alpine
+    - name: Test multi-architecture Docker images
       run: |
-        set -vx
-        # Install up-to-date version of docker, with buildx support.
-        docker_apt_repo='https://download.docker.com/linux/ubuntu'
-        curl -fsSL "${docker_apt_repo}/gpg" | sudo apt-key add -
-        os="$(lsb_release -cs)"
-        sudo add-apt-repository "deb [arch=amd64] $docker_apt_repo $os stable"
-        sudo apt-get update
-        sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
-
         # Enable docker daemon experimental support (for 'pull --platform').
         config='/etc/docker/daemon.json'
         if [[ -e "$config" ]]; then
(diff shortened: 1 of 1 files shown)


fauust:
You can verify that the build/push was successful here https://github.com/fauust/sh/runs/1467387436?check_suite_focus=true and here https://hub.docker.com/repository/docker/fauust/shfmt.

mvdan (commented):
Impressive work! Nice to see that we can drop a significant amount of tricky code.

Could you undo the indentation changes? They seem unnecessary, and will make the change significantly smaller.

Also, is this still WIP given that you tested it?

.github/workflows/test.yml
```diff
@@ -69,119 +65,108 @@ jobs:
...
-            process.env.DOCKER_PASSWORD)
-          .then(function () {
-            let url = "https://github.com/" + process.env.GITHUB_REPOSITORY;
-            let repo = process.env.DOCKER_REPO;
-            dockerHubAPI.setRepositoryDescription(
+      - name: Install Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 13.x
+      - name: Update DockerHub description
```
ArturKlauser, line 146: You might also want to replace this hand-rolled code with https://github.com/peter-evans/dockerhub-description

ArturKlauser (approved):
Very nice. The support for buildx has come a long way since I put in that code here. If you have any doubts on the details of why certain pieces of my original code are there, I've written up a [post](https://nexus.eddiesinentropy.net/2020/01/12/Building-Multi-architecture-Docker-Images-With-Buildx/) that might help you.

.github/workflows/test.yml
```diff
@@ -69,119 +65,108 @@ jobs:
...
+      - name: Build and push (alpine)
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./cmd/shfmt/Dockerfile
+          platforms: ${{ env.DOCKER_PLATFORMS }}
+          push: true
+          tags: ${{ env.DOCKER_BASE }}:${{ env.TAG }}-alpine
+          target: alpine
+      - name: Test multi-architecture Docker images
```
ArturKlauser, line 111: Ideally this would also be reorganized. What the current flow does is:
* build and push the image to dockerhub
* test the image from dockerhub to check if it works
* if there is a problem, an error is reported, however, the bogus image is already uploaded to dockerhub

It seems that there is built-in support now that makes it easy to [spin up a local temporary docker registry for testing purposes](https://github.com/docker/build-push-action#local-registry). So the better approach would be:

* build and push the image to a local temporary docker registry
* test the image from the local docker registry to check if it works
* if there is a problem report an error and quit
* if the tests were successful, push the image to dockerhub

.github/workflows/test.yml
fauust: Yes, fully agree, that's why I also suggested it in above ;-)

To be honest, I do not like so much the adherence to GitHub that this PR brings and you code was very instructive and useful to understand the workflow! I'll definitively take a look at your post.

Anyway, in the long run, this PR will help maintenance so I guess it's a good step forward :-)

.github/workflows/test.yml
fauust: Again, yes! Fully agree. Let's try to implement this in a next PR. Also, I would suggest splitting the workflows in multiple files to facilitate maintenance.

fauust:
> Could you undo the indentation changes? They seem unnecessary, and will make the change significantly smaller.
Sure I will! But hey, that's sounds reasonable to me to lint the yml files in a shell linting project ;-)
Maybe in a future PR then...

> Also, is this still WIP given that you tested it?
The only thing that is missing is that you create a secret (that is not secret at all) but will facilitate GitHub action contributions:
Secret in your case is `DOCKER_USER = mvdan`

ArturKlauser:
Suggestion regarding the required DOCKER_USER secret (and others): It's always a pain for contributors to reverse-engineer the implicit requirements of CI/CD flows. Ideally the CI file would allow you to specify which secrets are necessary and throw a clear error if one is missing. Could we approximate that with a step in the file that just checks if all required secrets are non-empty and throw an error if one is missing?

fauust:
@ArturKlauser that's a good idea, how would you do that?

I thought that the GitHub action CI message would be pretty clear if it does not find:
```yaml
DOCKER_USER: ${{ secrets.DOCKER_USER }}
DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
or would it just break later when it tries to connect to the DockerHub, I must admit I didn't check this eventuality.

ArturKlauser:
> I thought that the GitHub action CI message would be pretty clear ...

Unfortunately that doesn't seem to be the case. The assignment of the ENV from non-existent secret fails silently and the actual failure happens when you're trying to use the ENV.

https://github.com/ArturKlauser/sh/runs/1468098862?check_suite_focus=true

```
...
Login to DockerHub
Run docker/login-action@v1
Error: Username and password required
...
```

fauust:
@mvdan It should be OK now, let me know if I should target another test branch.
The only step that was not tested is the last one, "Update DockerHub description".

mvdan:
Thank you for the effort! I'll take a look this week.

fauust:
@mvdan as discussed in https://github.com/mvdan/sh/issues/631 with @ArturKlauser, I have prepared a new [branch](https://github.com/fauust/sh/tree/build-and-test-locally-first) that includes his suggestions.

If you prefer to go with a big single PR, we can forget this one and I will submit a new PR (that will introduce more changes in one shot). If you prefer multiple small PR, then I will open the new one as soon as this one is merged (but no urgency at all).

Let me know how you want to proceed...

mvdan:
I definitely prefer smaller PRs :) reviewing now.

.github/workflows/test.yml
```diff
@@ -1,8 +1,8 @@
+---
```
mvdan, line 1: can we remove this too? it seems like an artifact of the formatter you used at one point

.github/workflows/test.yml
```diff
@@ -48,19 +48,14 @@ jobs:
...
       # We use all platforms for which FROM images in our Dockerfile are
       # available.
-      DOCKER_PLATFORMS: >
-        linux/386
-        linux/amd64
-        linux/arm/v7
-        linux/arm64/v8
-        linux/ppc64le
-
+      DOCKER_PLATFORMS: linux/386,linux/amd64,linux/arm/v7,linux/arm64/v8,linux/ppc64le
```
mvdan, line 58: I found the older syntax easier to maintain to be honest; adding or deleting an item would just affect that line, and it was easy to review. I imagine this list will get longer, too.

.github/workflows/test.yml
```diff
@@ -84,18 +97,27 @@ jobs:
           # Pushes tag - deploy tag name.
           echo "TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
         fi
-        echo "DOCKER_BASE=${DOCKER_USER}/${DOCKER_REPO}" >> $GITHUB_ENV
-    - name: Install Docker buildx
+        echo "DOCKER_BASE=${{ secrets.DOCKER_USER }}/${{ env.DOCKER_REPO }}" >> $GITHUB_ENV
+    - name: Build and push
+      id: buildx
```
mvdan, line 102: why do we need this id?

mvdan (approved):
Just minor nits, LGTM.

.github/workflows/test.yml
fauust: sure (this comes from [yaml specs](https://yaml.org/spec/1.2/spec.html#id2760395) but I have no strong opinion on whether it should be used or not. As far as I understand, this is not mandatory).

.github/workflows/test.yml
fauust: Agree, I also prefer the old syntax, but this seems how it should be used: https://github.com/docker/build-push-action#multi-platform-image

I can check if the following could work if you want:
```yaml
  DOCKER_PLATFORMS: |
    linux/386,
    linux/amd64,
    linux/arm/v7,
    linux/arm64/v8,
    linux/ppc64le
```

.github/workflows/test.yml
fauust: Not used, was part of a previous test, good catch!

.github/workflows/test.yml
ArturKlauser: The way that the previous version of the code did that is by programmatically converting the multi-line (space-separated) list into the right comma-separated format `--platform ${DOCKER_PLATFORMS// /,}`. I'm not sure if you can still do that directly where you need it, since `platforms: ${{ env.DOCKER_PLATFORMS }}` probably doesn't allow shell expansion. Alternatively, you could make it part of a `run` step and put it into a separate environment variable, like you do e.g. for setting up `$DOCKER_BASE`.

fauust: Sure, and we already have a run that defines some env variables (`Set up image tag`) so, I am happy with both solutions, let me know @mvdan what you prefer.

mvdan: I do prefer keeping the newline-separated list, if it is just one extra line of shell :)

fauust: ack

fauust:
Voilà, @mvdan don't forget to create the DOCKER_USER secret and this should be ready to merge

mvdan:
Thanks for all this work! Created the DOCKER_USER=mvdan secret, and merging.

mvdan:
Also, if either of you would like to contribute to this project more actively, just let me know and I'd be happy to discuss giving write access to test CI/CD and be able to get stuff merged without requiring my approval.

fauust:
Green, cool!

Yes, I am a big fan of this project, so if I still have extra time in the next month, I would like to contribute more.

The following are some ideas:
- finalize https://github.com/fauust/sh/tree/build-and-test-locally-first;
- implement pre-commit tests like this https://github.com/fauust/docker-systemd/runs/1467025202?check_suite_focus=true.

The first one was suggested by @ArturKlauser and is almost ready (we need to verify that we are re-using the cache efficiently). The second one is because I discovered that `cmd/shfmt/docker-entrypoint.sh` is not linted by shfmt and I think pre-commit.com project add a very elegant and easy to maintain way of not missing those things.

mvdan:
> finalize https://github.com/fauust/sh/tree/build-and-test-locally-first;

sounds good!

> The second one is because I discovered that `cmd/shfmt/docker-entrypoint.sh` is not linted by shfmt and I think that the pre-commit.com project adds a very elegant and easy to maintain way of not missing those things.

I don't have anything against pre-commit per se, but I think that's a bit overkill. CI could just do something like `shfmt -w .; git diff --exit-code`, which will fail if there are any changes.

mvdan:
In any case, the offer to give access to the repo is open; just reply here or ping me in the #shell channel at [the gophers slack](https://invite.slack.golangbridge.org/).

ArturKlauser:
Thanks for the offer @mvdan, but I don't currently have time to pay enough attention here, so having you look over any changes before merging is a good thing.

fauust:
> I don't have anything against pre-commit per se, but I think that's a bit overkill. CI could just do something like `shfmt -w .; git diff --exit-code`, which will fail if there are any changes.

Sure in this particular case yes, but it depends on how many linter/verifier you want to use in a repo. I see also markdown, yaml and maybe shellcheck pb. Anyway, maybe we should discuss this in an issue or a PR to have more input from other users.

fauust:
Daniel Martí <notifications@github.com>,
10/12/2020 – 07:26:48 (-0800):

> In any case, the offer to give access to the repo is open; just reply
> here or ping me in the #shell channel at [the gophers
> slack](https://invite.slack.golangbridge.org/).

Great, so I am interested. Thanks!

mvdan:
Sent an invite :)

Issue #20, opened 2016-06-24
parse: expression interpreted as BinaryCmd, should be LetClause


mkovacs:
Relevant documentation: http://wiki.bash-hackers.org/commands/builtin/let

Here's a program to reproduce:

```
package main

import (
	"fmt"
	"reflect"
	"strings"

	"github.com/mvdan/sh"
)

func main() {
	text :=
		`let x || {
  			y=1
			echo "test"
		}`
	ast, err := sh.Parse(strings.NewReader(text), "test.sh", sh.ParseComments)
	if err != nil {
		fmt.Println("error: ", err)
	}
	fmt.Println(reflect.TypeOf(ast.Stmts[0].Cmd))
}
```

mvdan:
This is as expected. Have you looked at the `.X` part of the `BinaryCmd`?

mkovacs:
You're right, I confused myself into thinking that it's
```
let "x || {
  y=1
  echo "test"
}"
```

PR #643, opened 2020-12-18
WIP: build locally first, then push to DockerHub if tests passes

fauust:
Based on the comment from @ArturKlauser
(https://github.com/mvdan/sh/pull/638#discussion_r532089154), the author
of the first buildx CI workflow, here is the workflow suggested for
building and pushing docker images:
  - build against a local docker registry
  - test locally
  - push on DockerHub if tests passes
  - update the description on DockerHub

This is WIP because I would like to improve the multi-arch tests, see my
comment in the code (//TEMP) and also all the previous CI steps are commented to speed up the CI development.


fauust:
@ArturKlauser I would appreciate if you could take a look.

Also, FYI on skopeo, the only tricky part was to find the good options (the command is already installed on GitHub ubuntu runner):
- `--src-tls-verify=0` because it uses HTTPS otherwise (another solution is to authorize the registry in `~/.config/containers/registries.conf`);
- `--all` select all platforms;

Now regarding times, [here](https://github.com/fauust/sh/runs/1575581224?check_suite_focus=true) is a build/push with buildx.

fauust:
Also @ArturKlauser, I am not sure that the test makes sense now that we use a local registry, see https://github.com/mvdan/sh/pull/643/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R140-R143. So, we may have to find some better tests, what do you think?

fauust:
CI is red because of:
- GitHub does not find the necessary secrets;
- I am using token auth on DockerHub so Update description step can not work.

You can see the full build process on my [fork](https://github.com/fauust/sh/runs/1577417793?check_suite_focus=true).

ArturKlauser:
* The skopeo code looks good. At 41 s for the copy it's slower that I would have expected, but still faster than the build from cache you had before, so a net win.
* Why do you think the tests don't make sense anymore now? The intention was for them to test 2 things:
  * (a) An integration test making sure the shfmt binary was built correctly, and the docker image was built correctly so it can run the shfmt binary in the expected way.
  * (b) Make sure the version number reported by shfmt matches our expectation.
* Red CI: We might have a problem here. When the test runs on the PR, it fails on all secrets. I have a vague memory of secrets not being accessible for PRs due to security concerns (I could send you a PR which reads your secrets and sends them back to me). That means that PRs would never get tested properly though, which is undesirable. I think we might have to conditionally skip the steps that depend on secrets when the CI runs on a PR, so that it still runs the core of the pipeline which is the build and test part and ends up being green when those pass. Just the parts related to deployment would be skipped.

mvdan:
I'm currently almost entirely AFK due to holidays, but I'll be ramping up next week :)

fauust:
Same here!

fauust:
> The skopeo code looks good. At 41 s for the copy it's slower that I would have expected, but still faster than the build from cache you had before, so a net win.

Yes that surprised me too.

> Why do you think the tests don't make sense anymore now? The intention was for them to test 2 things:

Forget that comment, I did not read the `Dockerfile` carefully and I thought that we were using the binary from the previous CI build step. So those tests make still perfectly sense.

> Red CI: We might have a problem here. When the test runs on the PR, it fails on all secrets. I have a vague memory of secrets not being accessible for PRs due to security concerns (I could send you a PR which reads your secrets and sends them back to me). That means that PRs would never get tested properly though, which is undesirable. I think we might have to conditionally skip the steps that depend on secrets when the CI runs on a PR, so that it still runs the core of the pipeline which is the build and test part and ends up being green when those pass. Just the parts related to deployment would be skipped.

I agree, my previous version was really not good enough.
So, let's skip the last 4 steps and only execute them on that condition:
```bash
if: >
  github.event_name == 'push' &&
  (github.ref == 'refs/heads/master' ||
  startsWith(github.ref, 'refs/tags/')) &&
  github.repository == 'mvdan/sh'
```
Indeed, we don't need docker hub credentials as we now build and test containers locally.

I am still not happy about having to use the conditional skip for the last 4 steps, this is too much duplicated code. But based on https://github.community/t/skip-following-steps-of-a-job/17687, there is maybe no better approach for the moment... Also we may want to launch the docker step in parallel to speed up the CI (remove the `needs: [test, test-linux-alpine]` line).

.github/workflows/test.yml
```diff
@@ -102,58 +103,98 @@ jobs:
...
-            docker run --rm -v "$PWD:/mnt" -w '/mnt' "$image" -d myscript
+            docker run --rm -v "$PWD:/mnt" -w '/mnt' "$image" -d cmd/shfmt/docker-entrypoint.sh
           done
         done
+    - name: Check GitHub settings
+      if: >
+        github.event_name == 'push' &&
+        (github.ref == 'refs/heads/master' ||
+        startsWith(github.ref, 'refs/tags/')) &&
+        github.repository == 'mvdan/sh'
```
ArturKlauser, line 159: I'd suggest refactoring this in one of 2 ways:
 1. Like in the link you sent, where they use the complex condition to set 1 environment variable (e.g. `DO_DEPLOY`) and later just check that one variable
 2. By using the more sophisticated YAML reference syntax.
```
- name: Check GitHub settings
  if: &deploy_condition >
    github.event_name == 'push' &&
    (github.ref == 'refs/heads/master' ||
    startsWith(github.ref, 'refs/tags/')) &&
    github.repository == 'mvdan/sh'
  run: |
    ... as you've got ...
- name: Login to DockerHub
  if: *deploy_condition
  uses: ... as you've got ...
```
FYI: `yaml2js` is a useful tool to check if you've written what you meant.
```
alias yaml2js='python -c '\''import sys, yaml, json; json.dump(yaml.load(sys.stdin), sys.stdout, indent=4)'\'''
```

.github/workflows/test.yml
```diff
@@ -102,58 +103,98 @@ jobs:
         elif echo "$GITHUB_REF" | grep -q '^refs/tags/'; then
           # Pushes tag - deploy tag name.
           echo "TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+        else
+          # Use commit SHA for PR and forks
+          echo "TAG=${GITHUB_SHA::8}" >> $GITHUB_ENV
         fi
-        echo "DOCKER_BASE=${{ secrets.DOCKER_USER }}/${{ env.DOCKER_REPO }}" >> $GITHUB_ENV
+        [[ -n "${{ env.DOCKER_USER }}" ]] || DOCKER_USER="test"
```
ArturKlauser, line 110: Nit: What if you just defined `DOCKER_TEST_USER="test"` and used that instead of `DOCKER_USER` during build and test steps? Only the 'skopeo copy' really needs the actual `DOCKER_USER` on the destination repository.

ArturKlauser (approved):
Looks good, just a few minor comments.

.github/workflows/test.yml
```diff
@@ -102,58 +103,98 @@ jobs:
         elif echo "$GITHUB_REF" | grep -q '^refs/tags/'; then
           # Pushes tag - deploy tag name.
           echo "TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+        else
+          # Use commit SHA for PR and forks
```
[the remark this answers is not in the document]
fauust: True, https://github.com/fauust/sh/runs/1656942253?check_suite_focus=true :+1:

.github/workflows/test.yml
fauust: Thanks for the tip I'll take a look!

Regarding Yaml anchors, I liked it a lot, but I seems that it's not supported in GitHub Actions at the moment: https://github.community/t/support-for-yaml-anchors/16128/48.

I guess I'll try to implement 1.

mvdan:
@fauust I see you've removed WIP - is this ready to review/merge?

fauust:
Maybe @ArturKlauser can have a final review? But yes it is ready IMO.
Warning, like on my previous PR, the last step could not be tested in my fork as I am using token for Docker Hub API.

.github/workflows/test.yml
mvdan: I assume this was done with DOCKER_DEPLOY_IMAGES, marking as resolved.

I would personally stay away from yaml anchors too, because they're generally not well understood or supported.

.github/workflows/test.yml
```diff
@@ -102,58 +104,87 @@ jobs:
...
+        password: ${{ secrets.DOCKER_PASSWORD }}
+    - name: Push images to DockerHub
+      if: ${{ env.DOCKER_DEPLOY_IMAGES == 'true' }}
+      run: |
+        for ext in '' '-alpine'; do
+          image_src="${DOCKER_BASE}:${TAG}${ext}"
+          image_dst="${{ secrets.DOCKER_USER }}/${{ env.DOCKER_REPO }}:${TAG}${ext}"
+          msg="Copy multi-arch docker images to DockerHub ($image)"
+          line="${msg//?/=}"
+          printf "\n${line}\n${msg}\n${line}\n"
```
mvdan, line 183: is this printf for debugging purposes? I'm not sure I understand the need for the `//?/=` on msg, for example.

mvdan (approved):
Interesting that a local registry is needed for this. I assume you tried just "building to files", which intuitively feels like it would be simpler and possibly faster, but who knows with Docker.

.github/workflows/test.yml
fauust: This is for formatting the output (it will create a line of the size of `$msg`). I did not knew this parameter expansion but I find id quite cool (this comes from @ArturKlauser I guess).

mvdan: Oh wow :P I think I'm impressed?

ArturKlauser: You're welcome ;)

ArturKlauser (approved):
looks good to me

fauust:
@mvdan you are maybe expecting me to merge it... but for now, I would prefer if you can do it so you can rollback in case something goes wrong.

May I suggest you to control that everything looks fine on dockerhub once it's merged and the CI is finished? I will also double check when this PR is merged.

mvdan:
We got a failure, but it seems to be a network flake:

> #20 30.19 go: github.com/creack/pty@v1.1.11: Get "https://proxy.golang.org/github.com/creack/pty/@v/v1.1.11.mod": dial tcp 142.250.73.209:443: i/o timeout

mvdan:
Failed with another network issue. Sigh. Let's kick a retry in a couple of hours, because I think their network is weird at the moment.

fauust:
Maybe because it's friday :)

fauust:
@mvdan all green and I have just checked that I am able to pull and use the new images.

PR #646, opened 2020-12-26
interp: LookPathDir

zimbatm:
The main motivation is to make the cwd more explicit. Previously it was implied by assuming that a "PWD" environment variable exists.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -389,7 +388,7 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			last = 0
 			if r.Funcs[arg] != nil || isBuiltin(arg) {
 				r.outf("%s\n", arg)
-			} else if path, err := exec.LookPath(arg); err == nil {
+			} else if path, err := LookPath(r.Dir, expandEnv{r}, arg); err == nil {
 				r.outf("%s\n", path)
 			} else {
 				last = 1
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -229,7 +229,7 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 		args := fp.args()
 		for _, arg := range args {
 			if mode == "-p" {
-				if path, err := LookPath(expandEnv{r}, arg); err == nil {
+				if path, err := LookPath(r.Dir, expandEnv{r}, arg); err == nil {
 					r.outf("%s\n", path)
 				} else {
 					anyNotFound = true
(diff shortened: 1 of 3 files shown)


zimbatm:
there is another instance of `exec.LookPath` in https://github.com/mvdan/sh/blob/b944d98009705cc63ef0ff9680b85701802c3db7/interp/test.go#L165 which I don't really know if it should be removed or not

mvdan:
Could you clarify if this fixes a bug? If so, please add a test case. If not, I'm not sure if this is a refactor for the sake of clarity, or something else.

Also, I don't think we can alter LookPath because of backwards compatibility. I think relying on env's PWD is perhaps not perfect, but it works :) We could always reconsider this API design for a future v4 if we think it was a mistake.

zimbatm:
Sorry for the lack of clarity.

Regarding the PWD argument, the issue is that this environment variable is not guaranteed to be set. It works *most* of the time, but not always. The reason is that PWD is set by the shell, but not other programs. For example, in Go you can change the working directory with `exec.Command` and it will not update the `PWD` environment variable. Another good example is this:
```console
$ cd /
$ cd /tmp
$ env | grep PWD
OLDPWD=/
PWD=/tmp
$ sudo env | grep PWD
<no output>
```

For a bit more context, this all started as a rabbit hole while trying to fix this security issue for direnv: https://github.com/golang/go/issues/38736 (this is fixed by 6a1ba16 ). Since direnv is dealing with environment variables, I wanted a version of LookPath that can take those as arguments, and started implementing something similar at https://github.com/direnv/go-lpenv . Then I took a look at this project because I was curious if you also suffered the security issue and discovered that LookPath almost does what I want.

zimbatm:
Since API compatibility is important, I propose to rename the implementation to `LookPathCwd`, and then have `LookPath` be the backward-compatible version. Would that be acceptable to you?

mvdan:
> The reason is that PWD is set by the shell, but not other programs.

Hm. I would normally say "this is the caller's fault as we document the API well", but I agree that it's a design that lends itself to misuse.

In fact, the API only ever needs three environment variables - PWD, PATH, and PATHEXT - the last one only on Windows. So we could instead add a func that takes `dir, path, pathext, file string` instead of `dir string, env expand.Environ, file string`.

Or, we could even go further and hide the complexity of the string-encoded path lists, using `dir string, path, pathext []string, file string`. Then it's less shell-like, and more Go-like.

> Since API compatibility is important, I propose to rename the implementation to `LookPathCwd`, and then have `LookPath` be the backward-compatible version. Would that be acceptable to you?

If I could name the funcs from scratch, I'd call the new one `LookPath` and the existing one `LookPathEnv`. Not an option, though.

I'm not a big fan of cwd/pwd in the function name or as a parameter name since they are shell acronyms that many Go developers might not be familiar with. So I would prefer something more intuitive like `LookPathDir` or `LookPathFrom`, which seem to imply that they are more explicit about the parameters.

zimbatm:
Thanks for taking a second look! :) I renamed the function to `LookPathDir`.

Regarding the function interface, all of these are possible. The reason I like the proposed one is that it just lifts the env and dir. In the stdlib exec.LookPath implementation, those are taken from the process and are effectively global variables. It's also the same parameters you would pass to an `exec.Command` (well the env is a bit different but that's another story).

mvdan (approved):
Sounds good to me. We should clean up these APIs for v4 anyway, and I agree that sticking to something similar to os/exec is sensible.

mvdan:
I did not notice the multiple commits - FYI I'll squash them manually and force-push to master.

zimbatm:
sounds good, thanks!

PR #647, opened 2020-12-27
interp: implement `type -t`

zimbatm:


mvdan (approved):
Nice :) Do you want me to squash-merge this as a single commit, or rebase-merge and keep the original commit messages? Each of the commit messages has useful info, but I feel like a single commit is better than four for what is essentially a small and pretty self-contained change.

zimbatm:
That's really up to you if you want to squash, rebase or merge. The self-contained commits that I created are there to make the review easier.

PR #654, opened 2021-01-07
interp: Fix panic when trying to `read` and stdin is `nil`

andreynering:


interp/builtin.go
```diff
@@ -719,6 +720,11 @@ func (r *Runner) printOptLine(name string, enabled bool) {
 }
 
 func (r *Runner) readLine(raw bool) ([]byte, error) {
+	if r.stdin == nil {
+		r.errf("can't read, there's no stdin\n")
```
mvdan, line 724: if we're returning an error, I don't think we want to print one as well

interp/interp_test.go
```diff
@@ -3548,3 +3548,21 @@ func TestMalformedPathOnWindows(t *testing.T) {
 		t.Fatalf("wrong output:\nwant: %q\ngot:  %q", want, got)
 	}
 }
+
+func TestReadShouldNotPanicForEmptyStdin(t *testing.T) {
```
mvdan, line 3552: s/ForEmptyStdin/WithNilStdin/ maybe?

interp/interp_test.go
```diff
@@ -3548,3 +3548,21 @@ func TestMalformedPathOnWindows(t *testing.T) {
...
+
+	f := parse(t, nil, "read foobar")
+	if err := r.Run(context.Background(), f); err == nil {
+		t.Fatal("it should have retuned an error")
+	}
+
+	want := "can't read, there's no stdin\n"
+	if got := b.String(); got != want {
+		t.Fatalf("wrong output:\nwant: %q\ngot:  %q", want, got)
+	}
```
mvdan, line 3567: if you remove the error printing, then this part of the test can go as well

interp/builtin.go
andreynering: Thanks for the blazing fast review!

mvdan (approved):
Thanks for the fix!

mvdan:
Sigh, it appears one of the mac machines got stuck.

PR #660, opened 2021-01-27
Add mdformat-shfmt to readme integrations section

hukkinj1:
Hi!

I integrated `shfmt` with the [mdformat](https://github.com/executablebooks/mdformat) Markdown formatter. The idea is that when formatting Markdown, any embedded code blocks tagged with `sh` or `bash` are also formatted with `shfmt`.

I noticed you have a README section for integrations so thought you might be interested in adding this? I understand if there's some criteria (adoption, GitHub stars... ?) for this integrations list that might prevent doing so 😄


mvdan (approved):
I think it's reasonable :) Thanks!

Issue #672, opened 2021-02-24
expand.Variable cannot represent an indexed array that starts at non-zero


twilly:
A Bash arrays can start at any index, though it counts the number of set elements as part of the length.

For example, Bash says:
```
$ x=([1]=a); echo ${#x[@]}
1
```

gosh says:

```
$ x=([1]=a); echo ${#x[@]}
2
```

`expand.Variable.List` being a `[]string` is not flexible enough to describe these bespoke arrays. A zero string is a valid expansion after all.

mvdan:
Yikes, I wasn't aware that Bash indexed arrays are... not actually arrays. Here's another example:

```
$ empty=("" "foo") missing=([1]="bar")
$ echo "${empty[@]}" "${#empty[@]}"
 foo 2
$ echo "${missing[@]}" "${#missing[@]}"
bar 1
```

You can even have gaps anywhere in the array, not just the beginning:

```
$ middle=([1]="one" [3]="three")
$ echo "${middle[@]}" "${#middle[@]}"
one three 2
```

The question then becomes - how do we represent indexed arrays in an efficient way? We could give up entirely and use a `map[int]string` or the existing `map[string]string` directly, but as you can guess, maps have an inherent overhead when compared to a plain old slice.

Maybe the answer is just that indexed arrays in Bash are practically as costly as associative arrays, since both seem to behave like maps.

mvdan:
I'm trying to think if we could possibly fix this in v3 without breaking programs. I don't think so, because the Variable.List field is exposed directly as a field instead of a method.

mvdan:
There's one hack we could possibly do in v3. Fill the "gaps" with string values which are impossible to have in shell, such as a single null/zero byte:

```
$ a=$'\x00'
$ echo "${a} ${#a}"
 0
```

Then, the length of an indexed array means looping over List to count the non-null elements. Similarly, expansion would have to skip over such gaps.

This is a hack, but I think it should be backwards compatible - any direct users of `expand.Variable` would have their behavior changed for arrays with gaps, but those didn't work properly to begin with. So no breakages should happen in practice.

I also wonder if this hack is actually better in the long run than alternatives like `map[int]string` or `[]struct{idx int; val string}`. Assuming that gaps are rare, and that one would rarely leave immense gaps like `arr=([10000]=foo)`, then it makes sense to keep a contiguous slice representation.

twilly:
C strings strike again. Null as a sentinel value makes sense.

[...]

PR #673, opened 2021-02-24
interp: expand array arguments

### LINKED ISSUE #671: interp: array elements should be expanded
Bash array elements are word expanded, except when an index is provided.

Expected result:
```
$ x=($(echo a b)); echo ${x[0]}
a
```
gosh's result:

```
$ x=($(echo a b)); echo ${x[0]}
a b
```

I'll be posting a PR to fix this soon.

twilly:
Bash array values are generally expanded. For example:

```
$ x=($(echo a b)); echo ${x[0]}
a
```

The exception is when an index is provided. Those values are taken
literally.

```
$ x=([0]=$(echo a b)); echo ${x[0]}
a b
```

The runner expands values in `assignVal` and did so by mapping the
expression index to the array index. A fixed mapping fails when a value
generates more than one field. Hence the code interpreted all array
values as literal.

This change iterates over the elements, expanding literally or with
fields, while keeping track of the implicit index. The result is then
flattened down at the computed indexes.

Fixes #671

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -388,6 +388,14 @@ var runTests = []runTest{
 	{"a=bar; echo ${#a} ${#a[@]}", "3 1\n"},
 	{"a=世界; echo ${#a}", "2\n"},
 	{"a=(a bcd); echo ${#a} ${#a[@]} ${#a[*]} ${#a[1]}", "1 2 2 3\n"},
+	{
+		"a=($(echo a bcd)); echo ${#a} ${#a[@]} ${#a[*]} ${#a[1]}",
+		"1 2 2 3\n",
+	},
+	{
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Thank you for digging into this and providing a thoughtful fix. It's perhaps unfortunate that we allocate twice, and I imagine there's some room for saving cost in some small and common scenarios. But I'm not too worried right now, and correct is more important than fast.

twilly:
> Thank you for digging into this and providing a thoughtful fix. It's perhaps unfortunate that we allocate twice, and I imagine there's some room for saving cost in some small and common scenarios. But I'm not too worried right now, and correct is more important than fast.

Exactly. I'd like to generate less garbage but I stopped short when #672 cropped up to suggest a bigger refactor may be in store.

PR #676, opened 2021-02-25
interp: implicit casting of arrays

### LINKED ISSUE #675: interp: implicit casting of arrays fails
A variable being assigned an array type value should be type cast to the array type.

For example, this should print "b":

```bash
declare a; a+=(b); echo ${a[@]}
```
Strings were covered but the unset variable case was not.

twilly:
`declare a` would create an unset variable of an unknown type. `a`
A variable should be type casted to an array when an array typed
assignment occurs. String typed variables being cast to arrays was
covered but missed the unset case.

Fixes #675


twilly:
@mvdan something to consider is a default panic as missing a type conversion here is a clear programming bug.

interp/interp_test.go
```diff
@@ -556,6 +556,10 @@ var runTests = []runTest{
 		`a='"\n'; printf "%s %s" "${a}" "${a@E}"`,
 		"\"\\n \"\n",
 	},
+	{
+		"declare a; a+=(b); echo ${a[@]}",
```
mvdan, line 560: add `{#a[@]}` too; I'm not sure if the code should result in 1 or 2 array elements, and if we got it wrong in the code we wouldn't notice.

mvdan (approved):
> @mvdan something to consider is a default panic as missing a type conversion here is a clear programming bug.

That sounds good to me, too. These Kind switches should generally panic on an unknown value.

interp/interp_test.go
twilly: And it's a real issue in this PR. Using prev.Str on an unset makes the length two (or if we fix #672). I'll special case this conversion since it's pretty common to declare and then add arrays in a loop.

twilly:
Changes implemented! I'm not sure about your policy for approval stickyness, so I'll wait for you to reapprove or land. :)

mvdan (approved):

PR #677, opened 2021-02-25
interp: local vars are mutable

### LINKED ISSUE #674: interp: local variables are immutable
f4c774aa15046ef006508e182fde10c4b56876fa introduced a regression where local variables are immutable.

`f() { local a; a=bad; a=good; echo $a; }; f` should print "good" but instead prints "bad." I pushed a test case at https://github.com/mvdan/sh/compare/master...twilly:bug-locals-are-immutable.

twilly:
Local variables should be mutable but f4c774a accidentally made them
immutable. A new assignment is made with a new `expand.Variable` struct,
which defaults the Local flag to false. Failing to preserve the
localness of an overlayed variable means future sets are incorrectly
forwarded to the global scope.

Fixes #674

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -564,6 +564,10 @@ var runTests = []runTest{
 		`a=""; a+=(b); echo ${a[@]} ${#a[@]}`,
 		"b 2\n",
 	},
+	{
+		"f() { local a; a=bad; a=good; echo $a; }; f",
+		"good\n",
+	},
 
 	// if
 	{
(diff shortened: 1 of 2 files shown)


twilly:
Note: this will conflict with pr #676. It's here for review purposes and I'll rebase with conflict resolution on this or the other pr.

mvdan:
Unfortunate that GitHub does not support "chains" of PRs, really. But I do appreciate keeping to one commit per PR.

twilly:
Rebased and conflict free. :)

mvdan (approved):

PR #687, opened 2021-03-19
Add Bash 5.1 parameter transformations

lollipopman:
Per the NEWS:

dd. New `U', `u', and `L' parameter transformations to convert to uppercase,
    convert first character to uppercase, and convert to lowercase,
    respectively.

hh. New `K' parameter transformation to display associative arrays as key-
    value pairs.

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -2421,7 +2421,7 @@ var fileTests = []testCase{
 	},
 	{
 		Strs: []string{`${a@E} ${b@a} ${@@Q} ${!ref@P}`},
-		bsmk: call(
+		bash: call(
 			word(&ParamExp{
 				Param: lit("a"),
 				Exp: &Expansion{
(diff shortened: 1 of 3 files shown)


lollipopman:
I am happy to rework this with tests and excluding mksh if needed, just wasn't sure what direction you would prefer?

mvdan:
Yeah, tests would be good - there should already be tests covering the other transformations, so you could copy-paste a bit.

lollipopman:
> Yeah, tests would be good - there should already be tests covering the other transformations, so you could copy-paste a bit.

Will do, do you want me to prevent them working on mksh?

mvdan:
Up to you - I think the current code is slightly wrong too, as mksh doesn't seem to support some older bash transformations like @E. We need tests either way, though.

lollipopman:
tests added, let me know if you want any further improvements, thanks!

lollipopman:
hmm, I can't reproduce this test failure on Debian sid, I'll need to spend some more time to debug it

mvdan:
The tests which run the real bash/mksh/etc require a specific major version, since older or newer versions behave differently. If a different version is installed, they are skipped.

Error inputs can have the ` #NOERR` trailing string to signal that our parser errors, but a real shell's does not. You will need that here, because for better or worse, bash/mksh/etc do not fully parse parameter expansions until they are executed.

lollipopman:
> The tests which run the real bash/mksh/etc require a specific major version, since older or newer versions behave differently. If a different version is installed, they are skipped.
> 
> Error inputs can have the ` #NOERR` trailing string to signal that our parser errors, but a real shell's does not. You will need that here, because for better or worse, bash/mksh/etc do not fully parse parameter expansions until they are executed.

Thanks @mvdan that would have taken me quite a while to suss out on my own. I think everything is in place now. Perhaps it would be valuable to add a `TESTING.md` or similar doc to provide guidance on testing? Until you explained that the tests were being skipped I didn't realize that my local testing wasn't exhaustive. It would be great to be able to run all tests locally, perhaps via docker containers? Thanks again

mvdan:
Indeed, you're right. A bunch of people have contributed test cases in the past already, so I'm overdue for documenting this a bit.

mvdan (approved):
Thank you!

PR #689, opened 2021-03-28
Add -bn to flags for Google's shell style

mikepqr:
You need `-bn` to get Google style with shfmt. This commit changes the
man page to reflect that.

Google's shell style guide says long pipelines:

    should be split at one pipe segment per line with the pipe on the
    newline and a 2 space indent for the next section of the pipe. This
    applies to a chain of commands combined using | as well as to
    logical compounds using || and &&.

        # All fits on one line
        command1 | command2

        # Long commands
        command1 \
          | command2 \
          | command3 \
          | command4

See https://google.github.io/styleguide/shellguide.html#s5.3-pipelines


mvdan (approved):
Thanks! Nicely spotted.

PR #717, opened 2021-07-26
Add note about what to check when executable does not seem to work

usrme:
First off, thanks for all the hard work you've done with this project! Second, since I am not a Go programmer I ran into some minor troubles installing the package using the example under "Quick start". Hopefully those few lines can save a bit of time for other newcomers as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ viewed directly as Markdown or rendered with [scdoc].
 Packages are available on [Alpine], [Arch], [Docker], [FreeBSD], [Homebrew],
 [MacPorts], [NixOS], [Scoop], [Snapcraft], [Void] and [webi].
 
-If the above command does not seem to yield a working executable, then make sure the `GOBIN` environment variable is set to whatever you want it to be or that either `$GOPATH/bin` or `$HOME/go/bin` are in the `PATH` environment variable. For more information, see [the official documentation for 'go get'](https://golang.org/ref/mod#go-get).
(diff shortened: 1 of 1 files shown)


mvdan:
Thanks for this PR! I have mixed feelings about extra instructions like these. Should every README telling people to `go get` or `go install` also tell people how to use Go? I tend to think not, because it's generally going to be repetitive and just extra text that's not particularly specific to this project :)

I wouldn't mind a very short sentence pointing to clear instructions on how to use the Go tool with $PATH, but I don't think such an upstream doc exists.

usrme:
I wholly agree with you about not being repetitive and that was one of the reasons I was somewhat hesitant in making this PR. How would you feel about the following sentence in replacement of what I've committed?

> If the above `go get` command does not seem to yield a working executable, then verify the [relevant path environment variables](https://zchee.github.io/golang-wiki/GOPATH/).

mvdan:
I actually forgot I wrote https://play-with-go.dev/installing-go-programs-directly_go116_en/ some time back, which I think would be a good place to link to. I was planning on replacing `go get` with `go install` once Go 1.17 comes out, for the sake of not breaking the README instructions for Go 1.15.x too early. When I do that, I'll also add a line to link to that interactive guide :)

usrme:
Awesome, that's good to hear! I'll close this PR then if a change is by you is upcoming anyway.

Issue #551, opened 2020-04-25
interp: Support more flags on the "read" builtin


andreynering:
Originally asked on https://github.com/go-task/task/issues/314.

There are some missing options on the `read` builtin. The `-p` one seems particularly useful on scripts.

https://github.com/mvdan/sh/blob/b7aede07e8af0320deb8b5f835bb1df6ec59face/interp/builtin.go#L473-L479

mvdan:
Sure - patches welcome :)

blackjid:
I think is already supported. 

I wonder if the `read -s` silent would be possible to be able to read password or secrets.

mvdan:
`read -a` was done in 66b52698a5815d7cdd9d167e264af6644d552f5b.

PR #722, opened 2021-08-17
Read Prompt Support

parkerduckworth:
Hey @mvdan 👋 

I have implemented the `read -p` flag, as requested in issue #551, and included new test cases to cover the change, as well as the integration of `-p` with the existing code that handles `read`.  All test cases old and new are passing.

Let me know if there are any changes still needed, and thanks for your hard work maintaining this library!

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -519,11 +519,14 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 		r.setErr(returnStatus(code))
 	case "read":
 		raw := false
+		prompt := struct{provided bool; message string}{}
 		fp := flagParser{remaining: args}
 		for fp.more() {
 			switch flag := fp.flag(); flag {
 			case "-r":
 				raw = true
+			case "-p":
+				prompt.provided, prompt.message = true, fp.value()
 			default:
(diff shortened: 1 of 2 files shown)


interp/builtin.go
```diff
@@ -538,6 +544,14 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			}
 		}
 
+		if prompt.provided {
+			if prompt.message == "" {
```
mvdan, line 548: you could make this check inside `case "-p"`, then you don't need `provided` at all - you'd just do `message != ""`.

and you could also just do `message := ""`; I don't think the struct is necessary

mvdan (approved):
Appreciated! Just one comment to simplify the flag parsing logic.

interp/builtin.go
parkerduckworth: Ok gotcha, making the change 👍

parkerduckworth:
@mvdan I left the `if` statement checking to see if `prompt` was provided before writing to stdout.  Although this check is not strictly necessary, it seems worth it to prevent the unnecessary IO, but can remove it if you prefer.

mvdan (approved):

Issue #514, opened 2020-02-23
Interp: Set GID


jjzcru:
I know that you can get the PID from the terminal but there is a way to make the process that is going to work with an specificy GID (Group ID)?

Cheers

mvdan:
Sorry, I don't get what you're trying to do.

mvdan:
Closing for now.

sylv-io:
Hello there,

I guess I understand the issue. Analog to bash, only `UID` is provided as environment variable, but not `GID`.
Other shells like zsh does also export `GID`.

This issue can be solved with the following patch:

```diff
diff --git a/interp/api.go b/interp/api.go
index 6ac3dc34..b7526c56 100644
--- a/interp/api.go
+++ b/interp/api.go
@@ -444,6 +444,13 @@ func (r *Runner) Reset() {
                        Str:      strconv.Itoa(os.Getuid()),
                })
        }
+       if !r.writeEnv.Get("GID").IsSet() {
+               r.setVar("GID", nil, expand.Variable{
+                       Kind:     expand.String,
+                       ReadOnly: true,
+                       Str:      strconv.Itoa(os.Getgid()),
+               })
+       }
        r.setVarString("PWD", r.Dir)
        r.setVarString("IFS", " \t\n")
        r.setVarString("OPTIND", "1")
```

PR #723, opened 2021-09-06
interp/api: set GID

sylv-io:
Set GID environment variable analog to UID.

Rel: https://github.com/go-task/task/issues/561

Signed-off-by: Marcello Sylvester Bauer <sylv@sylv.io>

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -444,6 +444,13 @@ func (r *Runner) Reset() {
 			Str:      strconv.Itoa(os.Getuid()),
 		})
 	}
+	if !r.writeEnv.Get("GID").IsSet() {
+		r.setVar("GID", nil, expand.Variable{
+			Kind:     expand.String,
+			ReadOnly: true,
+			Str:      strconv.Itoa(os.Getgid()),
+		})
+	}
 	r.setVarString("PWD", r.Dir)
 	r.setVarString("IFS", " \t\n")
 	r.setVarString("OPTIND", "1")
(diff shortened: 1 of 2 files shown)


sylv-io:
Related Issue:  #514

sylv-io:
cc: @jjzcru

sylv-io:
cc @aCLr

sylv-io:
> Thanks! Can you add a very simple test to check that GID isn't empty? Near where we have the UID tests. Something like `[[ -z $GID ]] && echo "GID not set"` expecting the output to be empty.

I had to add `#IGNORE` since bash does not set GID by default. Is this PR still a desirable adaptation or should it be 1 to 1 bash complaint? I personally think it's great addtion.

mvdan:
Oh, right, Bash sets UID but not GID. I was under the assumption it set both.

On the other hand, Zsh does set it: https://askubuntu.com/questions/1124674/why-is-my-gid-environment-variable-empty

And I think it's probably harmless if we provide extra built-in information. The interpreter tends to default to bash behavior, but it's also not just for interpreting bash - ideally it should also do an OK job at interpreting POSIX shell, mksh, etc. mksh has KSHGID too, for example.

So, sounds good to me :)

mvdan (approved):

mvdan:
Ah, your last-minute push confused me while merging. Are you done with the changes?

mvdan:
Also, please squash your commits, as otherwise I'd squash-merge.

sylv-io:
> Ah, your last-minute push confused me while merging. Are you done with the changes?

sorry, i did not see you answer in time. i will squash and only keep the readonly checks.

mvdan:
I can also squash of course, but best if you write your own unified commit message :)

sylv-io:
> I can also squash of course, but best if you write your own unified commit message :)

already done :). I removed the stat test, since it isn't a shell buildin. :+1:

mvdan (approved):

PR #727, opened 2021-09-08
syntax: add a Quote func

mvdan:
(see commit message)


andreynering (approved):
Seems great to me!

mvdan:
Thanks for the review! I extended the fuzz func a bit, and it's now incapable of finding cases that we don't handle. In the process I also taught Quote to avoid issues with comments and assignments.

PR #737, opened 2021-09-29
syntax: add a LangVariant parameter to Quote

mvdan:
(see commit message)

### DIFF
--- a/syntax/lexer.go
+++ b/syntax/lexer.go
@@ -5,11 +5,7 @@ package syntax
 
 import (
 	"bytes"
-	"fmt"
 	"io"
-	"strconv"
-	"strings"
-	"unicode"
 	"unicode/utf8"
 )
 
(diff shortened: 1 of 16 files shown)


mvdan:
cc @andreynering @pwaller if you're up for a review :)

syntax/lexer.go
```diff
@@ -1193,6 +1193,11 @@ func Quote(s string) (_ string, ok bool) {
 			return "", false
 		}
 		if r == utf8.RuneError || !unicode.IsPrint(r) {
+			if lang == LangPOSIX {
+				// POSIX shell lacks $'.
+				// Without it, we can't use escape sequences such as \x.
+				return "", false
```
andreynering, line 1199: Wouldn't it be useful to return an error instead of a boolean? Then if this failed, people would have a message explaining why.

The same applies to other possible errors.

syntax/testdata/fuzz/FuzzQuote/25f36feab4af00bc4dfc3cf56da02b842b62ba8c5ac44862b5b3b776a0d519b4
```diff
@@ -0,0 +1,3 @@
+go test fuzz v1
```
andreynering, line 1: It's not clear for me if these files are really supposed to be here or were just forgotten.

andreynering (approved):

syntax/lexer.go
mvdan: What would you do with an error value?

Having some information like the offset might be useful, but beyond that I'm not sure.

syntax/testdata/fuzz/FuzzQuote/25f36feab4af00bc4dfc3cf56da02b842b62ba8c5ac44862b5b3b776a0d519b4
mvdan: Yup they are meant to be left behind: they serve as fuzz seed corpus entries, and ran as tests too.

syntax/lexer.go
andreynering: It's more about helping the user to understand what gone wrong.

"syntax: it is not possible to quote \x in POSIX mode" is more useful than "false" when trying to understand what happened.

Just a suggestion, though. Likely not a big deal to keep as is if you want. 🙂

mvdan: Yeah that's fair. The signature is from when we only returned false on null bytes, so back then it made less sense to use an error. I'll add one.

mvdan: Done.

mvdan:
Thanks for the review :) Added more tests, moved the code to quote.go, and I've added a QuoteError type.

mvdan:
One last pre-merge change: QuoteError.RuneOffset is now ByteOffset, because a rune offset isn't particularly helpful on its own. It's not even easy to slice the string to get to the concerning rune quickly, but you can do that with a byte offset.

Issue #738, opened 2021-10-02
interp: randomize not found name in tests


creack:
Very nit-picky, I just happened to have a binary called `foo` in my path, causing the tests to fail. It might be a good idea to randomize the name when checking for something not found or to use a more lengthy, unlikely-to-be-used name.

```
--- FAIL: TestRunnerRun (0.28s)
    --- FAIL: TestRunnerRun/#641 (0.01s)
        interp_test.go:2981: wrong output in "alias foo='bar baz'\ntype foo":
            want: "type: foo: not found\nexit status 1"
            got:  "foo is /home/creack/go/bin/foo\n"
```

mvdan:
Oops, thanks for catching this :)

PR #741, opened 2021-10-10
Randomise not found name in tests

stoovon:
- If tests looking for e.g. "foo not found" then add a random int64 suffix to foo.
- In the event that someone has an executable with the same name in their path, such that a test that should fail will pass, then re-running the tests will reveal this to them.

Fixes https://github.com/mvdan/sh/issues/738.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"math/bits"
+	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


mvdan:
Friendly ping @stoovon :)

stoovon:
Thanks! Will take a look

stoovon:
Closing and re-opening under https://github.com/mvdan/sh/pull/821.

PR #745, opened 2021-10-17
Fix a problem with numbered environment variables

jespino:
I found this problem here: go-task/task#586

I traced it to here. For some reason, the sorting + the binary search, were
behaving wrong. The inclusion of the `=` sign in the ordering looks like fixes
the problem. I'm not 100% sure why, maybe is again, an upstream bug.

I added a unit test to reproduce the problem.


mvdan:
I'll merge on green, using a squash-merge to include the explanation of the fix in the commit message :)

PR #752, opened 2021-11-05
Add example for docker-based pre-commit hook

wwuck:
This adds an example for running shfmt in a local pre-commit hook using the official shfmt docker image for cases where golang is not installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,19 @@ repo configuration like:
         args: [-w]
         types: [shell]
 ```
+An alternative configuration for running shfmt via the official shfmt docker image is:
+```yaml
+  - repo: local
+    hooks:
+      - id: shfmt
+        name: Format Shell Script Files
+        description: Runs shfmt Docker image to format Shell Script files
+        minimum_pre_commit_version: 2.4.0
+        language: docker_image
+        types: [shell]
+        entry: mvdan/shfmt
+        args: [-w]
+```
 
 ### Related projects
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,6 +103,19 @@ repo configuration like:
         args: [-w]
         types: [shell]
 ```
+An alternative configuration for running shfmt via the official shfmt docker image is:
```
mvdan, line 106: please add empty lines around this paragraph for readability

README.md
```diff
@@ -103,6 +103,19 @@ repo configuration like:
...
+```yaml
+  - repo: local
+    hooks:
+      - id: shfmt
+        name: Format Shell Script Files
```
mvdan, line 111: use the same name as the other example, for consistency

README.md
```diff
@@ -103,6 +103,19 @@ repo configuration like:
...
+        description: Runs shfmt Docker image to format Shell Script files
```
mvdan, line 112: also for consistency with the other example, remove the description, as IMO it's a bit redundant

mvdan (commented):
How does this pin the version of shfmt itself? I assume this follows `latest`, which is not a good idea, as that will change over time. Ideally it would pin a specific version, just like the other example uses `@v3.2.2` in `additional_dependencies`.

wwuck:
Yes, I should add a pin on specific tag in the example.  Anyone using this would have to upgrade the tag manually for each new release as `pre-commit upgrade` won't do that automatically here.

mvdan (approved):
Thanks!

PR #759, opened 2021-11-17
interp: fix variables scope in functions

cclerget:
Variables declared as exported or readonly in functions
without assignment are not correctly propagated to the
global scope, this snippet should return a `bar: readonly variable`
error:

```
_foo() {
    bar=foo
    readonly bar
}

_foo
bar=bar
```

This snippet should also export `bar` variable correctly:

```
_foo() {
    export bar
}

_foo
bar=foo
env | grep ^bar=
```

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -2304,6 +2304,10 @@ set +o pipefail
 	{"export foo=(1 2); $ENV_PROG | grep '^foo='", "exit status 1"},
 	{"declare -A foo=([a]=b); export foo; $ENV_PROG | grep '^foo='", "exit status 1"},
 	{"export foo=(b c); foo=x; $ENV_PROG | grep '^foo='", "exit status 1"},
+	{"foo() { bar=foo; export bar; }; foo; $ENV_PROG | grep ^bar=", "bar=foo\n"},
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Ah, this sounds like an unintended regression from the scoped vars refactor back in February, in f4c774aa15046ef006508e182fde10c4b56876fa. That change did improve multiple edge cases and add more tests, but evidently I missed a few edge cases.

The tests and fix itself looks great, as always :) I can't say that the changes to the "Set" logic are immediately obvious, but they seem reasonable and I trust the tests.

mvdan:
FYI, it is likely I'll cherry-pick some of the important fixes into a v3.4.1 in a week or so. I assume that pulling master is fine in the meantime; there haven't been any large or breaking changes since then.

mvdan:
I've started the cherry-picking with what we've got so far at https://github.com/mvdan/sh/tree/v3.4, if you want to use the HEAD of that branch rather than master, which is now the start of v3.5.

cclerget:
> I've started the cherry-picking with what we've got so far at https://github.com/mvdan/sh/tree/v3.4, if you want to use the HEAD of that branch rather than master, which is now the start of v3.5.

Great thanks :+1: :)

Issue #705, opened 2021-05-19
interp: builtins skip the exec handler


TorchedSammy:
im sure that this is definitely intentional, but this is actually undesirable in my use case
im developing [hilbish](https://github.com/Hilbis/Hilbish) and use this module to run sh of course, but used to only use this as a last option fallback instead of running my custom builtins or lua code.
but now im considering actually using an exec handler for the runner but realized `cd` wasnt behaving how i expected it to be since it's this module's cd and not mine

so is there a way for me to change how the builtin works from outside the package? i havent found anything in docs like that

mvdan:
You're right in assuming this is not supported, right now.

The question is - if you replaced builtins like `cd`, how could our runner actually be aware of when the current directory changes? Many other builtins change the internal state of the runner, like `set` and `exit`, so the same applies to them.

So I struggle to imagine what an interface would look like. If it were truly complete, it would expose a lot of internals of the runner, at which point it stops being a useful interface :)

What are you trying to do with your custom builtins? If I understand your goal and use case better, maybe that's going to help suggest or design something we can actually pull off.

TorchedSammy:
my goal, or what i want here is so that i can be able to define my shell's builtins in lua for easier changes, and that i can add any other enhancements whenever, and so i can send my hooks/events whenever something happens (example being the cd command throws a cd hook)

mvdan:
To make sure I understand correctly - you want to run extra code when a builtin is run, but not replace the default behavior, right?

Would you want to run that extra code before, or after the built-in behavior?

TorchedSammy:
oh no, i do want to replace it
> able to define my shell's builtins in lua for easier changes, and that i can add any other enhancements whenever

mvdan:
Right, and how would you implement that? If you entirely replace the runner's `cd`, then the runner will never actually change its working directory. If you replace the runner's built-ins, you're half way to replacing the entire runner.

TorchedSammy:
i actually have no idea, what do u propose i do by myself instead? i only reported what happened, i have no idea how the runner works

mvdan:
I'd be able to propose something if I clearly understood what you're trying to do.

A) If you're trying to reimplement the runner's builtins yourself, then you're basically reimplementing a big chunk of the runner. I don't see how we could ever support this well. A "handler" interface to implement all builtins externally would be huge.

[...]

PR #766, opened 2021-11-29
builtin(cd): Support hyphen-minus operand

donorp:
Hi, thanks for this amazing package, I'm relying on it for cross platform build automation tasks.

And I came across the unexpected behavior of `cd -`, which errored when there is no directory named `-`, instead of chdir into last working directory.

In this pull request, the existence of directory named `-` in the current working directory is not checked as the man page says `cd -` is actually an alias of `cd ${OLDPWD}`, so to my understanding, a directory named `-` is ignored in this case (and I can confirm ash, zsh, bash on macos and linux all have this behavior).

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -214,6 +214,12 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			path = r.envGet("HOME")
 		case 1:
 			path = args[0]
+
+			// replicate the commonly implemented behavior of `cd -`
+			// ref: https://www.man7.org/linux/man-pages/man1/cd.1p.html#OPERANDS
+			if len(path) == 1 && path[0] == '-' {
+				path = r.envGet("OLDPWD")
+			}
 		default:
 			r.errf("usage: cd [dir]\n")
 			return 2
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -217,7 +217,7 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 
 			// replicate the commonly implemented behavior of `cd -`
 			// ref: https://www.man7.org/linux/man-pages/man1/cd.1p.html#OPERANDS
-			if len(path) == 1 && path[0] == '-' {
+			if path == "-" {
 				path = r.envGet("OLDPWD")
 			}
 		default:
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -929,7 +929,7 @@ var runTests = []runTest{
 		"1\n",
 	},
 	{
-		`orig="$PWD"; mkdir a && cd a && cd - && [[ "$PWD" == "$orig" ]]`,
+		`orig="$PWD"; mkdir a; cd a; cd -; [[ "$PWD" == "$orig" ]]`,
 		"",
 	},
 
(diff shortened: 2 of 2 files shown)


donorp:
may be related to the `cd` issue mentioned in #705

donorp:
`cd` implemented in ash, bash, zsh errors if there is no `OLDPWD` env set, but in this package, `OLDPWD` is always set before running, so the following patch does nothing (and not applied in this pull request).

```patch
diff --git a/interp/builtin.go b/interp/builtin.go
index 08bbdf6c..5cb910f6 100644
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -218,7 +218,12 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			// replicate the commonly implemented behavior of `cd -`
 			// ref: https://www.man7.org/linux/man-pages/man1/cd.1p.html#OPERANDS
 			if len(path) == 1 && path[0] == '-' {
-				path = r.envGet("OLDPWD")
+				v := r.lookupVar("OLDPWD")
+				if !v.IsSet() {
+					r.errf("cd: OLDPWD not set\n")
+					return 1
+				}
+				path = v.String()
 			}
 		default:
 			r.errf("usage: cd [dir]\n")

```

I guess we are fine with such minor behavior difference from traditional shells.

interp/builtin.go
```diff
@@ -214,6 +214,12 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			path = r.envGet("HOME")
 		case 1:
 			path = args[0]
+
+			// replicate the commonly implemented behavior of `cd -`
+			// ref: https://www.man7.org/linux/man-pages/man1/cd.1p.html#OPERANDS
+			if len(path) == 1 && path[0] == '-' {
```
mvdan, line 220:
```suggestion
			if path == "-" {
```

interp/interp_test.go
```diff
@@ -928,6 +928,10 @@ var runTests = []runTest{
 		`mkdir a; ln -s a b; [[ "$(cd a && pwd -P)" == "$(cd b && pwd -L)" ]]; echo $?`,
 		"1\n",
 	},
+	{
+		`orig="$PWD"; mkdir a && cd a && cd - && [[ "$PWD" == "$orig" ]]`,
```
mvdan, line 932: a bit easier/shorter without the &&, and it will still fail if the feature isn't working:

```suggestion
		`orig="$PWD"; mkdir a; cd a; cd -; [[ "$PWD" == "$orig" ]]`,
```

mvdan (approved):
Thanks! Just two nits.

mvdan:
> I guess we are fine with such minor behavior difference from traditional shells.

I agree that the behavior you added is fine.

> may be related to the `cd` issue mentioned in #705

Not sure - they didn't really give enough details :)

interp/builtin.go
donorp: ah yeah, maybe I was too skeptical on string comparison, I wrote in this verbose style to avoid accidental zero width characters in source code.

Anyway, I will accept your suggestion if you think that doesn't matter. ;)

mvdan: You mean zero width characters in the `"-"` string?

If that were a worry, I'd have to refactor hundreds of other lines :) Not a worry for now, so I'd say simplify.

PR #769, opened 2021-12-01
interp: fix a PATH regression on Windows, fix tests on Wine

### LINKED ISSUE #768: interp: Regression on $PATH handling on Windows
Originally reported at:

- https://github.com/go-task/task/issues/619
- https://github.com/arduino/arduino-cli/pull/1576

Didn't had the time to try it myself yet (sorry), but wanted to report this sooner than later.

Apparently the latest release is messing with how executables see $PATH on Windows. For example, when running `go build foo.go` Gosh has no problems finding `go`, but the `go` command has problems finding other executables like `gcc`.

mvdan:
(see commit messages - please do not squash)

Fixes #768.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -14,9 +14,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


mvdan:
FYI @cclerget @riacataquian @andreynering @glumia

riacataquian (approved):
> Some versions of Windows, such as Wine, don't ship with "tr".
Just like we do with other commands like "wc" or "sed",
implement a very basic one as a builtin.

Got it! Thanks!

cclerget:
Oops sorry for the inconvenience !

> We had no tests for this behavior,
> and any script using Unix-like paths in any non-trivial way
> would likely need to be adjusted to be compatible with Windows anyway.

Yes, handling PATH specifically breaks consistency with other PATH like variables

cclerget (approved):

PR #770, opened 2021-12-05
interp: fix negative elapsed time

### LINKED ISSUE #767: `time` builtin returns negative seconds if duration is 30-60s, or 90-120s, etc, basically the 2nd half of the minute
```sh
% gosh
$ date ; time sleep 31 ; date
Tue Nov 30 12:12:05 EST 2021

real	0m-28.990s
user	0m0.000s
sys	0m0.000s
Tue Nov 30 12:12:36 EST 2021
```

Implementation of `time` uses `math.Remainder`, which can return negative numbers for two positive arguments.

https://github.com/mvdan/sh/blob/23a22509643c2fa771e7b722c784707c1d3b4b2f/interp/runner.go#L712

See [this](https://go.dev/play/p/hEcbeWxh41i) Playground code:
```go
// ...
	for x := 0.0; x < 60.0; x++ {
		fmt.Println(x, math.Remainder(x, 60.0))
	}
// ...
```

output:
[...]
28 28
29 29
30 30
31 -29
32 -28
33 -27
[...]

The doc for `math.Remainder` says "Remainder returns the IEEE 754 floating-point remainder of x/y."

The Wikipedia entry for [IEEE-754-1985](https://en.wikipedia.org/wiki/IEEE_754-1985) notes "Floating point remainder. **This is not like a normal modulo operation, it can be negative for two positive numbers**. It returns the exact value of x–(round(x/y)·y)."  (Emphasis added.)  The 1985 spec has been superceded twice, first in 2008, but the Go source code cites a Sun C program with a copyright date of 1993, so pretty sure the Go code implements the 1985 spec.

fsmiamoto:
Change the operation to calculate the elapsed seconds from `Remainder` to `Mod`, as `Remainder` can return negative values.

Let me know if we need more test cases :)

Fixes #767

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3713,6 +3713,8 @@ func TestElapsedString(t *testing.T) {
 			true,
 			"610.00",
 		},
+		{31 * time.Second, false, "0m31.000s"},
+		{102 * time.Second, false, "1m42.000s"},
 	}
 	for _, tc := range tests {
 		t.Run(tc.in.String(), func(t *testing.T) {
--- a/interp/runner.go
+++ b/interp/runner.go
@@ -709,7 +709,7 @@ func elapsedString(d time.Duration, posix bool) string {
 		return fmt.Sprintf("%.2f", d.Seconds())
 	}
 	min := int(d.Minutes())
-	sec := math.Remainder(d.Seconds(), 60.0)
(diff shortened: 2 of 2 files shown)


mvdan (approved):
Nice, thanks!

PR #771, opened 2021-12-08
lexer: allow extglob wildcards as function names

### LINKED ISSUE #739: syntax: got `"}" can only be used to close a block` on a seemingly normal block
I'm struggling to understand what's wrong with this sample:
```console
$ shfmt
@() {
    echo "$@"
}
@ hello world\!
<standard input>:3:1: "}" can only be used to close a block
```

riacataquian:
the lexer assumes an extglob token if any of the wildcards expressions
(such as `@`, and `+`) are succeeded by a left parenthesis but that
proves to be an issue if the wildcard is used as a function name.

example input:
```
$ cat in.sh
@() {
  echo "$@";
}
```

`bash` and `gosh` comparison:
```
$ bash ./in.sh
hello
$ ./gosh in.sh
in.sh:5:1: "}" can only be used to close a block
```

given `in.sh`, gosh reports about a syntax error - this is because
a closing bracket is found while the lexer isn't assuming a function
block

fix the issue by assuming a function if one of the conditions below is
true:
* if the expression is found at the beginning of the statement or if its
  preceded by a "function"
* if `(` is immediately succeeded by a `)` - although this is a valid
  bash syntax, we'll operate on the likelihood that it is a function

fixes #739

### DIFF
--- a/syntax/lexer.go
+++ b/syntax/lexer.go
@@ -290,7 +290,7 @@ skipSpace:
 				p.advanceLitNone(r)
 			}
 		case '?', '*', '+', '@', '!':
-			if p.peekByte('(') {
+			if p.tokenizeGlob() {
 				switch r {
 				case '?':
 					p.tok = globQuest
@@ -344,6 +344,28 @@ skipSpace:
 	if p.err != nil && p.tok != _EOF {
 		p.tok = _EOF
 	}
+}
+
+// tokenizeGlob determines whether the expression should be tokenized as a glob literal
+func (p *Parser) tokenizeGlob() bool {
+	if p.val == "function" {
+		return false
+	}
+	// NOTE: empty pattern list is a valid globbing syntax, eg @()
+	// but we'll operate on the "likelihood" that it is a function;
+	// only tokenize if its a non-empty pattern list
+	if p.peekBytes("()") {
+		return false
+	}
+	return p.peekByte('(')
+}
+
+func (p *Parser) peekBytes(s string) bool {
+	for p.bsp+(len(p.bs)-1) >= len(p.bs) {
+		p.fill()
+	}
+	bw := p.bsp + len(s)
+	return bw < len(p.bs) && bytes.HasPrefix(p.bs[p.bsp:bw], []byte(s))
 }
 
 func (p *Parser) peekByte(b byte) bool {
@@ -882,7 +904,7 @@ loop:
 			tok = _Lit
 			break loop
 		case '?', '*', '+', '@', '!':
-			if p.peekByte('(') {
+			if p.tokenizeGlob() {
 				tok = _Lit
 				break loop
 			}
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3117,6 +3117,20 @@ hello, world
 hello, world
 `,
 	},
+	{
+		// globbing wildcard as function name
+		`@() { echo "$@"; }; @ lala; function +() { echo "$@"; }; + foo`,
+		"lala\nfoo\n",
+	},
+	{
+		`      @() { echo "$@"; }; @ lala;`,
+		"lala\n",
+	},
+	{
+		// globbing wildcard as function name but with space after the name
+		`+ () { echo "$@"; }; + foo; @ () { echo "$@"; }; @ lala; ? () { echo "$@"; }; ? bar`,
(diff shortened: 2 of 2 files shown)


syntax/lexer.go
```diff
@@ -346,6 +346,26 @@ skipSpace:
...
+	}
+	// it must not be a globbing if its at the beginning of a statement or
+	// if its preceded by "function"
+	if p.pos.Col() == 1 || p.val == "function" {
+		return false
+	}
+	// its must not be a globbing if left parens is succeeded by a right parens
+	// NOTE: empty pattern list is a valid globbing syntax, eg @()
+	// but we'll operate on the "likelihood" that it is a function
+	if len(p.bs) > p.bsp+p.w {
```
mvdan, line 363: wouldn't func declarations be already covered by the `if` above?

syntax/lexer.go
```diff
@@ -346,6 +346,26 @@ skipSpace:
...
 
+// isExprGlob determines whether the expression is an extglob wildcard literal
+func (p *Parser) isExprGlob() bool {
+	leftParen := p.peekByte('(')
+	if !leftParen {
+		return false
+	}
+	// it must not be a globbing if its at the beginning of a statement or
+	// if its preceded by "function"
+	if p.pos.Col() == 1 || p.val == "function" {
```
mvdan, line 357: hmm, I don't think `p.pos.Col() == 1` is right; I can declare a function like:

```
# three spaces first
   f() { body; }
```

mvdan (commented):
I think a more reliable condition would be:

when I see `@`, if I peek the next two bytes to be exactly `()`, don't tokenize as a glob.

The slight wrinkle there is that, currently, we only have `peekByte` which only looks one byte ahead. However, looking two bytes ahead is still reasonable - we only need to be worried about peeking a large or unbounded amount of bytes. So, mimicking `peekByte`, you can imagine a `peekBytes` call - taking a string, ensuring that the buffer has enough bytes for it via `fill`, and then comparing with e.g. `bytes.HasPrefix`.

riacataquian:
Thanks for the helpful inputs @mvdan, I added `peekBytes` and it seems a lot cleaner now. I didn't put any restrictions to the length of the string in question/for comparison, thought I'd leave that to the discretion of the caller.

syntax/lexer.go
```diff
@@ -346,6 +346,29 @@ skipSpace:
...
+	// but we'll operate on the "likelihood" that it is a function;
+	// only tokenize if its a non-empty pattern list
+	if p.peekBytes("()") {
+		return false
+	}
+	return p.peekByte('(')
+}
+
+func (p *Parser) peekBytes(s string) bool {
+	if p.bsp == len(p.bs) {
```
mvdan, line 364: use len(s) here

mvdan (approved):

syntax/lexer.go
riacataquian: hmm sorry, I don't think I understand. my understanding is, with `p.bsp == len(p.bs)`, the intention is to load more bytes from the source if we're at the last bit. 

do you mean `len(s) > len(p.bs)` as in, if `p.bs` doesn't have enough bytes to compare with `s` load more to the buffer? wouldn't `p.bsp == len(p.bs) || len(s) > len(p.bs)` make more sense?

mvdan: Think of `peekByte(b byte)` as a shortcut for `peekBytes(s string)` where `len(s) == 1`. It does `if p.bsp == len(p.bs)` as an equivalent to `if p.bsp >= len(p.bs)`; it just needs one byte, and `p.bsp` is the position of the next byte within `bs`, so that's enough.

In `peekBytes`, you want to peek `len(s)` bytes, not just one. So the right check seems to be `if p.bsp + (len(bs)-1) >= len(p.bs)`.

Another question here is how `p.fill` only guarantees to fill at least one byte. So I think you should swap `if` for `for`; otherwise, if you need to peek 2 bytes and you have 0, one `fill` call could fill just one byte, so you'd need to call it again.

tl;dr:

```
for p.bsp + (len(bs)-1) >= len(p.bs) {
    p.fill()
}
```

riacataquian: I totally misunderstood. Thanks for the thorough explanation :)

PR #775, opened 2021-12-14
all: make tests pass on the debian-minimal docker image

### LINKED ISSUE #773: some tests fail when system have unconfigured locales
Hi there!

On Linux, when you run tests on a chroot/docker that do not have locales
installed and configured, some tests, like parser_test.go and interp_test.go, fail.

One example of fail:

=== CONT  TestRunnerRunConfirm/#133
    interp_test.go:3280: wrong bash output in "INTERP_GLOBAL=; $ENV_PROG | grep '^INTERP_GLOBAL='":
        want: "INTERP_GLOBAL=\n"
        got:  "bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)\nINTERP_GLOBAL=\n"

As some systems, like Debian, use CI tests in an environment like that,
please, can you consider ignoring warnings like this on the tests?

Cheers,
mt

mvdan:
(see commit message)

Fixes #773.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -130,8 +130,18 @@ func TestMain(m *testing.M) {
 	}
 	os.Setenv("GOSH_PROG", prog)
 
-	os.Setenv("LANGUAGE", "en_US.UTF-8")
-	os.Setenv("LC_ALL", "en_US.UTF-8")
+	// Set the locale to computer-friendly English and UTF-8.
+	// Some systems like Arch miss C.UTF8, so fall back to the US English locale.
+	if out, _ := exec.Command("locale", "-a").Output(); strings.Contains(
+		strings.ToLower(string(out)), "c.utf",
+	) {
+		os.Setenv("LANGUAGE", "C.UTF-8")
+		os.Setenv("LC_ALL", "C.UTF-8")
+	} else {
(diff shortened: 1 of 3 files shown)


mvdan:
@talau this makes the tests pass for me on `debian:11.1-slim` on Docker, where they used to fail before, with similar errors as yours. Can you confirm this fix is enough for you?

I want to release v3.4.2 before the holidays, so I could backport this fix as well.

talau:
Hi there!

I tested on Debian and it's fixed the issue! Thanks @mvdan!

You can see the log here: https://salsa.debian.org/talau-guest/shfmt-tests-minimal/-/jobs/2277914/raw

By the way, shfmt has already been sent to Debian. As it is a new package, it went to the NEW queue, I believe that in a few weeks it will be approved. You can see it here: https://ftp-master.debian.org/new/shfmt_3.4.1-1.html


Regards,
mt

mvdan:
Thanks for confirming! Merging then. You can subscribe to releases if you'd like to be notified when v3.4.2 is out.

riacataquian:
I missed to review this but I agree with the changes. Thanks for fixing! On my machine, this will still run on the fallback en_US.

PR #780, opened 2021-12-17
interp: re-enable process substitution tests on Mac

### LINKED ISSUE #576: interp: process substitution named pipes seem to be flaky on Darwin

mvdan:
(see commit message)

Fixes #576.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3179,15 +3179,10 @@ func init() {
 // ln -s: wine doesn't implement symlinks; see https://bugs.winehq.org/show_bug.cgi?id=44948
 var skipOnWindows = regexp.MustCompile(`ln -s`)
 
-// process substitutions seemflaky on mac; see https://github.com/mvdan/sh/issues/576
-var skipOnMac = regexp.MustCompile(`>\(|<\(`)
-
 func skipIfUnsupported(tb testing.TB, src string) {
 	switch {
 	case runtime.GOOS == "windows" && skipOnWindows.MatchString(src):
 		tb.Skipf("skipping non-portable test on windows")
-	case runtime.GOOS == "darwin" && skipOnMac.MatchString(src):
-		tb.Skipf("skipping non-portable test on mac")
 	}
 }
 


mvdan:
@theclapp would love to see if you can still reproduce those failures :)

mvdan:
Going to merge this for now, as I don't want to ping people during the holidays, and it's a test-only change.

mvdan:
I had to revert this PR; see https://github.com/mvdan/sh/commit/71b5012474a5e3a9255a1628bf035c738cc6f738.

PR #782, opened 2021-12-25
Check for empty paths in builtin test

### LINKED ISSUE #781: interp: test -d "" succeeds when it should fail
I've been testing shell portability for a silly script of mine, and hit upon a behavior difference with the built-in command `test`:

```shell
for shell in sh bash dash zsh ksh csh tcsh gosh; do 
    printf "%s\t" "$shell "
    $shell -c "test -d '' && echo true || echo false"
done
```
```shell
sh 	false
bash 	false
dash 	false
zsh 	false
ksh 	false
csh 	false
tcsh 	false
gosh 	true
```

I noticed this as I had code relying upon [ -d "${XDG_CACHE_HOME:-}" ] to fail when the env var was unset.

Poking around the code for a few minutes led me to [Runner.absPath](https://github.com/mvdan/sh/blob/4a4c1600341f6413719f5e8a50105b8d13a96d6d/interp/builtin.go#L870-L875). It has a few callsites however, and I haven't audited them to see if the obvious change of checking for `path == ""` would impact them.

raylee:
`test -d ""` succeeds when it should fail. Checking all the flags which take a path in `gosh` (ignoring G and O which are not yet implemented):

```sh
for flag in b c d e f g h k L p r s S u w x; do
    test -$flag ""
    echo -n "$flag$? "
done
```

Results in `b1 c1 d0 e0 f1 g1 h1 k1 L1 p1 r0 s0 S1 u1 w1 x1`. Some are correctly failing, but for the wrong reasons.

`absPath` is updated to check for empty paths and return them untouched. `changeDir` was relying on the old behavior, so it now explicitly implements the same behavior as bash, dash, mksh, and zsh, where a `cd ''` is equivalent to a `cd .`. (This was also the previous behavior in gosh.)

Shell behavior was surveyed with:

```sh

try() {
	local cmd="$1" want="$2"
	for shell in bash csh dash gosh ksh mksh posh sh tcsh zsh; do
		local got
		got="$($shell -c "$cmd" 2>&1)"
		if [ "$got" != "$want" ]; then
			printf "~   FAIL: %s\n" "$shell: $cmd"
			printf "    want: %s\n" "$want"
			printf "     got: %s\n" "$got"
			echo
		fi
	done
}

try 'cd /; cd /tmp; cd ""; echo $? $PWD $OLDPWD' '0 /tmp /tmp'

try 'for flag in b c d e f g h k L p r s S u w x; do test -$flag ""; echo -n "$flag$? "; done' \
    'b1 c1 d1 e1 f1 g1 h1 k1 L1 p1 r1 s1 S1 u1 w1 x1 '
```

Closes #781

### DIFF
--- a/interp/handler.go
+++ b/interp/handler.go
@@ -276,7 +276,7 @@ type OpenHandlerFunc func(ctx context.Context, path string, flag int, perm os.Fi
 func DefaultOpenHandler() OpenHandlerFunc {
 	return func(ctx context.Context, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
 		mc := HandlerCtx(ctx)
-		if !filepath.IsAbs(path) {
+		if path != "" && !filepath.IsAbs(path) {
 			path = filepath.Join(mc.Dir, path)
 		}
 		return os.OpenFile(path, flag, perm)
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -192,13 +192,6 @@ type concBuffer struct {
 	sync.Mutex
 }
 
-func absPath(dir, path string) string {
-	if !filepath.IsAbs(path) {
-		path = filepath.Join(dir, path)
-	}
-	return filepath.Clean(path)
-}
-
 func (c *concBuffer) Write(p []byte) (int, error) {
 	c.Lock()
 	n, err := c.buf.Write(p)
@@ -858,6 +851,10 @@ var runTests = []runTest{
 	{"[[ a == [ab ]]", "exit status 1"},
 	{`HOME='/*'; echo ~; echo "$HOME"`, "/*\n/*\n"},
 	{`test -d ~`, ""},
+	{
+		`for flag in b c d e f g h k L p r s S u w x; do test -$flag ""; echo -n "$flag$? "; done`,
+		`b1 c1 d1 e1 f1 g1 h1 k1 L1 p1 r1 s1 S1 u1 w1 x1 `,
+	},
 	{`foo=~; test -d $foo`, ""},
 	{`foo=~; test -d "$foo"`, ""},
 	{`foo='~'; test -d $foo`, "exit status 1"},
@@ -880,6 +877,10 @@ var runTests = []runTest{
 	},
 	{
 		`w="$HOME"; cd; [[ $PWD == "$w" ]]`,
+		"",
+	},
+	{
+		`mkdir test.cd; cd test.cd; cd ''; [[ "$PWD" == "$OLDPWD" ]]`,
 		"",
 	},
 	{
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -853,6 +853,9 @@ func (r *Runner) readLine(raw bool) ([]byte, error) {
 }
 
 func (r *Runner) changeDir(path string) int {
+	if path == "" {
+		path = "."
+	}
 	path = r.absPath(path)
 	info, err := r.stat(path)
 	if err != nil || !info.IsDir() {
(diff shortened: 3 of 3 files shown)


interp/interp_test.go
```diff
@@ -882,6 +889,10 @@ var runTests = []runTest{
 		`w="$HOME"; cd; [[ $PWD == "$w" ]]`,
 		"",
 	},
+	{
+		"cd /; cd /tmp; cd ''; echo $? $PWD $OLDPWD",
```
mvdan, line 893: this might fail as it depends on /tmp being present, and the tests also run on e.g. mac and windows. perhaps mkdir some directories and use those instead.

interp/interp_test.go
```diff
@@ -193,6 +193,9 @@ type concBuffer struct {
 }
 
 func absPath(dir, path string) string {
+	if path == "" {
+		return ""
+	}
```
mvdan, line 198: I hadn't noticed we had two copies of this func, one as a method and another as a standalone func. Perhaps move this next to the method, and make the method do `return absPath(r.Dir, path)`.

interp/handler.go
```diff
@@ -276,7 +276,7 @@ type OpenHandlerFunc func(ctx context.Context, path string, flag int, perm os.Fi
 func DefaultOpenHandler() OpenHandlerFunc {
 	return func(ctx context.Context, path string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
 		mc := HandlerCtx(ctx)
-		if !filepath.IsAbs(path) {
+		if path != "" && !filepath.IsAbs(path) {
```
mvdan, line 279: Can you add a test for this? for instance, `"" foo`.

mvdan (commented):
Looks good! Just some nits.

interp/interp_test.go
raylee: Ah, thanks. It passes on my Macbook and under git-bash on Windows, but does indeed fail on Windows when not under the emulation layer. Fixed.

interp/handler.go
raylee: This is implicitly checked via `test -r ""` (runner: unTest -> open -> openHandler), but that's probably a bit too far away for a clear test.

I've added a test for DefaultOpenHandler.

raylee:
The new test for DefaultOpenHandler fails on Windows at the moment because I've forgotten NTFS semantics. Don't approve the workflow yet!

interp/handler.go
raylee: Oh, you probably meant to add just another shell test here rather than a Go test, didn't you. If so I'll respin the patch.

mvdan: Oh, this is the default *open* handler, not the default *exec* handler. My suggestion above about `"" foo` is entirely irrelevant.

My apologies. Please just back out TestDefaultOpenHandler, as the added test cases already cover this code, like you correctly pointed out. More test code isn't necessary.

mvdan (approved):
LGTM if you remove the test.

The CI failure is also not your fault; see https://github.com/mvdan/sh/commit/71b5012474a5e3a9255a1628bf035c738cc6f738. If you rebase on the latest master, it should be green.

It would also help if you squashed the commits so I can rebase-merge directly. Also see https://github.com/golang/go/wiki/CommitMessage; that's the commit message style I try to follow. Not a problem if you don't want to do this part; I can edit the message when merging too.

raylee:
I appreciate the feedback. I've reworded the commit message to be more aligned with that guide, but it's still my slightly-chatty style. You're more than welcome to edit it as you see fit.

mvdan (approved):
Great, thanks again!

PR #787, opened 2022-01-01
syntax: add support for heredocs directly inside backquotes

### LINKED ISSUE #729: syntax: unable to parse heredoc inside backtick
First off, I'd like to say thank you for maintaining this project. We're using your syntax parser in [ossf/scorecard](https://github.com/ossf/scorecard) shell files from thousands of repos. We're getting an error when trying to parse a specific file, and since the syntax is valid bash, I'm opening this issue. I've created a small program that reproduces the error:
```bash
#!/usr/bin/env bash

OUTPUT=`passwd 2>&1 << EOT
old_password
old_password
new_password
EOT`

echo $OUTPUT
```
Passing this file to `shfmt` gives this error:
```
<standard input>:7:4: reached EOF without closing quote `
```

mvdan:
(see commit message)

Fixes #729.

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -1035,6 +1035,24 @@ var fileTests = []testCase{
 		},
 	},
 	{
+		Strs: []string{
+			"a <<EOF\nfoo$(bar)baz\nEOF",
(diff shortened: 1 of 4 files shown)


riacataquian (approved):
LGTM!

PR #788, opened 2022-01-02
interp_test: fix broken touch implementation

### LINKED ISSUE #784: interp: touch timestamp test is flaky
https://github.com/mvdan/sh/runs/4669017549?check_suite_focus=true

```
    --- FAIL: TestRunnerRun/#435 (0.01s)
        interp_test.go:3233: wrong output in "touch -t 202111050000 a; touch -t 202111060000 b; [[ a -ot b ]]":
            want: ""
            got:  "exit status 1"
FAIL
```

I believe @riacataquian also ran into this failure, also on Windows.

The tests use `touch -t` instead of `touch -d` since d36d14903235de79f41ca3bea388f83ce33ad174, but the test builtin only implements `-d`. So we are effectively having `test -t 123 foo` create three files `-t`, `123`, and `foo` - without a custom timestamp :see_no_evil: 

The fix is: 1) fix up the touch implementation in interp_test.go to support the `-t` flag, and 2) add some sanity check that none of the interp_test.go builtins are misusing flags as arguments. For instance, by failing if any argument begins with a dash.

This is not a bug in the interpreter, but still a bug - in the test code.

riacataquian:
the interp_test's `touch` does not support the `-t` flag, instead it
treats it as filenames to be created

to fix, remove the `-d` support in favor of `-t` as part of the
compatibility fixes introduced in d36d14903235de79f41ca3bea388f83ce33ad174

if the `-t` flag is detected, parse the next argument to time.Time
as timestamp to the files (the rest of the argument) that will be created

fixes #784

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -15,7 +15,6 @@ import (
 	"regexp"
 	"runtime"
 	"sort"
-	"strconv"
 	"strings"
 	"sync"
 	"testing"
@@ -1408,15 +1407,15 @@ var runTests = []runTest{
 		"exit status 1",
 	},
 	{
-		"touch -t 202111050000 a b; [[ a -nt b || a -ot b ]]",
+		"touch -t 202111050000.30 a b; [[ a -nt b || a -ot b ]]",
 		"exit status 1",
 	},
 	{
-		"touch -t 202111050000 a; touch -t 202111060000 b; [[ a -nt b ]]",
+		"touch -t 202111050200.00 a; touch -t 202111060100.00 b; [[ a -nt b ]]",
 		"exit status 1",
 	},
 	{
-		"touch -t 202111050000 a; touch -t 202111060000 b; [[ a -ot b ]]",
+		"touch -t 202111050000.00 a; touch -t 202111060000.00 b; [[ a -ot b ]]",
 		"",
 	},
 	{
@@ -1616,11 +1615,11 @@ var runTests = []runTest{
 		"1:1: -lt must be followed by a word\nexit status 2 #JUSTERR",
 	},
 	{
-		"touch -t 202111050000 a; touch -t 202111060000 b; [ a -nt b ]",
+		"touch -t 202111050000.00 a; touch -t 202111060000.00 b; [ a -nt b ]",
 		"exit status 1",
 	},
 	{
-		"touch -t 202111050000 a; touch -t 202111060000 b; [ a -ot b ]",
+		"touch -t 202111050000.00 a; touch -t 202111060000.00 b; [ a -ot b ]",
 		"",
 	},
 	{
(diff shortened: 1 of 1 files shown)


interp/interp_test.go
```diff
@@ -3427,19 +3426,55 @@ var testBuiltinsMap = map[string]func(HandlerContext, []string) error{
...
 			if err != nil {
 				return err
 			}
-			newTime = time.Unix(sec, 0)
-			args = args[2:]
+			newTime = s
 		}
-		for _, arg := range args {
+
+		for _, arg := range filenames {
```
mvdan, line 3477: nit: also error if any of the filenames begin with a dash

interp/interp_test.go
```diff
@@ -3427,19 +3426,55 @@ var testBuiltinsMap = map[string]func(HandlerContext, []string) error{
...
+			}
+			filenames = args[2:] // treat the rest of the args as filenames
+
+			arg := args[1]
+			if strings.HasPrefix(arg, "-") {
+				return fmt.Errorf("usage: touch [-t [[CC]YY]MMDDhhmm[.SS]] file")
+			}
+			var offset int // index after yy
+			var (          // optional date fields
+				yy    = "2022"
```
mvdan, line 3444: we just need this builtin for the tests; I think we can just force the user to always include the year :)

interp/interp_test.go
```diff
@@ -3427,19 +3426,55 @@ var testBuiltinsMap = map[string]func(HandlerContext, []string) error{
...
-			if !strings.HasPrefix(args[1], "@") {
-				return fmt.Errorf("unimplemented")
+		if args[0] == "-t" {
+			if len(args) < 3 {
+				return fmt.Errorf("usage: touch [-t [[CC]YY]MMDDhhmm[.SS]] file")
+			}
+			filenames = args[2:] // treat the rest of the args as filenames
+
+			arg := args[1]
+			if strings.HasPrefix(arg, "-") {
```
mvdan, line 3439: I think this check is unnecessary; if the timestamp begins with a dash, time.Parse should fail anyway.

interp/interp_test.go
```diff
@@ -3427,19 +3426,55 @@ var testBuiltinsMap = map[string]func(HandlerContext, []string) error{
...
+			var ( // field start indices
+				month = offset
+				dd    = month + chunk
+				hh    = dd + chunk
+				mm    = hh + chunk
+			)
+			v := func(i int) string {
+				return arg[i : i+chunk]
+			}
+			s, err := time.Parse(time.RFC3339, fmt.Sprintf("%s-%s-%sT%s:%s:%sZ", yy, v(month), v(dd), v(hh), v(mm), ss))
```
mvdan, line 3470: can you not use time.Parse with a time format string that matches what the touch command uses? using fmt.Sprintf to match onto RFC3339 seems backwards.

it especially seems backwards because you're doing most of the parsing yourself, splitting the input string to get the year, month, day, etc. let time.Parse do that for you :)

mvdan: I understand you might have done this to support all forms of timestamps that `touch -t` supports, but we just need this for the tests, so we're fine just supporting one static timestamp format.

riacataquian: gotcha, make sense. fixed to simplify things :)

interp/interp_test.go
riacataquian: nice catch, bash would actually error. added check.

interp/interp_test.go
```diff
@@ -3427,19 +3426,31 @@ var testBuiltinsMap = map[string]func(HandlerContext, []string) error{
...
+				return fmt.Errorf("usage: touch [-t [[CC]YY]MMDDhhmm[.SS]] file")
+			}
+			filenames = args[2:] // treat the rest of the args as filenames
+
+			arg := args[1]
+			if len(arg) > 15 {
+				return fmt.Errorf("usage: touch [-t [[CC]YY]MMDDhhmm[.SS]] file")
 			}
-			sec, err := strconv.ParseInt(args[1][1:], 10, 64)
+			arg = strings.ReplaceAll(arg, ".", "")
```
mvdan, line 3442: nit: I think you can just asd the period to the format string passed to time.Parse.

mvdan (approved):

PR #794, opened 2022-01-21
readme: document trunk

sxlijin:
Trunk is a recently launched universal linter which integrates with a wide variety of linters and formatters, `shfmt` included.

We're big fans of `shfmt` and figured that you might find our tool to be interesting enough to include it here.


mvdan (commented):
Sounds fine with me, but could you please just do with one link? It wouldn't be fair to the others to make one item stand out like this, and it feels unnecessary. Perhaps trunk.io, as that links to the marketplace and presumably the other products as well.

sxlijin:
Absolutely!

mvdan (approved):
Thanks!

PR #796, opened 2022-01-22
cmd/shfmt: change -ln default to "auto"

### LINKED ISSUE #622: cmd/shfmt: change the language default from "bash" to "auto"
That is, auto-detection of what language the input shell should be in. This is because, right now, running something like `shfmt -l -w .` on a large repository will simply use the default of `-ln=bash` for all files. One can use `-ln=posix` to change that, but then all files are parsed as POSIX.

It's not terribly common to mix different shell languages in a single repository, but it happens. For example, imagine a repository which has a few simple POSIX shell scripts for extreme portability, some complex Bash scripts for development, and a few Bats files for unit testing other shell scripts. Right now, this would require at least three `shfmt` calls, and manually hard-coding which files are in which language.

We can do better. The files themselves often tell us what language they are in via the filename extension and the shebang.

I propose the following algorithm to auto-detect the language. When a language is selected, the following steps aren't considered.

1) If the filename ends with `.bash`, `.mksh`, or `.bats`, we use that language.
2) If the content begins with a valid shell shebang like `#!/bin/sh` or `#!/usr/bin/env bash`, we use that language. Here, `sh` means POSIX shell.
3) We fall back to the current default, assuming Bash.

Step 1 does not assume `.sh` means POSIX, because Bash scripts with that extension are too common. If a file is meant to be POSIX shell, give it a `sh` shebang, which will be matched by step 2.

Step 3 allows us to not break backwards compatibility. If we don't know what language the input is in, we'll just keep the current conservative fallback that is Bash.

To my mind, the only downside to this change is that it adds a bit of complexity; we now need to explain what `-ln=auto` means. But I think that should be doable in a few lines in the added man page, just like we did in three list items above.

--

This is a reboot of https://github.com/mvdan/sh/issues/429, which I actually forgot about. Fun how I came up with almost exactly the same idea again. The reason I think this is worth considering again is two-fold:

* The default of `-ln=bash` is still too stiff, in my opinion. Support for Bats is a clear example; a project having both Bash and Bats files is entirely normal.
* EditorConfig does kind of mitigate this problem, but having to enter `[*.bats]\nshell_variant = "bats"` is silly. Plus, since EditorConfig is entirely filename-based, it does not cover Step 2 from this proposal.

CC @jansorg @kolyshkin @lassik @kaey since you all contributed to related issues. If we reach consensus on this, I'd implement this for 3.3.0, which should come out in early 2021.

mvdan:
(see commit message)

Fixes #622.

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -13,6 +14,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime/debug"
+	"strings"
 
 	maybeio "github.com/google/renameio/maybe"
 	"github.com/pkg/diff"
@@ -24,7 +26,8 @@ import (
 	"mvdan.cc/sh/v3/syntax"
 )
 
-const unsetLang = syntax.LangVariant(-1)
+// When unset, we auto-detect what language to use.
+const langAuto = syntax.LangVariant(-1)
 
 var (
 	showVersion = flag.Bool("version", false, "")
@@ -39,7 +42,7 @@ var (
 	// useEditorConfig will be false if any parser or printer flags were used.
 	useEditorConfig = true
 
-	lang     = unsetLang
+	langFlag = langAuto
 	posix    = flag.Bool("p", false, "")
 	filename = flag.String("filename", "", "")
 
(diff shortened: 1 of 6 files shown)


mvdan:
Going to merge this for now. Reviews and testing are still very much welcome, even after this lands in master.

kolyshkin:
Tested (better late than never I guess); filed https://github.com/mvdan/sh/issues/802

PR #800, opened 2022-01-25
Add BashSupport Pro as related project

jansorg:
Adds the link to the existing list, which seems to be sorted alphabetical.  Trunk, the latest addition, seems to a paid software, too. 
If links to paid software are unwanted, please feel free to close.

BashSupport Pro is a paid plugin for JetBrains IDEs and bundles shfmt, but also support external installations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ To use the Docker image, run:
 
 The following editor integrations wrap `shfmt`:
 
+- [BashSupport-Pro] - Bash plugin for JetBrains IDEs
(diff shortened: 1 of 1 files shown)


mvdan (approved):
SGTM. I am not keeping a high bar in terms of license or source availability; I think what matters more is whether the integration with this project is meaningful. In that sense, BashSupport-Pro seems more on topic than, say, an open source "multi linter" that simply knows how to call `shfmt`.

Issue #806, opened 2022-02-02
interp: unset of a global var from inside a function broken since v3.4.1


dtrudg:
Since v3.4.1, an `unset` call in a function is not un-setting the variable correctly:

```
$ cat unset.sh 
#!/bin/sh

myfunc() {
        echo "Unsetting MYVAR"
        unset MYVAR
        echo "MYVAR is now: ${MYVAR}"
}

MYVAR="I AM SET"
myfunc
```

```
# bash
$ bash unset.sh 
Unsetting MYVAR
MYVAR is now: 

# dash
$ dash unset.sh 
Unsetting MYVAR
MYVAR is now: 

# gosh (master)
$ gosh unset.sh 
Unsetting MYVAR
MYVAR is now: I AM SET

# gosh (v3.4.2)
$ gosh unset.sh 
Unsetting MYVAR
MYVAR is now: I AM SET

# gosh (v3.4.1)
$ gosh unset.sh 
Unsetting MYVAR
MYVAR is now: I AM SET

# gosh (v3.4.0)
$ gosh unset.sh 
Unsetting MYVAR
MYVAR is now:
```

If the `unset` is performed outside of the function it works.

mvdan:
Sounds like https://github.com/mvdan/sh/commit/dc2e11e6cf3536833b6a0e2c7b4cc7d3c609dfd4 could be the culprit; could you confirm by reverting the change, or comparing the commit with its parent?

mvdan:
cc @cclerget, who authored the commit above, and who I think works on the same project.

dtrudg:
Hi @mvdan - I'd just looked into the changelog and pinged @cclerget out of band. I'm afraid that we actually work for 2 different companies now, on different sides of a fork of the Singularity project so we haven't been in direct communication for some time.

Can confirm the commit introduces the problem, while the parent is okay:

```
09:51 AM $ go install mvdan.cc/sh/v3/cmd/gosh@dc2e11e6
09:51 AM $ gosh unset.sh 
Unsetting MYVAR
MYVAR is now: I AM SET

09:51 AM $ go install mvdan.cc/sh/v3/cmd/gosh@423f205
09:52 AM $ gosh unset.sh 
Unsetting MYVAR
MYVAR is now: 
```

dtrudg:
> Ah I see, I wasn't aware of the forking. I imagine that means both are using the shell interpreter, at least for now.

Yep, we are not very diverged at present, and both using the same code that hits this.

[...]

PR #807, opened 2022-02-02
interp: fix unset global inside function

dtrudg:
In dc2e11e `unset` of a global var inside a function was broken, as the special handling of `export` or `readonly` in function scope introduced in that commit also incorrectly captured `unset`, calling a parent.Set with the previous value.

Ensure we do not capture unset in the function scope global handling.

Unset of a global can fall-through this code. Globals are unset correctly further down the Set function, so we do not need to invoke a parent.Set at all.

Fixes mvdan#806

### DIFF
--- a/interp/vars.go
+++ b/interp/vars.go
@@ -31,17 +31,21 @@ func (o *overlayEnviron) Get(name string) expand.Variable {
 }
 
 func (o *overlayEnviron) Set(name string, vr expand.Variable) error {
-	// "foo=bar" in a function updates the global scope, unless the variable
-	// has been declared as local.
+	// Manipulation of a global var inside a function
 	if o.funcScope && !vr.Local && !o.values[name].Local {
+		// "foo=bar" on a global var in a function updates the global scope
+		if vr.IsSet() {
+			return o.parent.(expand.WriteEnviron).Set(name, vr)
+		}
 		// "foo=bar" followed by "export foo" or "readonly foo"
-		if !vr.IsSet() {
+		if vr.Exported || vr.ReadOnly {
 			prev := o.Get(name)
 			prev.Exported = prev.Exported || vr.Exported
 			prev.ReadOnly = prev.ReadOnly || vr.ReadOnly
 			vr = prev
+			return o.parent.(expand.WriteEnviron).Set(name, vr)
 		}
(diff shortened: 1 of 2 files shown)


dtrudg:
@mvdan @cclerget - I believe that this is a valid fix for #806 - cheers.

interp/vars.go
```diff
@@ -31,17 +31,21 @@ func (o *overlayEnviron) Get(name string) expand.Variable {
...
-	// has been declared as local.
+	// Manipulation of a global var inside a function
 	if o.funcScope && !vr.Local && !o.values[name].Local {
+		// "foo=bar" on a global var in a function updates the global scope
+		if vr.IsSet() {
+			return o.parent.(expand.WriteEnviron).Set(name, vr)
+		}
 		// "foo=bar" followed by "export foo" or "readonly foo"
-		if !vr.IsSet() {
+		if !vr.IsSet() && (vr.Exported || vr.ReadOnly) {
```
mvdan, line 41:
```suggestion
		if vr.Exported || vr.ReadOnly {
```

since we already know IsSet is false.

interp/vars.go
```diff
@@ -31,17 +31,21 @@ func (o *overlayEnviron) Get(name string) expand.Variable {
...
 			prev := o.Get(name)
 			prev.Exported = prev.Exported || vr.Exported
 			prev.ReadOnly = prev.ReadOnly || vr.ReadOnly
 			vr = prev
+			return o.parent.(expand.WriteEnviron).Set(name, vr)
 		}
-		return o.parent.(expand.WriteEnviron).Set(name, vr)
+		// `unset` will fall-through as it is handled below.
```
mvdan, line 48:
```suggestion
		// "unset" is handled below.
```

also since we don't use markdown in comments

mvdan (approved):
Indeed looks like the right fix, thank you!

interp/vars.go
dtrudg: resolved in 3999919

mvdan:
Thanks! I'll most likely tag v3.4.3 including this fix in a few days. In the meantime, you can either use master or the v3.4 branch, which is where I'm cherry-picking the bugfix changes on top of v3.4.2.

PR #812, opened 2022-02-15
all: fix a few typos spotted by gospel

mvdan:
(see commit message)

### DIFF
--- a/interp/handler.go
+++ b/interp/handler.go
@@ -79,7 +79,7 @@ type ExecHandlerFunc func(ctx context.Context, args []string) error
 // DefaultExecHandler returns the ExecHandlerFunc used by default.
 // It finds binaries in PATH and executes them.
 // When context is cancelled, an interrupt signal is sent to running processes.
-// KillTimeout is a duration to wait before sending the kill signal.
+// killTimeout is a duration to wait before sending the kill signal.
 // A negative value means that a kill signal will be sent immediately.
 //
 // On Windows, the kill signal is always sent immediately,
--- a/interp/runner.go
+++ b/interp/runner.go
@@ -660,7 +660,7 @@ func (r *Runner) trapCallback(ctx context.Context, callback, name string) {
 	r.handlingTrap = false
 }
 
-// setExit call this function to exit the shell with status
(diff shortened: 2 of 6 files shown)


mvdan:
cc @kortschak

interp/handler.go
```diff
@@ -79,7 +79,7 @@ type ExecHandlerFunc func(ctx context.Context, args []string) error
 // DefaultExecHandler returns the ExecHandlerFunc used by default.
 // It finds binaries in PATH and executes them.
 // When context is cancelled, an interrupt signal is sent to running processes.
-// KillTimeout is a duration to wait before sending the kill signal.
```
kortschak, line 82: We should have allowed this. It wasn't found by gospel was it?

mvdan: It was found by gospel, and it did puzzle me for a minute, but it's right. The docs initially referenced a field named `KillTimeout` (capitalized), but we switched to a _parameter_ named `killTimeout` (not capitalized), and we did not update the docs.

kortschak: I'm glad it found it, but I don't understand why (I'll look into it); any initial mismatch of "KillTimeout" would have kicked in the camel splitter to check "Kill" and "Timeout", both of which are standard dictionary words. On tip, it passes that label without any additional information.

mvdan: Ah, this was a branch I had worked on locally a few days ago, so perhaps it picked up the typo before you merged the camel case changes.

kortschak: I'd like to be able to get this back while retaining the camel splitting. There are a couple of ways we can do that; 1. keep a `map[string]string` of identifiers lowercased=>verbatim and if there is an initial spelling mismatch check that before camel splitting or 2. on split retry if a success happens check whether the each success part is case-insensitively matching but case-different and gives a case-insenstivly matching suggestion. 1. is simpler.

riacataquian (approved):
LGTM 🍪

interp/handler.go
mvdan: To close the loop, that was https://github.com/kortschak/gospel/pull/42.

PR #820, opened 2022-02-27
interp: run "Confirm" tests on CI

### LINKED ISSUE #785: interp: run "Confirm" tests on CI
Our interpreter test suite is checked against the real bash (as well as other shells, like mksh and dash). However, since those shells had bugs in older versions, the tests are skipped if the version is too old (or too new, as sometimes changes in behavior happen!).

This results in test failures being merged to master, as CI doesn't run those tests. For example, see https://github.com/mvdan/sh/commit/4a4c1600341f6413719f5e8a50105b8d13a96d6d, which fixed one such skipped test case.

We should make CI run those. For instance, we could use https://hub.docker.com/_/bash via https://github.com/mvdan/dockexec, which allows us to pin the same version of Bash that the tests want. We could possibly do something similar for mksh or other shells, but Bash is by far the most important one.

If anyone wants to contribute a patch: note that you can test this on GitHub Actions CI by forking the repo and pushing to a branch. You could also test how to use dockexec and that docker image with `go test` locally.

dhanusaputra:
update bash directly instead of use dockexec since got `no such file` error
close #785

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,8 +30,10 @@ jobs:
     
     - name: Install dockexec
       run: go install mvdan.cc/dockexec@latest
+      if: matrix.os == 'ubuntu-latest'
     - name: Test bash interpreter
-      run: go test ./interp/. -run TestRunnerRunConfirm -exec='dockexec bash:5.1'
+      run: go test ./interp/. -run TestRunnerRunConfirm -exec='dockexec bash:5.1 -e=CGO_ENABLED=0'
+      if: matrix.os == 'ubuntu-latest'
 
   test-linux-alpine:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
 
     - name: Run confirm tests with Bash 5.1
-      run: | 
+      run: |
         go install mvdan.cc/dockexec@latest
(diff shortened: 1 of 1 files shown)


mvdan:
> update bash directly instead of use dockexec since got `no such file` error

Did you see https://github.com/mvdan/dockexec#caveats? I still think using dockexec would be better long-term; mixing the use of docker images is more composable than directly installing programs on the host system.

dhanusaputra:
yes already tried that part but no luck
will try again

mvdan:
If you get stuck, you can push your changes here and I can help debug what's going on.

.github/workflows/test.yml
```diff
@@ -28,6 +28,13 @@ jobs:
       run: diff <(echo -n) <(gofmt -d .)
       if: matrix.os == 'ubuntu-latest'
 
+    - name: Install dockexec
+      run: go install mvdan.cc/dockexec@latest
+      if: matrix.os == 'ubuntu-latest'
+    - name: Test interpreter confirm
+      run: CGO_ENABLED=0 go test -run TestRunnerRunConfirm -exec 'dockexec bash:5.1' ./interp/.
```
mvdan, line 35: I would instead have one step that runs two commands, calling it something like `Run confirm tests with Bash 5.1`.

Also, you can just use `./interp` as the final argument.

mvdan (approved):
SGTM! Just one last nit.

mvdan (approved):
Thanks! cc @riacataquian

PR #821, opened 2022-03-05
Replace foo/bar with longer names

stoovon:
This will avoid the path collision seen in https://github.com/mvdan/sh/issues/738.

Fixes https://github.com/mvdan/sh/issues/738.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -55,8 +55,8 @@ func BenchmarkRun(b *testing.B) {
 	b.StopTimer()
 	src := `
 echo a b c d
-echo ./$foo/etc $(echo foo bar)
-foo="bar"
+echo ./$foo_interp_missing/etc $(echo foo_interp_missing bar_interp_missing)
+foo_interp_missing="bar_interp_missing"
 x=y :
 fn() {
 	local a=b
@@ -64,7 +64,7 @@ fn() {
 		echo $i | cat
 	done
 }
-[[ $foo == bar ]] && fn
+[[ $foo_interp_missing == bar_interp_missing ]] && fn
 echo a{b,c}d *.go
 let i=(2 + 3)
 `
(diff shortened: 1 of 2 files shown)


interp/interp_test.go
```diff
@@ -329,58 +329,58 @@ var runTests = []runTest{
...
-		"echo $'foo\\x00bar'",
-		"foo\n",
+		"echo $'foo_interp_does_not_exist\\x00bar_interp_does_not_exist'",
+		"foo_interp_does_not_exist\n",
 	},
 	{
-		"echo $'foo\\xbar'",
-		"foo\xbar\n",
+		"echo $'foo_interp_does_not_exist\\xbar_interp_does_not_exist'",
+		"foo_interp_does_not_exist\xbar_interp_does_not_exist\n",
```
stoovon, line 362:
```suggestion
		"interp_does_not_exist\xinterp_does_not_exist_alt\n",
```

Note that such a form wouldn't work because `\xin` is a forbidden sequence (`\x must be followed by two hexadecimal digits`). `\xbar` is acceptable.

Hence we use `foo_interp_does_not_exist` rather than `interp_does_not_exist`

mvdan: Huh, I never intended hex escape codes to expand here. Can we use `\\x` in the second string as well?

mvdan (approved):
Thanks! The names are very long, but I can't object against being explicit :)

mvdan:
Friendly nudge :)

interp/interp_test.go
stoovon: > Huh, I never intended hex escape codes to expand here. Can we use `\\x` in the second string as well?

Unfortunately not:

```
    interp_test.go:3242: wrong output in "echo $'foo_interp_does_not_exist\\xbar_interp_does_not_exist'":
        want: "foo_interp_does_not_exist\\xbar_interp_does_not_exist\n"
        got:  "foo_interp_does_not_exist\xbar_interp_does_not_exist\n"
```

stoovon:
> Thanks! The names are very long, but I can't object against being explicit :)

Can swap to `foo_interp_nexist` instead? Would need to iterate through the tests but shouldn't take long. What do you think?

mvdan:
not a bad idea actually - how about `foo_interp_missing`?

stoovon:
> not a bad idea actually - how about `foo_interp_missing`?

Done in 2bb7c381 👍🏻

mvdan (approved):
Thanks for bearing with me!

stoovon:
> Thanks for bearing with me!

Happy to help. Just found a slightly over-zealous one (I think `barks` is OK but `bar_interp_missingks` isn't). So that's now up in b287052.

PR #825, opened 2022-03-14
Add Stat field to expand.Config struct

### LINKED ISSUE #824: expand: Config stubs out ReadDir but Config.glob calls os.Stat directly
So I can give a custom `ReadDir` that talks to a remote server, or references a map (for caching or unit testing), but `os.Stat` breaks all of those ideas.

theclapp:
Config.glob used to call os.Stat directly.  Now it calls Config.Stat if it's set, and if not, calls cfg.Readdir on the given file's directory, and returns the entry for the file.

Fixes #824.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -14,6 +14,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"mvdan.cc/sh/v3/pattern"
 	"mvdan.cc/sh/v3/syntax"
@@ -56,6 +57,12 @@ type Config struct {
 	// ReadDir is used for file path globbing. If nil, globbing is disabled.
 	// Use ioutil.ReadDir to use the filesystem directly.
 	ReadDir func(string) ([]os.FileInfo, error)
+
+	// Stat is used as an optimization, to get an os.FileInfo for a single
+	// file. Only needed for Fields, and then only if ReadDir is set. If Stat
+	// is not set, ReadDir is called on name's parent directory. Use os.Stat to
+	// use the filesystem directly.
+	Stat func(name string) (os.FileInfo, error)
 
 	// GlobStar corresponds to the shell option that allows globbing with
 	// "**".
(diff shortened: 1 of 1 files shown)


expand/expand.go
```diff
@@ -835,6 +842,28 @@ func (cfg *Config) glob(base, pat string) ([]string, error) {
...
 }
 
+// stat calls cfg.Stat if it's set, otherwise calls cfg.ReadDir on name's
+// parent directory, and returns the entry for name.
+func (cfg *Config) stat(name string) (os.FileInfo, error) {
+	if cfg.Stat != nil {
+		return cfg.Stat(name)
+	}
+
+	dir, basename := filepath.Split(name)
```
mvdan, line 852: Trying to think if there's any edge case we need to worry about for `filepath.Split`, like `/`. I think not, because the shortest globs possible are `*`, which will use PWD, and `/*`, where we'll start the globbing after the root as per `IsAbs` above.

I would write a brief bit of docs here, along the lines of "we don't need to worry about name being the root of the filesystem".

expand/expand.go
```diff
@@ -57,6 +58,12 @@ type Config struct {
 	// Use ioutil.ReadDir to use the filesystem directly.
 	ReadDir func(string) ([]os.FileInfo, error)
 
+	// Stat is used as an optimization, to get an os.FileInfo for a single
+	// file. Only needed for Fields, and then only if ReadDir is set. If Stat
+	// is not set, ReadDir is called on name's parent directory. Use os.Stat to
+	// use the filesystem directly.
```
mvdan, line 64: I'm not sure if I agree with the docs. If ReadDir is set, Stat is about optimizing, but if ReadDir isn't set, it's still about sandboxing.

mvdan (commented):
I'm re-reading the implementation of globbing, and starting to wonder if we really need to use Stat. We only use that for IsDir. And, presumably, one ReadDir call per directory level should be plenty to give us all the information we need.

Let me poke at it for a bit to see if we could remove the need for Stat entirely :)

expand/expand.go
theclapp: If `ReadDir` isn't set, `Stat` is never called.  It's only called in `Config.glob`, and that's only called if `ReadDir != nil`.

mvdan: You're right, I had forgotten that detail.

mvdan:
We went with https://github.com/mvdan/sh/pull/826 instead :) Thanks again for your help!

PR #826, opened 2022-03-14
expand: replace os.Stat with cfg.ReadDir

### LINKED ISSUE #824: expand: Config stubs out ReadDir but Config.glob calls os.Stat directly

mvdan:
(see commit message)

Fixes #824.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -5,8 +5,10 @@ package expand
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -14,6 +16,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"mvdan.cc/sh/v3/pattern"
 	"mvdan.cc/sh/v3/syntax"
@@ -762,6 +765,13 @@ func (cfg *Config) glob(base, pat string) ([]string, error) {
 	}
 	// TODO: as an optimization, we could do chunks of the path all at once,
 	// like doing a single stat for "/foo/bar" in "/foo/bar/*".
+
+	// TODO: Another optimization would be to reduce the number of ReadDir calls.
+	// For example, /foo/* can end up doing one duplicate call:
+	//
+	//    ReadDir("/foo") to ensure that "/foo/" exists and only matches a directory
+	//    ReadDir("/foo") glob "*"
+
 	for i, part := range parts {
 		wantDir := i < len(parts)-1
 		switch {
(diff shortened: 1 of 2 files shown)


mvdan:
cc @theclapp

mvdan:
I broke windows, because of course I did. But it shouldn't be too hard to fix it.

mvdan:
Fun: I ran into an upstream bug relating to "not a directory" and "does not exist" being confused in Go's `os` package: https://github.com/golang/go/issues/46734#issuecomment-1066684134

theclapp (approved):
LGTM.  In case it matters.  :)

mvdan:
I'm puzzled why we fail on 1.16.x; it works on 1.16.x on Wine for me, even though it failed on 1.17.x on Wine before the fix.

mvdan:
Ah, right, Wine and Windows format the error code as different strings. How fun.

mvdan:
> LGTM. In case it matters. :)

Thanks for the review? Thoughts on this patch versus the other? I lean towards this one for the reasons outlined in my commit message. The fault is on me for starting to use `os.Stat` a long time ago, not with your patch.

theclapp:
Oh, I definitely prefer this one.  I did notice that you only call stat to see if the match is a directory, but it didn't occur to me to call ReadDir on the match to find out the same thing.  I prefer this code to adding a Stat slot.

I'm only sad that we have to read the whole directory only to throw away the results.  In the case of very large directories, it could be a problem.  But I'm willing to cross that bridge when and if we get there.  In general I suspect that if you're stat-ing a directory there's a high chance that you'll want to ReadDir it for real shortly thereafter, so a little caching (which I'm doing, on my end) should solve the problem.

mvdan:
I actually tried to use a ReadDir call on the parent, matching the FileInfo via a linear search on each name. But that actually broke on Windows due to short paths, as you can end up with two paths (normal and short) both referring to the same file, so string matching may not work.

There is an optimal setup with either Stat+ReadDir or just ReadDir, but both have slightly bad edge cases:

1) With Stat+ReadDir, if you do `/foo/*`, you need to call `Stat` on `/foo` to ensure it exists and it's a directory, then `ReadDir` on it as well to glob, even though the latter already tells you the "stat" info you need.
2) With just ReadDir, if you do `/*/bar/`, you have to call `ReadDir` on `/foo/bar` to make sure it's a directory, even though you don't care about its contents at all - not even for further globbing. Relatively rare.

All in all, I doubt this is going to be a noticeable slow-down either way. The second case with huge directories should hopefully be very rare. But if anyone does find it to be too slow, we can always reconsider adding a `Stat` optional optimization.

Thanks for your help!

PR #830, opened 2022-03-23
Better docker run example

alenzen:
I ran into some file permission and word splitting issues when using the -w flag in a directory with spaces in the name. I hope you find this pull request / documentation example useful.


mvdan (approved):
Nicely spotted. I've done the same in dockexec too: https://github.com/mvdan/dockexec/blob/81dca586c573eee268fd20b8b2763989de3a5265/main.go#L142-L143

PR #831, opened 2022-03-28
interp: surface exit codes from command substitutions

### LINKED ISSUE #828: interp: exit codes not working correctly with some substitutions
Hey there,

Originally reported at https://github.com/go-task/task/discussions/687

Given this code:

```bash
test=$(exit 255)
echo $?
```

In Bash, it correctly returns the right exit code (255 in this case). But with this lib, it's always 0.

Is this a by design decision to not support this or a possible bug?

mvdan:
(see commit message)

Fixes #828.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -81,6 +81,8 @@ type Runner struct {
 	ecfg *expand.Config
 	ectx context.Context // just so that Runner.Subshell can use it again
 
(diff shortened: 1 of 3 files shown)


mvdan:
cc @andreynering @ghostsquad

andreynering (approved):
Thanks @mvdan!

PR #837, opened 2022-04-03
syntax: strip escaped newlines from all non-quoted words

### LINKED ISSUE #832: syntax: the printer can split up strings when joined with escaped newlines
I came across this kind of assignment in a legacy codebase:

```sh
# Has the same effect as: wrapped='a b c'
wrapped=\
"a "\
"b "\
"c"
```

which `shfmt` turns into:
```sh
wrapped="a "\
	"b "\
	"c"
```

and causes the runtime error: `line 4: b : command not found`.

Tested on `GNU bash, version 5.1.4(1)-release (x86_64-pc-linux-gnu)` and the latest `shfmt`.

mvdan:
(see commit message)

Fixes #832.

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -583,16 +583,15 @@ func (p *Printer) wordParts(wps []WordPart, quoted bool) {
 		if i+1 < len(wps) {
 			next = wps[i+1]
 		}
+		// Keep escaped newlines separating word parts when quoted.
+		// Note that those escaped newlines don't cause indentaiton.
+		// When not quoted, we strip them out consistently,
+		// because attempting to keep them would prevent indentation.
 		// Can't use p.wantsNewline here, since this is only about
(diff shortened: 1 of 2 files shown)


mvdan:
cc @cilki

syntax/printer.go
```diff
@@ -583,16 +583,15 @@ func (p *Printer) wordParts(wps []WordPart, quoted bool) {
 		if i+1 < len(wps) {
 			next = wps[i+1]
 		}
+		// Keep escaped newlines separating word parts when quoted.
+		// Note that those escaped newlines don't cause indentaiton.
```
theclapp, line 587: "indentaiton"

mvdan: thainks :)

mvdan:
Thanks both for the reviews! Merging.

Issue #658, opened 2021-01-25
v4: remove syntax.KeepPadding and shfmt -kp


mvdan:
The point of this feature was to make using `shfmt` more reasonable for large existing codebases, which might be using custom formatting with regards to alignment or padding. For example:

```
func one   two   three
func four  five  six
func seven eight nine
```

The intent was good, but I think the idea is ultimately not a great one. I've spent a significant amount of code and time towards this feature, and it's still riddled with bugs and edge cases that misbehave ([1](https://github.com/mvdan/sh/issues/362), [2](https://github.com/mvdan/sh/issues/652)). Someone also proposed [redesigning the entire feature](https://github.com/mvdan/sh/issues/595), which might count as a breaking change in itself.

I think it's time to reconsider whether this flag is a good idea at all. I already outlined the main advantage above. Now to list the disadvantages:

* Extra cost in size, complexity, maintenance, and bugs
* More formatting flags tend to mean less consistency in format and a worse UX for the average user
* Conflicts with core features such as consistent spacing between tokens and vertical comment alignment
* Not clear how the feature should work; should it count column offsets, or [relative offsets](https://github.com/mvdan/sh/issues/595)? Should it do [escaped newlines](https://github.com/mvdan/sh/issues/498) too?

All in all, this is a common source of bugs and maintenance pain, and honestly I have never used the feature myself. A good portion of the `-kp` users stopped using the flag after they realised the downsides, and overall they got better formatting too; [here's an example](https://github.com/Bash-it/bash-it/pull/1810).

So I propose to entirely remove this feature in v4. Anyone who wants to keep using this feature, I'd ask you to give the formatter an honest try without the flag. If any programs get formatted in a clearly worse way, I'd like to hear about those - perhaps we can fix those cases independently of `-kp`.

I'll point the existing KeepPadding threads towards this one for input. Thanks for reading!

mvdan:
As for breaking existing users, it's clear this would be a breaking change, hence why it cannot happen in the current v3. If a v4 removes this feature, attempting to use the command-line flag or EditorConfig option would result in an error.

It's also worth noting we could always bring the feature back in the future in some other form, if we so wished.

cornfeedhobo:
I support this move, given the experience we're going through with bash-it. If I had more time, I'd maybe take a look at re-writing this feature, but I just don't have the time I wish I did. Thanks for the hard work.

maximbaz:
[...]

PR #842, opened 2022-04-05
Separate comment from sub-shell opening parenthesis with a space

### LINKED ISSUE #839: syntax: comments aren't spaced when following parentheses
### Discussed in https://github.com/mvdan/sh/discussions/838

<div type='discussions-op-text'>

<sup>Originally posted by **antichris** April  3, 2022</sup>
Is there a specific design reason why the space preceding a comment is removed when it follows an open parenthesis, or is it by accident?

Given the input

```sh
#!/bin/sh

foo() { ## (bar, [baz])
	:
} ## qux

corge() ( ## (grault, [garply])
	:
) ## waldo

_=$( ## lorem ipsum
	:
) ## dolor sit amet
```

`shfmt` yields

```sh
#!/bin/sh

foo() { ## (bar, [baz])
        :
} ## qux

corge() (## (grault, [garply])
        :
) ## waldo

_=$(## lorem ipsum
        :
) ## dolor sit amet
```
</details></div>

antichris:
Also applies to sub-shell functions and command substitutions (including the deprecated legacy back-tick syntax).

Fixes #839.

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1392,6 +1392,14 @@ func (p *Printer) assigns(assigns []*Assign) {
 	p.decLevel()
 }
 
+type wantSpaceState uint8
+
+const (
+	spaceNotRequired wantSpaceState = iota
+	spaceRequired                   // we should generally print a space or a newline next
+	spaceWritten                    // we have just written a space or newline
+)
+
 // extraIndenter ensures that all lines in a '<<-' heredoc body have at least
 // baseIndent leading tabs. Those that had more tab indentation than the first
 // heredoc line will keep that relative indentation.
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1285,41 +1285,6 @@ func (p *Printer) ifClause(ic *IfClause, elif bool) {
 	p.semiRsrv("fi", ic.FiPos)
 }
 
-func startsWithLparen(node Node) bool {
-	switch node := node.(type) {
-	case *Stmt:
-		return startsWithLparen(node.Cmd)
-	case *BinaryCmd:
-		return startsWithLparen(node.X)
-	case *Subshell:
-		return true // keep ( (
-	case *ArithmCmd:
-		return true // keep ( ((
-	}
-	return false
-}
-
-func startsWithComment(node Node) bool {
-	var stmts []*Stmt
-	switch node := node.(type) {
-	case *CmdSubst:
-		if node.TempFile || node.ReplyVar {
-			return false
-		}
-		stmts = node.Stmts
-	case *Subshell:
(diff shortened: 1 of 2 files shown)


antichris:
Truth to be told, even though it gets the job done, this feels more like a dirty hack than a hard enforcement of the rule "a comment may only be directly preceded by white-space". That, I suspect, would require more extensive changes.

antichris:
<details><summary>I'm pretty sure this is not really for a this issue and/or PR, but...</summary>

1. `spacePad()` https://github.com/mvdan/sh/blob/ba7908ece540171e3bfd3a6f7f13f7fd2591afaa/syntax/printer.go#L295-L298 could probably use `p.space()` instead:

	```go
		if p.wantSpace { // if p.wantSpace == spaceRequired {
			p.space()
		}
	```

2. `space()` itself could be implemented in terms of `spaces()`:

	```diff
	func (p *Printer) spaces(n uint) {
		for i := uint(0); i < n; i++ {
			p.WriteByte(' ')
		}
	+	p.wantSpace = false // p.wantSpace = spaceWritten
	}

	func (p *Printer) space() {
	-	p.WriteByte(' ')
	-	p.wantSpace = false // p.wantSpace = spaceWritten
	+	p.spaces(1)
	}
	```

3. the padding code of `spacePad()` https://github.com/mvdan/sh/blob/ba7908ece540171e3bfd3a6f7f13f7fd2591afaa/syntax/printer.go#L299-L301 could also be rewritten to use `spaces()`:

	```go
		if p.cols.column > 0 {
			padding := pos.Col() - uint(p.cols.column)
			if padding < pos.Col() {
				p.spaces(padding)
			}
		}
	```

	although this one doesn't really feel like *that* much of an improvement to me.

</details>

I'm not sure if it warrants opening an issue, but I felt like it's worth mentioning.

<details><summary>Also, there's something funky going on with sub-shell trailing comments.</summary>

For example:

```sh
( #foo
        bar #qux
) #corge
```

gets turned into

```sh
( #foo
        bar #qux
)    #corge
```

With `keepPadding` such an input passes unchanged. Almost as if the meaning of `keepPadding` is inverted for those trailing comments. I'm guessing this will probably go away once #658 gets dealt with.
</details>

mvdan (approved):
Nice, absolutely nothing to add here!

mvdan:
Squashed and tweaked your commit message a bit, hope that's okay.

antichris:
Thanks, that's perfect — I would've suggested squashing as well. I usually try to stick to making one logical change per commit as I work, because it's tricky (when not impossible) to pry them apart later, if need be, but squashing them together when rebasing onto a main branch is always a breeze.

PR #848, opened 2022-04-14
interp: use custom handler for os.stat/os.Lstat operations

ihar-orca:
This feature allows intercepting the os.stat/os.Lstat operations in the interp to provide more flexible usage.

### DIFF
--- a/interp/handler.go
+++ b/interp/handler.go
@@ -318,12 +318,11 @@ func DefaultReadDirHandler() ReadDirHandlerFunc {
 
 // StatHandlerFunc is a handler which gets the file stat. the first argument provides directory to use as
 // basedir if name is relative path
-type StatHandlerFunc func(ctx context.Context, cwd string, name string, followSymlinks bool) (os.FileInfo, error)
+type StatHandlerFunc func(ctx context.Context, name string, followSymlinks bool) (os.FileInfo, error)
 
 // DefaultStatHandler returns a StatHandlerFunc used by default. It uses os.Stat()
 func DefaultStatHandler() StatHandlerFunc {
-	return func(ctx context.Context, cwd string, name string, followSymlinks bool) (os.FileInfo, error) {
-		path := absPath(cwd, name)
+	return func(ctx context.Context, path string, followSymlinks bool) (os.FileInfo, error) {
 		if !followSymlinks {
 			return os.Lstat(path)
 		} else {
--- a/interp/handler.go
+++ b/interp/handler.go
@@ -315,3 +315,15 @@ func DefaultReadDirHandler() ReadDirHandlerFunc {
 		return ioutil.ReadDir(path)
 	}
 }
+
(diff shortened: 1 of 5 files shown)


mvdan:
FYI @alecthomas given https://github.com/mvdan/sh/pull/799.

interp/handler.go
```diff
@@ -315,3 +315,19 @@ func DefaultReadDirHandler() ReadDirHandlerFunc {
 		return ioutil.ReadDir(path)
 	}
 }
+
+// StatHandlerFunc is a handler which gets the file stat. the first argument provides directory to use as
+// basedir if name is relative path
+type StatHandlerFunc func(cwd string, name string, lstat bool) (os.FileInfo, error)
```
mvdan, line 321: any reason to pass two strings? the handler could just receive one absolute path, like e.g. ReadDirHandlerFunc.

mvdan: I would also invert the boolean and call it `followSymlinks` instead; then the default `false` behavior is to do less work, and the behavior is overall more intuitive.

ihar-orca:
@mvdan in general, yes. I have a couple of shell scripts that run on a target Linux machine. this script checks some system integrity. the idea is to run these scripts on the target machine but on mounted volume without any chrooted environment. to achieve this I should mock some apps (I can do it using execHandler), mock dirs and files functions (stat, opendirhandler, substitute any paths in variables/arguments using the parser).

interp/handler.go
ihar-orca: I need to pass the `cwd` argument because the origin function uses the `r.Dir` to build an absolute path in the case when the `name` contains a relative path.

mvdan: That's fine, but please move the cwd stuff to the caller - there's no reason to make every implementer of the handler deal with it. They should just receive the absolute path. We can make the call to absPath in Runner.stat and Runner.lstat.

interp/runner.go
```diff
@@ -916,5 +916,9 @@ func (r *Runner) open(ctx context.Context, path string, flags int, mode os.FileM
 }
 
 func (r *Runner) stat(name string) (os.FileInfo, error) {
```
codefromthecrypt, line 918: curious if you are thinking these would never benefit from a ctx param or not (thinking log correlation)

mvdan: Good catch @codefromthecrypt; StatHandlerFunc should take a context, even if just for consistency with the other handler funcs.

ihar-orca: Agreed. Now stat handler contains a context.

interp/handler.go
ihar-orca: I moved absPath in the parent functions, cwd arg is dropped from the handler func.

ihar-orca:
Now the StatHandlerFunc contains a Context. `cwd` argument removed and replaced with the absPath constructor in the parent function.

mvdan (approved):
Thanks!

PR #853, opened 2022-04-29
syntax: write shebang in minify mode

### LINKED ISSUE #833: syntax: Minify discards shebangs
Running a simple script like:
```sh
#!/bin/bash
echo a
```
through `shfmt -mn` yields:

```sh
echo a
```

The shebang is stripped out (like a comment) which isn't ideal.

Thanks for the otherwise great library!

hristiy4n:
Fixes #833

### DIFF
--- a/fileutil/file.go
+++ b/fileutil/file.go
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	shebangRe = regexp.MustCompile(`^#!\s?/(usr/)?bin/(env\s+)?(sh|bash|mksh|bats|zsh)\s`)
+	shebangRe = regexp.MustCompile(`^#!\s?/(usr/)?bin/(env\s+)?(sh|bash|mksh|bats|zsh)(\s|$)`)
 	extRe     = regexp.MustCompile(`\.(sh|bash|mksh|bats|zsh)$`)
 )
 
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"text/tabwriter"
 	"unicode"
+
+	"mvdan.cc/sh/v3/fileutil"
 )
 
 // PrinterOption is a function which can be passed to NewPrinter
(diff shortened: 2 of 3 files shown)


syntax/nodes.go
```diff
@@ -168,6 +170,7 @@ type Comment struct {
 
 func (c *Comment) Pos() Pos { return c.Hash }
 func (c *Comment) End() Pos { return posAddCol(c.Hash, 1+len(c.Text)) }
+func (c *Comment) IsShebang() bool { return fileutil.Shebang([]byte("#"+c.Text)) != "" && c.Hash.Col() == 1 && c.Hash.Line() == 1 }
```
mvdan, line 173: I would prefer to not add new API here; `fileutil.Shebang` is always available to end users. You can move the three conditions into printer.go.

fileutil/file.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	shebangRe = regexp.MustCompile(`^#!\s?/(usr/)?bin/(env\s+)?(sh|bash|mksh|bats|zsh)\s`)
+	shebangRe = regexp.MustCompile(`^#!\s?/(usr/)?bin/(env\s+)?(sh|bash|mksh|bats|zsh)\s?`)
```
mvdan, line 16: The `\s` there is for good reason; see 31883e5e05b63c0a5a8f97e90d3f0576fbc86fec. Why add the `?` at the end? Note that we don't have a trailing `$` either.

mvdan (commented):
This needs tests :) In particular, it would be good to also include a test case where a shebang comment is not at the start of the file. See TestPrintMinify.

fileutil/file.go
mvdan: Ah, right. The regular expression was only used on the unparsed file bytes before, so the comment would pretty much always be followed by a newline - which is also whitespace. Since you're using it on a parsed comment now, which omits the newline, then there's no whitespace.

I think you want `(\s|$)` then - because otherwise `\s?` will mean we match `#!/usr/bin/shoo`, as `sh`, whereas we only want to match `#!/usr/bin/sh some args` or `#!/usr/bin/sh`.

mvdan (approved):
Thank you!

PR #859, opened 2022-05-08
cmd/shfmt: support long flags

### LINKED ISSUE #852: cmd/shfmt: longer CLI flags
Apologies if this is a duplicate, but would you consider adding a longer version of each CLI flag e.g -i -> --indent. It increases readability and is therefore a fairly common practice. Thanks!

mvdan:
(see commit message)

Fixes #852.

### DIFF
--- a/cmd/shfmt/Dockerfile
+++ b/cmd/shfmt/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.17.1-alpine AS build
 WORKDIR /src
 RUN apk add --no-cache git
 COPY . .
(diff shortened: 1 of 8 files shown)


mvdan:
cc @p0da if you can take a look :)

andreynering (approved):
You could benefit from using `pflag` (https://github.com/spf13/pflag) which is a non-dependencies package which uses the same API as the stdlib's `flag` package.

But that's your choice, of course.

mvdan:
It's nearly six thousand lines of code, so I'm reluctant to pull it for the sake of convenience :) I don't _really_ need a reimplementation of a flag parser.

PR #861, opened 2022-05-12
Enable multiline regexp mode for patterm matcher

ihar-orca:
In the case of pattern comparison with the multiline string, an incorrect result may occur when the matched substring appears in the not first line.

Example:
```
#!/bin/bash

apache2_conf="/etc/httpd/conf/httpd.conf
/etc/apache2/httpd.conf"

if [[ "$apache2_conf" == */etc/apache2* ]]; then
  exit 0
else
  exit 1
fi

```
returns 1 because generated regexp `^.*/etc/apache2.*$` (interp/runner.go, func match()) is not multiline enabled.
this patch adds the `(?m)` flag into all autogenerated regexps in this function.


ihar-orca:
Sorry for multiple commits, it seems that it's not squashed on my side

ihar-orca:
@mvdan test added

mvdan (approved):
Thank you!

mvdan:
> This branch cannot be rebased due to conflicts

Not sure how you ended up there. Can you try squashing and rebasing your commits?

mvdan:
Ah, ignore me, I can squash-merge.

PR #862, opened 2022-05-16
interp: Return invalid globs when nullglob is set

lollipopman:
Prior to this commit invalid globs were returned as null when nullglob
was set, which resulted in test expressions failing:

  $ shopt -s nullglob; [ -n butter ] && echo bubbles
  "-n": executable file not found in $PATH

After this commit invalid globs are returned verbatim when nullglob is
set.


mvdan:
I'm a bit confused by your commit message: `butter` is not an invalid glob, as far as I can tell.

mvdan:
Ah, I see, thanks for the clarification.

Do we need to add InvalidGlobError? It's the only error that the method ever returns, so we can avoid extending the API and simply use `err != nil` in the caller to tell whether the glob is valid or not.

Otherwise the change sounds good to me!

mvdan:
Err, scratch that. I hadn't noticed that the call is to the `glob` method, which then indirectly calls `pattern.Regexp`. We do need to differentiate the error. So I think what you've done is good.

lollipopman:
> Err, scratch that. I hadn't noticed that the call is to the `glob` method, which then indirectly calls `pattern.Regexp`. We do need to differentiate the error. So I think what you've done is good.

I also considered having the glob function return nil, nil when the glob is invalid and then having the caller differentiate between a nil slice and an empty slice. That method worked, but it seemed a little *too* clever and brittle.

mvdan:
Yeah, I agree that's too subtle - I don't like APIs that give different meanings to nil vs empty.

pattern/pattern.go
```diff
@@ -20,6 +20,15 @@ import (
 // Not all functions change their behavior with all of the options below.
 type Mode uint
 
+type InvalidGlobError struct {
```
mvdan, line 23: How about `pattern.SyntaxError` rather than `pattern.InvalidGlobError`? I don't think we will ever need any kind of "invalid pattern" error that isn't related to syntax, and I would rather avoid the name "glob" as the API is centered around the name "pattern" already - especially in the package name.

mvdan (approved):
SGTM with a naming nit

lollipopman:
Thanks for the careful review @mvdan, I believe this is now ready to merge.

PR #863, opened 2022-05-18
interp: add initial mapfile or readarray builtin

lollipopman:
Add an initial implementation of mapfile or readarray, this version only
supports the basic options of `-d`, `-t`, and a custom array variable
name. At present it only support splitting on ASCII delimiters.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -4,6 +4,7 @@
 package interp
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


interp/builtin.go
```diff
@@ -800,13 +801,95 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
...
+		return 0
+
 	default:
 		// "umask", "fg", "bg",
 		panic(fmt.Sprintf("unhandled builtin: %s", name))
 	}
 	return 0
 }
 
+// Returns a suitable Split function for a bufio.Scanner, the code is mostly
```
mvdan, line 869: use the godoc style:

> // mapFileSplit returns ...

and also end with a period

mvdan (approved):
Thanks! Just some nits.

I take it that these are fairly Bash-specific, which is fine.

interp/builtin.go
lollipopman: fixed, thanks

lollipopman:
Thanks for the careful review @mvdan, I believe this is now ready to merge.

PR #865, opened 2022-05-23
read builtin: implement -a, -s, -n, -N and -d

lollipopman:
This implements some of the commonly used options for the read builtin,
with the notable exception of -t.

Questions:

1. I don't love the duplication of some of the x/term code, but I
   couldn't find another option?
2. At present -s is completely untested, any suggestions on where or how
   to test code, which relies on interacting with a terminal?

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -12,9 +12,12 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
+	"golang.org/x/sys/unix"
+	"golang.org/x/term"
 	"mvdan.cc/sh/v3/expand"
 	"mvdan.cc/sh/v3/syntax"
 )
@@ -557,7 +560,12 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 		r.setErr(returnStatus(code))
 	case "read":
 		var prompt string
+		var readArrayName string
 		raw := false
+		delim := []byte("\n")
+		termNoEcho := false
+		numBytes := -1
+		ignoreDelim := false
 		fp := flagParser{remaining: args}
 		for fp.more() {
 			switch flag := fp.flag(); flag {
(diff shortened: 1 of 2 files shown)


mvdan:
Do we just need to copy the code to access unexported APIs, or do we need to copy bits of code to tweak them? If it's just the former, I would consider something like https://pkg.go.dev/golang.org/x/tools/cmd/bundle.

mvdan:
Do we just need to copy the code to access unexported APIs, or do we need to copy bits of code to tweak them? If it's just the former, I would consider something like https://pkg.go.dev/golang.org/x/tools/cmd/bundle.

lollipopman:
> Do we just need to copy the code to access unexported APIs, or do we need to copy bits of code to tweak them? If it's just the former, I would consider something like https://pkg.go.dev/golang.org/x/tools/cmd/bundle.

At present only the former, namely:

```
// These platform dependent definitions:
const ioctlReadTermios = unix.TCGETS
const ioctlWriteTermios = unix.TCSETS
```

Bundle looks neat, but I don't think it would work, as it has this caveat: "must not use conditional file compilation", which I assume won't work with build tags such as `//go:build darwin || dragonfly || freebsd || netbsd || openbsd` inside term's source?

mvdan:
Ah, that's unfortunate. The bundle tool outputs a single filename, so I guess it's only logical that it has that limitation given that it's a simple tool.

What about adding the API in x/term that we need here? It could take weeks or even months for such an API proposal to be accepted and merged, but at least we wouldn't be copy-pasting chunks of code.

For now, though, manual copying seems fine. I would just ask that you document it well, so that keeping the code up to date is reasonably easy in the future.

lollipopman:
> What about adding the API in x/term that we need here? It could take weeks or even months for such an API proposal to be accepted and merged, but at least we wouldn't be copy-pasting chunks of code.

Yeah, I think that is worthwhile, I'll try to get a pull request against x/term for those folks to peruse.

> For now, though, manual copying seems fine. I would just ask that you document it well, so that keeping the code up to date is reasonably easy in the future.

sounds good, I'll take that route after making the aforementioned pull request.

mvdan:
Friendly ping, do you intend to get back to this? It's gained conflicts since we last looked at it :)

lollipopman:
@mvdan thanks for the reminder, I lost steam on getting patches created for x/term, but you have re-inspired me!

mvdan:
I'm going to close this for now due to its age. Please feel free to open a new PR if you would like to revive this and I'll happily take a look.

FYI, this gained conflicts in part because https://github.com/mvdan/sh/pull/1063 did a subset of these changes.

PR #866, opened 2022-05-25
interp: fix multi-line shell pattern match equality

lollipopman:
commit be03afe7282aea8007fd3e961609a3b1f077db1d added the mult-line flag
to the regexp generateed for `==` and `=!` testing. However, this flag
is confusing as it allows for a regular expression to match one line out
of a multi-line input.

  (m) multi-line mode: ^ and $ match begin/end line in addition to begin/end
  text (default false)

This means we are now matching against just one line in a multi-line
string, rather than the entire text. Instead, to match Bash's behavior,
we add the `s` flag to the regexp:

  (s) let . match \n (default false)

With this a `*` in a shell pattern will match `\n`


ihar-orca:
@lollipopman looks really good!

pattern/pattern.go
```diff
@@ -59,11 +60,16 @@ noopLoop:
...
 		}
 	}
-	if !any { // short-cut without a string copy
+	if !any && mode&EntireString == 0 { // short-cut without a string copy
 		return pat, nil
 	}
 	closingBraces := []int{}
 	var buf bytes.Buffer
+	// Enable matching `\n` with `.` as globs match `\n`
+	buf.WriteString("(?s)")
```
mvdan, line 69: I think this means we can never take the `return pat, nil` shortcut above, unfortunately - the pattern `foo*bar` needs to be able to match the string `foo\nbar` with a newline between the two words when EntireString isn't set, so we want the resulting regular expression to be `(?s)foo.*bar` rather than just `foo.*bar`.

mvdan (commented):
Also thinking about the added API - it seems like we do need the added API, because in some cases we want the partial matching to happen. Or perhaps not - if one needs partial matching, perhaps `foobar` could be fed in as `*foobar*`, then it will still work as expected.

pattern/pattern.go
lollipopman: I can't seem to craft a test case where the shortcut fails. Go's regexp does in general traverse newlines, the exception is that the special character `.` does not by default. In the shortcut case we are returning patterns which could never contain the `.` character, so we don't need to add the `s` flag, correct?

lollipopman:
> Also thinking about the added API - it seems like we do need the added API, because in some cases we want the partial matching to happen. Or perhaps not - if one needs partial matching, perhaps `foobar` could be fed in as `*foobar*`, then it will still work as expected.

I had not thought about inverting the need for whole string matches, by converting each pattern into a whole string match as you suggest. I tried doing some rough benchmarks in Go and whole string matches appear significantly slower on larger texts, as I would expect.

```
func BenchmarkPartialString(b *testing.B) {
	b.ReportAllocs()
	out, err := exec.Command("man", "bash").Output()
	if err != nil {
		log.Fatal(err)
	}
	manBash := string(out)
	partial := regexp.MustCompile("COPYRIGHT")
	for i := 0; i < b.N; i++ {
		if !partial.MatchString(manBash) {
			b.Fatal(err)
		}
	}
}
func BenchmarkWholeString(b *testing.B) {
	b.ReportAllocs()
	out, err := exec.Command("man", "bash").Output()
	if err != nil {
		log.Fatal(err)
	}
	manBash := string(out)
	wholeString := regexp.MustCompile("^(?s).*COPYRIGHT.*$")
	for i := 0; i < b.N; i++ {
		if !wholeString.MatchString(manBash) {
			b.Fatal(err)
		}
	}
}
```

pattern/pattern.go
mvdan: Of course you're right :)

pattern/pattern_test.go
```diff
@@ -18,65 +18,65 @@ var translateTests = []struct {
...
+	{pat: `{3,4}`, want: `(?s)\{3,4\}`},
+	{pat: `{3,4}`, mode: Braces, want: `(?s)(?:3|4)`},
+	{pat: `{3,`, want: `(?s)\{3,`},
+	{pat: `{3,`, mode: Braces, want: `(?s)\{3,`},
+	{pat: `{3,{4}`, mode: Braces, want: `(?s)\{3,\{4\}`},
+	{pat: `{3,{4}}`, mode: Braces, want: `(?s)(?:3|\{4\})`},
+	{pat: `{3,{4,[56]}}`, mode: Braces, want: `(?s)(?:3|(?:4|[56]))`},
+	{pat: `{3..5}`, mode: Braces, want: `(?s)(?:3|4|5)`},
+	{pat: `{9..12}`, mode: Braces, want: `(?s)(?:9|10|11|12)`},
+	{pat: `[a]`, want: `(?s)[a]`},
```
mvdan, line 50: nit: patterns like these don't need `(?s)` because they do not use `.`. I understand that it probably doesn't make a difference in terms of performance, but as an API, it would be nice if `pattern.Regexp` doesn't return extra characters unless it really needs to.

One option is to always write `(?s)` into the buffer, and slice the prefix out of the returned string if we didn't write `.` nor `.*`.

Then the tests won't change as much either :)

mvdan (approved):
Very nice patch :) Thanks for bearing with me as I review this - it took me a while as I've been travelling.

More than happy to review more patches from you if they are of this quality.

pattern/pattern_test.go
lollipopman: Thanks for pointing this out, I wasn't in love with my approach either. I have taken your suggestion which definitely cleans up the diff. Hopefully, this won't cause too much consternation when I get back to working on my patch for nocaseglob 😉.

lollipopman:
> Very nice patch :) Thanks for bearing with me as I review this - it took me a while as I've been travelling.
> 
> More than happy to review more patches from you if they are of this quality.

Thanks, hopefully I will find the time to keep them coming!

mvdan (approved):

PR #867, opened 2022-05-27
shfmt: state -ln=auto on posix parse errors

### LINKED ISSUE #803: cmd/shfmt: when using -ln=auto, tell the user what language was detected on an error
I imagine some people will be confused when upgrading to the changed default if they have any scripts that contain Bash syntax but look like POSIX shell scripts. If the parser error message ends with something like `(parsed as posix via -ln=auto)`, then hopefully that would be less confusing.

riacataquian:
explicitly state language dialect on `syntax.LangError`s when ln
is set to `auto`

for example, incompatible bash constructs will cause parse errors
due to the parser using posix as language dialect

fixes #803

### DIFF
--- a/cmd/shfmt/testdata/scripts/flags.txt
+++ b/cmd/shfmt/testdata/scripts/flags.txt
@@ -98,7 +98,7 @@ stdin input-bash-arrays
 stderr 'parsed as posix via -ln=auto'
 
 stdin input-bash-extglobs
-! shfmt -ln=auto -filename=input.sh
+! shfmt -filename=input.sh
 stderr 'parsed as posix via -ln=auto'
 
 stdin flags-input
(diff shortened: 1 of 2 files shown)


cmd/shfmt/testdata/scripts/flags.txt
```diff
@@ -92,6 +92,15 @@ stdin input-mksh
...
 stdout 'coprocess'
 
+# Explicitly state language on parse errors
+stdin input-bash-arrays
+! shfmt -ln=auto -filename=input.sh
+stderr 'parsed as posix via -ln=auto'
+
+stdin input-bash-extglobs
+! shfmt -ln=auto -filename=input.sh
+stderr 'parsed as posix via -ln=auto'
```
mvdan, line 102: also add a test case without `-ln=auto`, which should behave the same as it's the default

mvdan (approved):
Nice!

cmd/shfmt/testdata/scripts/flags.txt
riacataquian: done!

riacataquian: I repurposed the other one since its testing the same thing.

mvdan: SGTM

PR #868, opened 2022-05-27
shfmt: fix typo on FormatPath's

riacataquian:
the `io.ReadAtLeast` at a minimum, checks for the shebang
directive but we're missing the `!`


mvdan (approved):
thanks!

PR #869, opened 2022-05-31
syntax: fix whitespace on nested subshells

### LINKED ISSUE #814: syntax: trailing white space added after opening parenthesis '('
In the second example below, a trailing blank/ white space is inserted after first opening '(' symbol, this is indicated through '^'. The expectation is like in first example above. This seems to happen when opening a subshell first after opening brackets of another subshell.
shfmt is called without any parameters in this case.

```
#!/bin/bash

# OK - no blank
(
    cd .
    (echo "Hello world!")
)

# nok - blank ^ after first (
(^ 
    (echo "Hello world!")
)
```

riacataquian:
we want white space on nested subshells if its in a single line since
its ambiguous, eg we want `( (` over `((`

this shouldn't be the case for multiple lines- fix the logic that adds
white space to only do so if the two subshells are in the same line

fixes #814

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1081,6 +1081,13 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
 		p.WriteByte('(')
 		if len(x.Stmts) > 0 && startsWithLparen(x.Stmts[0]) {
 			p.wantSpace = spaceRequired
+			if nested, ok := x.Stmts[0].Cmd.(*Subshell); ok {
+				// we only want to keep the space between two nested subshells' open brackets
+				// if its in a single line to avoid ambiguity
+				if x.Lparen.Line() != nested.Pos().Line() {
+					p.wantSpace = spaceNotRequired
+				}
+			}
 		} else {
 			p.wantSpace = spaceNotRequired
 		}
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1082,13 +1082,13 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
 		stmts := x.Stmts
 		if len(stmts) > 0 && startsWithLparen(stmts[0]) {
 			p.wantSpace = spaceRequired
-			// we only want to keep the space between two nested subshells' open brackets
-			// if its in a single line to avoid ambiguity
-			if x.Lparen.Line() != stmts[0].Pos().Line() || len(stmts) > 1 && !p.singleLine {
+			// Add a space between nested parentheses if we're printing them in a single line,
+			// to avoid the ambiguity between `((` and `( (`.
+			if (x.Lparen.Line() != stmts[0].Pos().Line() || len(stmts) > 1) && !p.singleLine {
 				p.wantSpace = spaceNotRequired
 
 				if p.minify {
-					p.newline(stmts[0].Pos())
+					p.mustNewline = true
 				}
 			}
 		} else {
@@ -1349,7 +1349,7 @@ func (p *Printer) stmtList(stmts []*Stmt, last []Comment) {
 			// statement.
 			p.comments(c)
 		}
-		if !p.minify || p.wantSpace == spaceRequired {
+		if p.mustNewline || !p.minify || p.wantSpace == spaceRequired {
 			p.newlines(pos)
 		}
 		p.line = pos.Line()
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1079,14 +1079,13 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
 		p.ifClause(x, false)
 	case *Subshell:
 		p.WriteByte('(')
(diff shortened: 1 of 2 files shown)


syntax/printer.go
```diff
@@ -1081,6 +1081,13 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
 		p.WriteByte('(')
 		if len(x.Stmts) > 0 && startsWithLparen(x.Stmts[0]) {
 			p.wantSpace = spaceRequired
+			if nested, ok := x.Stmts[0].Cmd.(*Subshell); ok {
```
mvdan, line 1084: I would do this as part of startsWithLparen, which already looks into the type of the first command. Moreover, it multiple node types, which you will want to do as well (add tests too, please).

syntax/printer.go
```diff
@@ -1081,6 +1081,13 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
...
+				// we only want to keep the space between two nested subshells' open brackets
+				// if its in a single line to avoid ambiguity
+				if x.Lparen.Line() != nested.Pos().Line() {
```
mvdan, line 1087: hmm, I'm not sure if this is exactly right - in some cases we'll add a newline, like:

```
$ cat f.sh
( (foo); bar)
$ shfmt f.sh | tr " \t" "*?"
(*
?(foo)
?bar
)
```

You can see that we add a space at the end of the line, which I replaced with `*` to make it visible. I think this edge case is still broken with your fix.

mvdan (commented):
Some thoughts :)

syntax/printer.go
riacataquian: thanks for the tip! I added some more scenarios and test cases.

syntax/printer_test.go
```diff
@@ -616,6 +616,24 @@ var printTests = []printCase{
...
+	{
+		"(\n(foo >redir)\n)",
+		"(\n\t(foo >redir)\n)",
+	},
+	{
+		"( (foo) )",
+		"( (foo))",
+	},
+	{
+		"( {foo}; bar; )",
```
mvdan, line 628: Did you mean to use (foo) here? That's what we want to test AFAICT

It could also be nice to include a test for the other form of ambiguity when we add a newline:

( ((foo)); bar)

mvdan (approved):

syntax/printer_test.go
riacataquian: actually, will include fix :)

riacataquian: @mvdan please see the latest [commit](https://github.com/mvdan/sh/pull/869/commits/8e4f46fc45ff1319f3cad818f5a78cd53ab2c17a) for my attempt to fix the scenario above.

syntax/printer.go
```diff
@@ -1079,11 +1079,22 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
 		p.ifClause(x, false)
 	case *Subshell:
 		p.WriteByte('(')
-		if len(x.Stmts) > 0 && startsWithLparen(x.Stmts[0]) {
+		stmts := x.Stmts
+		if len(stmts) > 0 && startsWithLparen(stmts[0]) {
 			p.wantSpace = spaceRequired
+			// we only want to keep the space between two nested subshells' open brackets
+			// if its in a single line to avoid ambiguity
+			if x.Lparen.Line() != stmts[0].Pos().Line() || len(stmts) > 1 && !p.singleLine {
```
mvdan, line 1087:
```suggestion
			if (x.Lparen.Line() != stmts[0].Pos().Line() || len(stmts) > 1) && !p.singleLine {
```

the parentheses clarify whether the "and" or "or" apply first; otherwise it's a bit hard for a human to parse

syntax/printer.go
```diff
@@ -1079,11 +1079,22 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
...
+				p.wantSpace = spaceNotRequired
+
+				if p.minify {
+					p.newline(stmts[0].Pos())
```
mvdan, line 1091: ah, so this was a bug? can we not just add it to the condition above, like singleLine?

mvdan (approved):
Tests LGTM, thanks! Just some code nits.

syntax/printer.go
```diff
@@ -1079,11 +1079,22 @@ func (p *Printer) command(cmd Command, redirs []*Redirect) (startRedirs int) {
 		p.ifClause(x, false)
 	case *Subshell:
 		p.WriteByte('(')
-		if len(x.Stmts) > 0 && startsWithLparen(x.Stmts[0]) {
+		stmts := x.Stmts
+		if len(stmts) > 0 && startsWithLparen(stmts[0]) {
 			p.wantSpace = spaceRequired
+			// we only want to keep the space between two nested subshells' open brackets
+			// if its in a single line to avoid ambiguity
```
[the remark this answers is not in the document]
riacataquian: thanks! 😄

mvdan (approved):

PR #872, opened 2022-06-05
interp: allow to register go functions

zimbatm:
As part of the long-running branch in https://github.com/direnv/direnv/tree/gosh, there are a number of bash functions that could be sped up by implementing them in Go instead.

This PR is a bit of a brute-force attempt to see if I could add the capability to this project. It seems to work but I don't know if a solution like that would be acceptable. I'm looking for an initial review to validate the overall approach before continuing on the PR.

### DIFF
--- a/interp/gocmd.go
+++ b/interp/gocmd.go
@@ -1,0 +1,48 @@
+package interp
+
+import (
+	"context"
+	"io"
+
+	"mvdan.cc/sh/v3/expand"
+	"mvdan.cc/sh/v3/syntax"
+)
+
+// DeclareGoCommand registers a Go function as if it was a bash command.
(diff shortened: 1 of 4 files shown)


mvdan:
Have you seen https://pkg.go.dev/mvdan.cc/sh/v3/interp#ExecHandler? That already exists as a way to add commands implemented in Go. See the added example; if you want to add multiple commands, you can imagine a string switch in that function.

zimbatm:
Hmm right, I missed that. I guess ExecHandler is not going to make them visible for introspection but I can live with that.

zimbatm:
thanks! :)

mvdan:
Yeah, it is opaque to the shell code, but the advantage is that you can have arbitrary logic - like implementing any command based on a prefix :)

PR #874, opened 2022-06-06
syntax: fix whitespace on nested subshells

riacataquian:
we want white space on nested subshells if its in a single line since
its ambiguous, eg we want `( (` over `((`

this shouldn't be the case for multiple lines- fix the logic that adds
white space to only do so if the two subshells are in the same line

fixes https://github.com/mvdan/sh/issues/814


mvdan (approved):
This is re-landing https://github.com/mvdan/sh/pull/869 as we merged it wrong.

PR #883, opened 2022-06-19
interp: display all Bash's `shopt` option

### LINKED ISSUE #877: interp: display all options for bash's `shopt` and explicitly state unsupported ones
We don't support all of bash's shell options, at the moment we support these three:
```
$ gosh -c "shopt"
expand_aliases	off
globstar	off
nullglob	off
```

Bash actually has more options than what's listed above, see [shopt documentation](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html) but we only print what's supported which could be misleading and unhelpful, eg when someone looks up for an option `shopt | grep extglob`, they'll get empty results.

Also trying to set an unsupported bash option leads to a confusing error message:
```
$ gosh -c "shopt -s extglob"
shopt: invalid option name "extglob"
```
We could improve the error message by explicitly saying it is instead unsupported.

riacataquian:
Trying to set an unsupported but valid Bash option leads to a
potentially confusing error message:
```
$ gosh -c "shopt -s extglob"
shopt: invalid option name "extglob"
```

Fix that by handling the unsupported options differently from the
invalid ones:
```
$ gosh -c "shopt -s extglob"
bash: line 1: shopt: extglob off ("on" not supported)
exit status 1
```

Additionally, this commit lists all of the Bash options when `shopt`
without arguments is called and explicitly identify the unsupported
options, for example:
```
$ gosh -c "shopt"
expand_aliases  off
globstar        off
nullglob        off
// .. cut for brevity
hostcomplete    on      ("off" not supported)
inherit_errexit on      ("off" not supported)
interactive_comments    on      ("off" not supported)
```

While at it, rewrite the `bashOptsTable` so that it can keep two option
states: 1) Bash's default options and 2) whether we support it

Fixes #877

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -185,6 +185,12 @@ func New(opts ...RunnerOption) (*Runner, error) {
 			return nil, err
 		}
 	}
+
+	// turn "on" the default Bash options
+	for i, opt := range bashOptsTable {
+		r.opts[len(shellOptsTable)+i] = opt.defaultState
+	}
+
 	// Set the default fallbacks, if necessary.
 	if r.Env == nil {
 		Env(nil)(r)
@@ -274,7 +280,7 @@ func Params(args ...string) RunnerOption {
 			value := fp.value()
 			if value == "" && enable {
 				for i, opt := range &shellOptsTable {
-					r.printOptLine(opt.name, r.opts[i])
+					r.printOptLine(opt.name, r.opts[i], true)
 				}
 				continue
 			}
@@ -288,7 +294,7 @@ func Params(args ...string) RunnerOption {
 				}
 				continue
 			}
-			opt := r.optByName(value, false)
+			_, opt := r.optByName(value, false)
 			if opt == nil {
 				return fmt.Errorf("invalid option: %q", value)
 			}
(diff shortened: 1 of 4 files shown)


interp/api.go
```diff
@@ -399,11 +410,108 @@ var shellOptsTable = [...]struct {
...
+		name:           "globstar",
+		supported:      true,
+		allow_override: true,
+	},
+	{
+		name:           "nullglob",
+		supported:      true,
+		allow_override: true,
+	},
+	// options that are enabled by default in Bash, sorted alphabetically by name
```
riacataquian, line 430: @mvdan if you agree, I'm thinking to also display `on` for the default Bash options (that we also implement) for the `shopt` command, which means enabling interp to determine those.

mvdan:
I also think we need to record the default values, as I'm fairly sure there are a good number of Bash options that default to "on".

riacataquian:
That sounds better, thanks! :)

Also, I agree on:
> I also think we need to record the default values anyway, as I'm fairly sure there are a good number of Bash options that default to "on".

I briefly mentioned it here https://github.com/mvdan/sh/pull/883#discussion_r901299481, will also include in this PR.

riacataquian:
@mvdan could you take a look please? I incorporated the feedbacks above. Thanks!

interp/api.go
```diff
@@ -366,28 +372,37 @@ func StdIO(in io.Reader, out, err io.Writer) RunnerOption {
...
-var shellOptsTable = [...]struct {
+type shellOpt struct {
 	flag byte
 	name string
-}{
+}
+
+type bashOpt struct {
+	name          string
+	default_state bool // Bash's default value for this option
```
mvdan, line 401:
```suggestion
	defaultState bool // Bash's default value for this option
```

interp/builtin.go
```diff
@@ -19,6 +19,12 @@ import (
 	"mvdan.cc/sh/v3/syntax"
 )
 
+// optStatuses are the option statuses' text display
+var optStatuses = map[bool]string{
```
mvdan, line 23: nit: make this a function with an if statement, there isn't really a need for a map for just two entries.

interp/builtin.go
```diff
@@ -668,29 +674,38 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
...
 			}
 			break
 		}
 		for _, arg := range args {
-			opt := r.optByName(arg, !posixOpts)
-			if opt == nil {
-				r.errf("shopt: invalid option name %q\n", arg)
+			v, _, opt := r.optByName(arg, bash)
+			if v == nil {
+				r.errf("bash: line %d: shopt: %s: invalid shell option name\n", pos.Line(), arg)
```
mvdan, line 693: I sorta disagree with using `bash:` in error lines, because we are not Bash :)

It's OK if the error lines don't match Bash exactly character by character. See `#JUSTERR` in some of the test cases; those tests check that both shells error, but doesn't require Bash to match the output exactly.

interp/api.go
```diff
@@ -366,28 +372,37 @@ func StdIO(in io.Reader, out, err io.Writer) RunnerOption {
 	}
 }
 
-func (r *Runner) optByName(name string, bash bool) *bool {
+func (r *Runner) optByName(name string, bash bool) (*bool, *shellOpt, *bashOpt) {
```
mvdan, line 375: is the second result parameter ever used?

perhaps as a way to clarify what they mean: `(posix *bool, bash *bashOpt)`

Or, for consistency, using the struct for both: `(*shellOpt, *bashOpt)`

mvdan (approved):

interp/api.go
riacataquian: oops. SORRY!

PR #890, opened 2022-06-28
expand: fix parameters with leading/trailing spaces

### LINKED ISSUE #886: Whitespace incorrectly trimmed from variables?
It feels like gosh trims leading and trailing spaces from variables:
```
$ cat x.sh
#!/bin/sh
FOO=" Q "
echo a${FOO}b
$ sh x.sh
a Q b
$ go run main.go x.sh
aQb
```

I would have expected gosh x.sh and sh x.sh to produce identical output.

mvdan:
(see commit message)

Fixes #886.


mvdan:
cc @dkegel-fastly

riacataquian (approved):

PR #891, opened 2022-06-28
syntax: add error message for herestrings in POSIX mode

### LINKED ISSUE #881: syntax: give a helpful error message when <<< is used in POSIX mode
`shfmt` is having issues with the herestring operator (`<<<`).

The minimum reproducible example:
```bash
#!/bin/bash
# script.sh
grep 'foo' <<<'foo'
```

If I run:
```
shfmt script.sh
```
I get:
```
> script.sh:3:12: << must be followed by a word
```

hristiy4n:
Fixes #881

### DIFF
--- a/syntax/parser.go
+++ b/syntax/parser.go
@@ -1644,6 +1644,11 @@ func (p *Parser) doRedirect(s *Stmt) {
 			}
 			p.doHeredocs()
 		}
+	case WordHdoc:
+		if p.lang == LangPOSIX {
+			p.langErr(r.OpPos, "herestrings", LangBash, LangMirBSDKorn)
+		}
+		fallthrough
 	default:
 		r.Word = p.followWordTok(token(r.Op), r.OpPos)
 	}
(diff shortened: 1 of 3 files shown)


mvdan (commented):
Please add a test case for `foo << < bar` as well - I think your logic will incorrectly fire in that case too, as you just look at `p.tok`.

A better solution might be to _always_ tokenize `<<<` as a single token, and then error if the parser encounters the token in posix mode.

mvdan (commented):
SGTM with one last suggestion.

And yes, it's a bit subtle whether or not to tokenize non-POSIX stuff in POSIX. In this case we want to, because in POSIX mode it should just error. In some other cases, the code still parses correctly, and then behaves differently.

syntax/parser.go
```diff
@@ -1644,6 +1644,11 @@ func (p *Parser) doRedirect(s *Stmt) {
 			}
 			p.doHeredocs()
 		}
+	case WordHdoc:
+		if p.lang == LangPOSIX {
+			p.langErr(p.pos, "herestrings", LangBash, LangMirBSDKorn)
```
mvdan, line 1649:
```suggestion
			p.langErr(r.OpPos, "herestrings", LangBash, LangMirBSDKorn)
```

like in my suggestion before :)

mvdan (approved):

mvdan (approved):

PR #893, opened 2022-06-29
expand: improve support for array keys and values

### LINKED ISSUE #884: Bug for associative arrays when using gosh with -c vs without
I found this because [task](https://github.com/go-task/task) uses this and that's where it cropped up. So I came here and started trying to debug it and found gosh to use as a potential quick way to investigate.

It seems that when run without -c it iterates over each statement and executes them separately but when using -c it treats the whole file as a script. I guess there's something about context there that's being lost?

Anyway, consider this script:

```
#!/bin/bash

declare -A database_lookup=()
database_lookup[management]=management
database_lookup[devicetwin]=devicetwin
database_lookup[identity]=identity

for d in "${!database_lookup[@]}"
do
    echo "$d -- ${database_lookup[$d]}"
done
```

The output will be quite different depending on whether you use -c with gosh or not. This exactly models the behavior I am seeing in task.

Is this a bug or am I (and the task project by extension) doing it wrong? Thanks!

mvdan:
(see commit message)

Fixes #884.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -661,31 +661,57 @@ func (cfg *Config) wordFields(wps []syntax.WordPart) ([][]fieldPart, error) {
 }
 
 // quotedElemFields returns the list of elements resulting from a quoted
-// parameter expansion if it was in the form of ${*}, ${@}, ${foo[*], ${foo[@]},
-// or ${!foo@}.
+// parameter expansion that should be treated especially, like "${foo[@]}".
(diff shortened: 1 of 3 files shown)


mvdan:
cc @everactivetim @ghostsquad

everactivetim:
Thanks @mvdan! I've confirmed the branch outputs the expected results locally.

riacataquian (approved):

andreynering:
Thanks @mvdan!

PR #894, opened 2022-07-03
interp_test: fix sleep syntax compatibility

riacataquian:
Running the tests on master gives me the following error:
```
--- FAIL: TestRunnerRunConfirm (0.03s)
    --- FAIL: TestRunnerRunConfirm/#420 (0.02s)
        interp_test.go:3690: wrong bash output in "f() { echo 1; }; { sleep 0.01s; f; } & f() { echo 2; }; wait":
            want: "1\n"
            got:  "usage: sleep seconds\n1\n"
```

I am running on `bash version 5.1.16(1)-release`:
```
$ bash --version
GNU bash, version 5.1.16(1)-release (aarch64-apple-darwin21.5.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

The error I suppose, is because of the difference of the sleep
syntax on BSD/Unix/MacOS vs on GNU/Linux - the former requires
no unit and defaults to seconds, while the latter requires either
`smdh` as unit

To fix, fix the `sleep` test builtin to accept either of the syntax
and default to seconds when none is found

Finally, modify the tests to run without units so it would pass when
ran against real bash

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -1340,7 +1340,7 @@ var runTests = []runTest{
 	},
 	{`mkdir d; old=$PWD; cd d & wait; [[ $old == "$PWD" ]]`, ""},
 	{
-		"f() { echo 1; }; { sleep 0.01s; f; } & f() { echo 2; }; wait",
+		"f() { echo 1; }; { sleep 0.01; f; } & f() { echo 2; }; wait",
 		"1\n",
 	},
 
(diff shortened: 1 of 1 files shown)


mvdan:
I guess this was always wrong but we never noticed because CI runs MacOS with an older Bash by default, and none of us developed locally with a recent enough MacOS... until now :)

mvdan (approved):
slightly amusing that you added suffixes for hours and days - hopefully we'll never, ever, use those in tests :P

mvdan:
Happy for this to be merged btw - not sure if you added the WIP state on purpose.

mvdan:
An alternative is to only support seconds as units. That's all we really need.

riacataquian:
I wanted to keep the intention 😅 removed and fixed, thanks! I was away so I kept this as WIP, will merge now.

PR #899, opened 2022-07-06
cmd/shfmt: incorrect flag name for --case-indent

### LINKED ISSUE #898: cmd/shfmt: incorrect flag name for --case-indent
`-ci, --switch-case-indent Switch cases will be indented.`

but actually using `--switch-case-indent` doesn't work, rather it is supposed to be `--case-indent` apparently

tmillr:
The shfmt manpage had a typo for the long-option `--case-indent`. It has
been corrected from `--switch-case-indent`, which was not a valid option.

Fixes #898.


mvdan (approved):
Thanks!

PR #900, opened 2022-07-13
cmd/shfmt: implement --from-json and improve -tojson

### LINKED ISSUE #35: cmd/shfmt: add flags to read/write AST in JSON format to simplify integration with other languages
Quite a few projects out there implement their own shell/bash parsers. Not only the shells themselves, but also some programs that work with shell scripts such as linters.

It's really a little silly for each one of them, especially tooling around shell scripts, to have to implement their own. Sadly, the shells out there don't allow their use as libraries (that I know of).

This is a library, so it does solve that problem. But being written in Go, it's not easily usable from other languages like Python or Shell itself.

Some ideas come to mind:
- Building the packages as shared libraries, linkable by C: http://blog.ralch.com/tutorial/golang-sharing-libraries/
- Adding an executable alongside `shfmt` that parses an input file and outputs the AST in json or some other easy-to-import format

Both of those could be extended to also include the printer.

Any input will be appreciated, especially from people who would use something like this.

mvdan:
(see commit message)

Fixes #35 again, as we never implemented the "read JSON" side.

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -69,7 +69,8 @@ var (
 	keepPadding = &boolFlag{"kp", "keep-padding", false}
 	funcNext    = &boolFlag{"fn", "func-next-line", false}
 
-	toJSON = &boolFlag{"tojson", "", false} // TODO(v4): consider "to-json" for consistency
+	toJSON   = &boolFlag{"tojson", "to-json", false} // TODO(v4): remove "tojson" for consistency
+	fromJSON = &boolFlag{"", "from-json", false}
 
 	// useEditorConfig will be false if any parser or printer flags were used.
 	useEditorConfig = true
@@ -89,7 +90,7 @@ var (
 	allFlags = []interface{}{
 		versionFlag, list, write, simplify, minify, find, diff,
 		lang, posix, filename,
-		indent, binNext, caseIndent, spaceRedirs, keepPadding, funcNext, toJSON,
+		indent, binNext, caseIndent, spaceRedirs, keepPadding, funcNext, toJSON, fromJSON,
 	}
 )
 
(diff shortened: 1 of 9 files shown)


mvdan:
cc @cilki @JounQin

mvdan:
A large portion of the diff is JSON output in test files, which was automatically generated via `go test -u`. It shouldn't be closely reviewed, but I did skim it for any mistakes or unexpected changes.

cmd/shfmt/main.go
```diff
@@ -168,7 +169,8 @@ Printer options:
 Utilities:
 
   -f, --find   recursively find all shell files and print the paths
-  --tojson     print syntax tree to stdout as a typed JSON
+  --to-json    print syntax tree to stdout as a typed JSON
```
riacataquian, line 172: does it make sense to mention here the flag change, eg `available since v3`? perhaps to make the change obvious.

riacataquian (approved):
neat :) LGTM!

JounQin:
So it won't a core API but left in `shfmt` right? I'd like to use it without `shfmt` for small bundle, should I copy/paste related codes instead?

cmd/shfmt/main.go
mvdan: We generally record that info via the changelog. I'm not sure whether including that in the help text or man page helps. If you run `shfmt -h` and the flag is there, then you can use it. If the flag isn't there, then your version is too old, so you can't use it.

JounQin:
@mvdan Thanks for clarifying, I just want to make sure it will be in core API in the future.

mvdan:
Did one last push as I forgot to update the man page; see https://github.com/mvdan/sh/compare/46e0c4e09353c070bc3e23b1aac4a3b800c64d4b..e5d80446cfdede8b096dd7c9553edf1e73ce5ab7. I'll be merging on green. Thanks for the review!

PR #902, opened 2022-07-19
Use docker token

fauust:
See: https://github.com/peter-evans/dockerhub-description/issues/10

Not mandatory but good to implement IMO:
1. create a Docker Hub token (needs read/write/delete rights)
2. replace DOCKER_PASSWORD with DOCKER_TOKEN in GH actions secrets.
3. eventually add 2FA authentication on Docker Hub (it was not possible because of that).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,9 +55,7 @@ jobs:
       # Export environment variables for all stages.
       DOCKER_USER: ${{ secrets.DOCKER_USER }}
       DOCKER_DEPLOY_IMAGES: false
-      # Pushing READMEs to Dockerhub currently only works with username/password
-      # and not with personal access tokens (Step: Update DockerHub description)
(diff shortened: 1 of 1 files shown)


mvdan:
Thanks for this! I actually tried to make DOCKER_PASSWORD into a token in 2020, but then the "update description" action started failing, so we reverted to my actual password: https://github.com/mvdan/sh/pull/492#issuecomment-580384256

More than happy to give this another try. In the process, I'll change the current password and create an entirely new token. I also think that using a different env var name will indeed help to clarify that it's a token and not a password.

As with all these things, we'll do the final test by merging and seeing what happens :)

mvdan:
Passwords and tokens reset, 2FA enabled, new token created, old password secret deleted, new token secret created. Pulling the trigger; we'll see if anything breaks.

mvdan:
Looks like it all went well: https://github.com/mvdan/sh/runs/7411624808?check_suite_focus=true

Thanks again for your help!

PR #912, opened 2022-09-07
WIP: add $HOSTTYPE and $OSTYPE

zimbatm:
Bash has those implicit variables set. Do you think it's worth having
them in here as well? I don't have the full mapping but happy to keep
digging.


mvdan:
This seems perfectly fine to me :) The interpreter generally supports Bash features in a best-effort way, simply because it would be a full-time job for many months to fully implement `man bash`. So I certainly welcome PRs with tests :)

lawrence-c:
Hey! I was wondering if there was an update to this PR? Just hit a similar issue when using Taskfile, it seems to link back to here.

Followed from: https://github.com/go-task/task/issues/396 cc @andreynering

zimbatm:
Feel free to take over; my plate is currently quite full.

Issue #923, opened 2022-10-01
Minifying heredoc values in command substitutions produces invalid shell


andreykaipov:
For example, running the following file through `shfmt -mn`:

```
#!/bin/sh
a=$(
        cat <<EOF
	hello
EOF
)
echo "$a"
```

Produces the following shell:

```
#!/bin/sh
a=$(cat \
<<EOF)
	hello
EOF
echo "$a"

```

Which can't be ran:

```
sh: line 4: warning: here-document at line 4 delimited by end-of-file (wanted `EOF')
sh: line 3: warning: here-document at line 3 delimited by end-of-file (wanted `EOF')
sh: line 4: hello: command not found
sh: line 5: EOF: command not found
```

For some context, I'm using this kind of pattern to assign literal strings to a variable. I can use something like `read -r -d '' a <<'EOF'` in bash, but afaik assigning `cat <<'EOF'` is the way to do something like that with posix sh.

andreykaipov:
So this behavior is actually not just with command substitutions, but also exists for heredocs in subshells and in process substitutions too.

andreykaipov:
Digging through the code I've at least figured a workaround:

```
#!/bin/sh
a=$(
        cat <<EOF
	hello
EOF
        :
)
echo "$a"
```

Adding an extra statement after the ending heredoc word allows the formatter to correctly flush the heredoc, I'm guessing because of the new line. Now running `shfmt -mn` produces:

```
#!/bin/sh
a=$(cat \
<<EOF
        hello
EOF
:)
echo "$a"
```

PR #924, opened 2022-10-02
wip: syntax: flush heredocs before right parens

andreykaipov:
Fixes https://github.com/mvdan/sh/issues/923 😄

### DIFF
--- a/syntax/printer_test.go
+++ b/syntax/printer_test.go
@@ -1059,6 +1059,18 @@ func TestPrintMinify(t *testing.T) {
 		},
 		samePrint("foo >bar 2>baz <etc"),
 		samePrint("<<-EOF\n$(a|b)\nEOF"),
+		{
+			"a=$(\n\tcat <<'EOF'\n  hello\nEOF\n)",
+			"a=$(cat \\\n<<'EOF'\n  hello\nEOF\n)",
+		},
+		{
+			"(\n\tcat <<EOF\n hello\nEOF\n)",
+			"(cat \\\n<<EOF\n hello\nEOF\n)",
+		},
+		{
+			"diff -y <(cat <<EOF\n1\n2\n3\nEOF\n) <(cat <<EOF\n1\n4\n3\nEOF\n)",
+			"diff -y <(cat \\\n<<EOF\n1\n2\n3\nEOF\n) <(\ncat <<EOF\n1\n4\n3\nEOF\n)",
+		},
 	}
 	parser := NewParser(KeepComments(true))
 	printer := NewPrinter(Minify(true))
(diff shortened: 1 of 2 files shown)


syntax/printer_test.go
```diff
@@ -1059,6 +1059,18 @@ func TestPrintMinify(t *testing.T) {
...
+			"a=$(\n\tcat <<'EOF'\n  hello\nEOF\n)",
+			"a=$(cat \\\n<<'EOF'\n  hello\nEOF\n)",
+		},
+		{
+			"(\n\tcat <<EOF\n hello\nEOF\n)",
+			"(cat \\\n<<EOF\n hello\nEOF\n)",
+		},
+		{
+			"diff -y <(cat <<EOF\n1\n2\n3\nEOF\n) <(cat <<EOF\n1\n4\n3\nEOF\n)",
+			"diff -y <(cat \\\n<<EOF\n1\n2\n3\nEOF\n) <(\ncat <<EOF\n1\n4\n3\nEOF\n)",
```
andreykaipov, line 1072: Something I realized while writing these examples is minified heredocs might not necessarily be minified! For example the following is 21 bytes:
```
a=$(cat <<EOF
EOF
)
```

And produces the following 22 bytes:
```
a=$(cat \
<<EOF
EOF
)
```

It's because of the `cat \`, but digging through the code I couldn't immediately see what causing that, or how to address it without breaking too much else. In the meantime, I figure valid shell is a lot better than invalid shell though!

andreykaipov: The `diff` example in the test is very funky, because the minified code has the first `cat` continuing its `<<EOF` onto the next line, but the next `cat` in the second proc subst is left alone as `cat <<EOF`.

andreykaipov:
Oh no, sorry this is embarrassing... I foolishly was only running `go test ./... -run=TestPrintMinify` during my testing...

This change as it is breaks a few other wacky cases, like:

```
foo <<EOF | $(bar)
3
EOF
```

I did some more digging through the code to see if I can satisfy both (the existing test case of a subshell after the beginning heredoc word, and the new one I added where the heredoc is within a subshell), but I think it would involve a new printer field to keep track of if we're in a heredoc or not.

mvdan:
Finally figured it out, it was a bit subtle :) I also noticed that the escaped newlines right before `<<` weren't necessary with the minify mode, that's fixed as well. Fix coming shortly, and I'll credit you for the test cases. Thank you!

andreykaipov:
Oh - thanks! Sorry this got lost in the inbox. Time sure does fly. Thanks so much! Really appreciate you and this project! 🥳

Issue #926, opened 2022-10-07
set -x breaks piping


graf0:
Hello!

When you enable tracer via set -x, it breaks pipes.

### how to reproduce?

install jq, then:

```
#!/bin/gosh
set -x
echo '{"number":1}' | jq '.number'
```

efect of this script should be:

```
1
```

is:

```
+ jq .number
parse error: Invalid numeric literal at line 1, column 2
```

it could be even simpler without jq:

```
#!/bin/gosh
set -x
echo "test" | tee test.txt
```

in test.txt there is following content:

```
+ echo test
test
```

### where is problem?

in file interp/trace.go, line 28 I think we should set stderr as output stream. Otherwise printed traces become input of next command in the pipe.

mvdan:
Nicely spotted - I think you're right:

```
$ bash -x -c 'echo foo' >/dev/null
+ echo foo
$ bash -x -c 'echo foo' 2>/dev/null
foo
```

Want to send a PR? cc @riacataquian

graf0:
> Want to send a PR? cc @riacataquian

Sure - you have it :)

PR #927, opened 2022-10-08
resolves trace issue

graf0:
when enabled trace should send its output to stderr
fixes  mvdan/sh#926

### DIFF
--- a/interp/trace.go
+++ b/interp/trace.go
@@ -14,7 +14,7 @@ import (
 type tracer struct {
 	buf       bytes.Buffer
 	printer   *syntax.Printer
-	stdout    io.Writer
+	stderr    io.Writer
 	needsPlus bool
 }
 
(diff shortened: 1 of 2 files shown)


interp/trace.go
```diff
@@ -14,7 +14,7 @@ import (
 type tracer struct {
 	buf       bytes.Buffer
 	printer   *syntax.Printer
-	stdout    io.Writer
+	stderr    io.Writer
```
mvdan, line 17: nit: might as well call this `output`, because it has just one output writer, and in the future we'll likely need to support outputs other than stderr for better bash compatibility

mvdan (commented):
Please add a test as well - something like what you showed in your original report would suffice. See how previous bug fixes in the interpreter have added integration tests.

interp/trace.go
graf0: done.

mvdan (approved):
Thanks!

mvdan:
The tests aren't passing :) Did you run them?

Also cc @riacataquian for a review

riacataquian (approved):
nice! thanks!

PR #935, opened 2022-10-23
expand,interp: trim whitespace in atoi

### LINKED ISSUE #928: interp: some binary tests should use space trim before atoi
It seems that the Bash trims spaces before integer comparison:
```
> if [ " 14 " -lt "22" ]; then  echo "true"; fi
< true

> if [ " 14 " -eq "14" ]; then  echo "true"; fi
< true
```

The sh/interp uses strict comparision rules and does not use any space trim functions before the binary tests that are defined in `interp/test.go` (`binTest` func, `atoi`).

mvdan:
(see commit message)

Fixes #928.

### DIFF
--- a/expand/arith.go
+++ b/expand/arith.go
@@ -6,6 +6,7 @@ package expand
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"mvdan.cc/sh/v3/syntax"
 )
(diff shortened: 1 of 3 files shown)


riacataquian (approved):

theclapp (commented):
A little late, but LGTM.

PR #947, opened 2022-11-22
syntax: fix a parser crash when trying to produce a friendly error

mvdan:
(see commit message)


mvdan:
Going to merge for now; always happy to get a review later :)

PR #949, opened 2022-11-24
add shfmt vscode extension

mkhl:
i wrote an extension for vscode that *only* formats shell script files and nothing else


mvdan (approved):
Thanks!

PR #952, opened 2022-12-03
syntax: fix two bugs when formatting case clauses

### LINKED ISSUE #779: syntax: case clause formatting can add empty line when removing a newline
Before:

```sh
case "$x"
in
(0)
        echo 'none!'
        ;;
(1)
        echo 'it'"'"'s one!'
        ;;
(*)
        echo 'something!'
        ;;
esac
```

After `shfmt -p -s`:

```sh
#!/bin/sh

set -e -f -u -x

x="${1:-0}"
readonly x

case "$x" in

0)
        echo 'none!'
        ;;
1)
        echo 'it'"'"'s one!'
        ;;
*)
        echo 'something!'
        ;;
esac
```

POSIX allows `case` items to start with a parenthesis, and I prefer it, because it makes `case` look more understandable, imo.  I would like an option to either add them or at least not remove them.

(Also, I've noticed that `shfmt` seems to add an empty line after `case … in`, which looks like a bug.)

### LINKED ISSUE #917: syntax: comments before `case` are incorrectly indented
```
$ shfmt --version
v3.3.1
```

Notice `# baz` is indented 2 extra spaces, when it shouldn't be.

### Actual

```sh
case foo in
  # bar
  "$bar") ;;
    # baz
  "$baz") ;;
esac
```

### Expected

```sh
case foo in
  # bar
  "$bar") ;;
  # baz
  "$baz") ;;
esac
```

mvdan:
(see commit messages - please do not squash)

Fixes #779.
Fixes #917.

cc @ainar-g @Julien-Elie @zx8 @DannyBen if you want to test this out

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This release drops support for Go 1.17 and includes many features and fixes.
   - Simplify `${name:-}` to the equivalent `${name-}` - [#849]
   - Don't print trailing whitespaces on nested subshells - [#814]
   - Don't print extra newlines in some case clauses - [#779]
+  - Don't indent comments preceding case clause items - [#917]
   - Allow escaped newlines before unquoted words again - [#873]
   - Parse a redirections edge case without spaces - [#879]
   - Give a helpful error when `<<<` is used in POSIX mode - [#881]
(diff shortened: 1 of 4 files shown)


DannyBen:
Want. But I am only using the binary. I will try to build me a small Docker with it for testing.

DannyBen:
Beautiful.

I can confirm this branch works as expected for me. Tested with the `test.sh` file mentioned in #950

```diff
➜ shfmt -d -i 2 -ci test.sh

--- test.sh.orig
+++ test.sh
@@ -3,8 +3,7 @@
 action=download

 case $action in
-  -*)
-    ;;
+  -*) ;;

   upload)
     # This upload section is being messed up with a weird backslash
```

ainar-g:
I can confirm that the issue with `case`'s is fixed.  (I seem to have found another issue, but that one seems to be unrelated to this PR, and I'll file it separately.)  Thanks!

PR #954, opened 2022-12-05
interp: make trap consistent with errexit logic

reubeno:
interp already excludes conditional contexts for exiting on error when `errexit` is set (see case above); this makes the logic consistently apply the same filtering for reporting errors to a trap routine.

To test this behavior, I ran the below script under `bash`, under `gosh` (without this change), and under `gosh` (with this change). `bash` reports that the handler was invoked once. Without this change `gosh` reports that it was invoked twice, and with this change, only once.

```bash
#!/bin/bash

count=0

my_handler()
{
  echo "Handler invoked"
  count=$((count + 1))
}

trap my_handler ERR

echo "Trying unconditional"
non_existent_cmd

echo "Trying conditional"
non_existent_cmd || true

echo "Handler called: $count time(s)"
```

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -2156,6 +2156,7 @@ set +o pipefail
 	// trap
 	{"trap 'echo at_exit' EXIT; true", "at_exit\n"},
 	{"trap 'echo on_err' ERR; false; echo FAIL", "on_err\nFAIL\n"},
+	{"trap 'echo on_err' ERR; false || true; echo OK", "OK\n"},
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks for the patch! I guess you're using the interpreter for a project, would be curious to hear what it is :)

The change sounds good to me - it just needs a regression test. See recent changes in the interpreter for some examples, or see the existing `trap` tests. The script you shared would work, but it's quite lengthy. The test cases tend to be much shorter for the sake of being able to maintain lots of tests more easily.

mvdan (approved):
Thanks - the test looks just right!

PR #955, opened 2022-12-07
interp: implement %b for printf builtin

reubeno:
Update `printf` implementation to do the same as `echo -e` for the `%b` format specifier: expand backslash escape sequences in the provided argument but not `%` format specifiers.

Extracts inner logic from `Format` into `formatIntoBuffer` to avoid issues with non-reentrancy of `Format` (due to use of a special cached buffer in `Config`.

---

_Request for feedback_

This was the simplest change that I could see, but I wasn't thrilled about having `formatIntoBuffer` recursively call itself for the `%b` case; this shouldn't generally be problematic, since the recursive call won't process `%` format specifiers and shouldn't further recur.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -215,6 +215,24 @@ func Pattern(cfg *Config, word *syntax.Word) (string, error) {
 func Format(cfg *Config, format string, args []string) (string, int, error) {
 	cfg = prepareConfig(cfg)
 	buf := cfg.strBuilder()
+
+	consumed, err := formatIntoBuffer(buf, format, args)
+	if err != nil {
+		return "", 0, err
+	}
+
+	return buf.String(), consumed, err
+}
+
+// Format expands a format string with a number of arguments, following the
+// shell's format specifications. These include printf(1), among others.
+//
+// The resulting string is written to the provided buffer, and the number
+// of arguments used is returned.
+//
+// The config specifies shell expansion options; nil behaves the same as an
+// empty config.
+func formatIntoBuffer(buf *bytes.Buffer, format string, args []string) (int, error) {
 	var fmts []byte
 	initialArgs := len(args)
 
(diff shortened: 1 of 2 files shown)


expand/expand.go
```diff
@@ -215,6 +215,24 @@ func Pattern(cfg *Config, word *syntax.Word) (string, error) {
...
+
+// Format expands a format string with a number of arguments, following the
+// shell's format specifications. These include printf(1), among others.
+//
+// The resulting string is written to the provided buffer, and the number
+// of arguments used is returned.
+//
+// The config specifies shell expansion options; nil behaves the same as an
+// empty config.
+func FormatIntoBuffer(buf *bytes.Buffer, format string, args []string) (int, error) {
```
mvdan, line 235: do we need to expose this function right away? I imagine users will still be calling Format - if for any reason that API isn't performant enough, we can discuss a lower-level API as a separate topic. Exporting doesn't seem required for supporting %b.

interp/interp_test.go
```diff
@@ -304,6 +304,9 @@ var runTests = []runTest{
 	{"printf 'nofmt' 1 2 3", "nofmt"},
 	{"printf '%d_' 1 2 3", "1_2_3_"},
 	{"printf '%02d %02d\n' 1 2 3", "01 02\n03 00\n"},
+	{`printf '0%s1' 'a\bc'`, `0a\bc1`},
+	{`printf '0%b1' 'a\bc'`, "0a\bc1"},
```
mvdan, line 308: would be good to have an example where %b is different from %s, I think :)

mvdan (commented):
Thanks! Two thoughts below.

I'm not too worried about the recursion. We can always optimize or refactor later if needed.

expand/expand.go
reubeno: That's a good point. I've updated the PR to not export this function.

mvdan (approved):

PR #958, opened 2022-12-17
interp: set EUID

anderbubble:
Set EUID variable analog to UID.

Signed-off-by: Jonathon Anderson <janderson@ciq.co>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [3.6.1] - 2023-??-??
+
+- **interp**
+  - Set `EUID` just like `UID`
+
 ## [3.6.0] - 2022-12-11
 
 This release drops support for Go 1.17 and includes many features and fixes.
--- a/interp/api.go
+++ b/interp/api.go
@@ -613,6 +613,13 @@ func (r *Runner) Reset() {
 			Str:      strconv.Itoa(os.Getuid()),
 		})
 	}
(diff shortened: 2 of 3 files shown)


anderbubble:
This is meant to be a better fix for https://github.com/apptainer/apptainer/issues/837 than is possible downstream. (e.g., https://github.com/apptainer/apptainer/pull/937)

mvdan (changes requested):
did you mean to use https://pkg.go.dev/os#Geteuid?

anderbubble:
Absolutely. I was just too far down my downstream thought process, where we had been planning to set EUID=$UID in a script. Thanks for pointing this out.

PR updated.

mvdan (approved):
SGTM, thanks. Add a couple of tests mirroring the existing ones for UID? I think one `test` and one `unset` test should be enough.

anderbubble:
@mvdan how about these?

mvdan (approved):

anderbubble:
Thanks for the merge!

Do you have an estimate for when this might be included in a release?

mvdan:
I'll probably do a bugfix release in the coming weeks. I just did a minor (feature) release, so that likely won't happen again for a few months. If you need this sooner, you should be able to use master.

DrDaveD:
So, would you include this in a bugfix release?

mvdan:
Sure, why not. It doesn't feel too risky. Though, again, you should be able to use master easily.

Issue #93, opened 2017-04-25
interp: add more coreutils as builtins


mvdan:
coreutils as 8.27 has a total of 103 executables, ranging from `wc` to `test` and `sort`.

Plenty of scripts depend on these. If we really want `interp` to be platform-independent, we will have to implement some of these as builtins as they won't be available in some systems like Windows (and Mac?).

Bash 4.4 already has a few as builtins:

```
 $ for c in $(pacman -Ql coreutils | grep '\/bin' | sed 's@.*/@@'); do type -a $c; done | grep builtin
[ is a shell builtin
echo is a shell builtin
false is a shell builtin
printf is a shell builtin
pwd is a shell builtin
test is a shell builtin
true is a shell builtin
```

We already have all of these except for `[` and `test` - see #92.

I obviously won't implement all 103 in one go, but I could start with the most common and simple. Unfortunately, these being GNU programs they all have tons of options and gotchas, like `cat` having a dozen options.

One open question is whether these should always be builtins like `echo`. Other options include:

* Use them as fallback only if the system version is not available
* Enable them via a flag, like "extended builtins"

For the sake of simplicity, I'm inclined towards them always being builtins. The only downside is scripts that depend on GNU flags that are somewhat obscure. But these scripts wouldn't be portable to begin with.

mvdan:
A good way to make the list of 103 executable shorter is to filter it through the list of POSIX Shell utilities, which has 160 executables: http://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html

Those docs are also a good start, as each command is simpler and has less options - `cat` only has a single option.

The number of executables that are both in `coreutils` and in the POSIX utilities list are 60:

```
basename
cat
chgrp
chmod
chown
cksum
comm
cp
csplit
cut
date
dd
df
dirname
du
echo
env
expand
expr
false
fold
head
id
join
link
ln
logname
ls
mkdir
mkfifo
mv
nice
nl
nohup
od
paste
pathchk
pr
printf
pwd
rm
rmdir
sleep
sort
split
stty
tail
tee
test
touch
tr
true
tsort
tty
uname
unexpand
uniq
unlink
wc
who
```

Some of these like `true` and `echo` we already have. And most of the ones we don't yet have are common and seem fairly simple to implement.

mvdan:
@andreynering this might be of interest to you if you want unix scripts to work on Windows

andreynering:
[...]

PR #964, opened 2023-01-15
interp: add ExecHandlers to support exec middlewares

mvdan:
(see commit message)

For #93.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -71,10 +71,15 @@ type Runner struct {
 	// arguments. It may be nil.
 	callHandler CallHandlerFunc
 
-	// execHandler is a function responsible for executing programs. It must be non-nil.
+	// execHandler is responsible for executing programs. It must not be nil.
 	execHandler ExecHandlerFunc
 
-	// openHandler is a function responsible for opening files. It must be non-nil.
+	// execMiddlewares grows with calls to ExecHandlers,
+	// and is used to construct execHandler when Reset is first called.
+	// The slice is needed to preserve the relative order of middlewares.
+	execMiddlewares []func(ExecHandlerFunc) ExecHandlerFunc
+
+	// openHandler is a function responsible for opening files. It must not be nil.
 	openHandler OpenHandlerFunc
 
 	// readDirHandler is a function responsible for reading directories during
@@ -181,7 +186,6 @@ func (r *Runner) optByFlag(flag byte) *bool {
 func New(opts ...RunnerOption) (*Runner, error) {
 	r := &Runner{
 		usedNew:        true,
-		execHandler:    DefaultExecHandler(2 * time.Second),
 		openHandler:    DefaultOpenHandler(),
 		readDirHandler: DefaultReadDirHandler(),
 		statHandler:    DefaultStatHandler(),
@@ -213,9 +217,11 @@ func New(opts ...RunnerOption) (*Runner, error) {
 	return r, nil
 }
 
-// RunnerOption is a function which can be passed to New to alter Runner behaviour.
-// To apply option to existing Runner call it directly,
-// for example interp.Params("-e")(runner).
+// RunnerOption can be passed to New to alter a Runner's behaviour.
+// It can also be applied directly on an existing Runner,
+// such as interp.Params("-e")(runner).
+// Note that options cannot be applied once Run or Reset have been called.
+// TODO: enforce that rule via didReset.
 type RunnerOption func(*Runner) error
 
 // Env sets the interpreter's environment. If nil, a copy of the current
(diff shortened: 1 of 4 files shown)


interp/api.go
```diff
@@ -329,14 +332,37 @@ func CallHandler(f CallHandlerFunc) RunnerOption {
...
 func ExecHandler(f ExecHandlerFunc) RunnerOption {
 	return func(r *Runner) error {
 		r.execHandler = f
 		return nil
 	}
 }
 
+// ExecHandlers appends command execution handler middlewares.
+// Each middleware is run in order from first to last,
+// and chooses whether to fall back to the next middleware.
```
theclapp, line 348: This sentence is contradictory to me.  Are the middlewares run in order first to last, or does each one choose whether to fall back to the next one?

The api seems awkward, too, since each middleware has to correctly call the `next` function.  It seems like it'd be better if the interpreter looped over the middlewares itself, and each one should return either a `result, continue` pair (either one or the other but not both), or better yet a single result with a special `interp.CallNextExecHandler` (or something) value to signal the scaffolding to call the next handler.

theclapp (changes requested):

interp/api.go
mvdan: You're right that the godoc is a bit awkward. How about:

> The middlewares are chained from first to last, and the first is called by the runner.
> The middleware can choose whether to handle the operation and return an error,
> or to not handle the operation and fall back to calling the `next` middleware.

About the awkward API - I actually implemented precisely what you mean in the original coreutils branch. See: https://github.com/mvdan/sh/compare/93-coreutils

It is conceptually simpler, as there are no `func(next Handler) Handler` shenanigans which may confuse the user. However, it is also less powerful. When a middleware can only handle an operation or return "skip me", those are its only two options. With the "next" call approach, you can imagine middlewares which _alter the operation parameters_ while letting the next middleware handle the operation.

For example, imagine an exec middleware which wants to replace all calls to `chrome` with `firefox`. With the "skip me" approach, it would have to detect `args[0] == "chrome"` and then fully implement the process execution, e.g. by calling `DefaultExecHandler` with the altered `args[0]`. If there was a middleware right after which would also have altered how this process execution happened, by e.g. using a higher exec timeout, that would be ignored.

With the "next" approach, the logic can just be as follows, and the middlewares coming later in the chain are obeyed as usual:

```
if args[0] == "chrome" {
    args[0] = "firefox"
}
next(args)
```

This is heavily used in go-chi to alter requests or responses without actually _finishing_ them. For example, see https://github.com/go-chi/chi/blob/v5.0.8/middleware/clean_path.go#L12, which cleans the path of an HTTP request and lets the next middleware do its thing. I think that's very powerful.

theclapp: The godoc you suggested is better.  It leaves out the nuance that a) you could filter/replace/log the args when calling next, and b) you could filter/log the return value when next returns.  If you feel like that's implicitly obvious, no problem.  If somebody's familiar with this kind of API (obviously I wasn't), then it probably would be obvious, but if they're not, it might not be.

You're right, this api is more powerful than just "handle or skip".  I was going to suggest that you go with my suggested api and just pass a pointer to args, which would let you change them, but then that wouldn't let you similarly filter (or even just log) the return value.

mvdan: That's a good point about other options, I'll expand it a bit further.

One disadvantage of this more powerful API is that being able to call "next" more than once, or with entirely wrong parameters, is a potential footgun. But hopefully that kind of mistake will be pretty easy to spot.

riacataquian (approved):

interp/api.go
```diff
@@ -71,10 +71,12 @@ type Runner struct {
 	// arguments. It may be nil.
 	callHandler CallHandlerFunc
 
-	// execHandler is a function responsible for executing programs. It must be non-nil.
+	// execHandler is responsible for executing programs. It must not be nil.
 	execHandler ExecHandlerFunc
 
-	// openHandler is a function responsible for opening files. It must be non-nil.
+	execMiddlewares []func(ExecHandlerFunc) ExecHandlerFunc
```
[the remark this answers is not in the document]
andreynering: You could potentially move this type to something like this, but this is only a suggestion 🙂 

```go
type ExecMiddlewareFunc func(ExecHandlerFunc) ExecHandlerFunc
```

```go
execMiddlewares []ExecMiddlewareFunc
```

interp/api.go
```diff
@@ -329,14 +332,37 @@ func CallHandler(f CallHandlerFunc) RunnerOption {
...
 		return nil
 	}
 }
 
+// ExecHandlers appends command execution handler middlewares.
+// Each middleware is run in order from first to last,
+// and chooses whether to fall back to the next middleware.
+//
+// The last exec handler is DefaultExecHandler(2 * time.Second).
+func ExecHandlers(middlewares ...func(next ExecHandlerFunc) ExecHandlerFunc) RunnerOption {
```
andreynering, line 351:
```suggestion
func ExecHandlers(middlewares ...ExecMiddlewareFunc) RunnerOption {
```

andreynering (approved):

interp/api.go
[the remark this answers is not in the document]
mvdan: I thought about a type, but note that we don't want a named type - we don't need methods and I don't want to add type conversions into the mix.

A type alias wouldn't have that problem, but it doesn't really gain us much. The API would go from

```
func ExecHandlers(middlewares ...func(next ExecHandlerFunc) ExecHandlerFunc) RunnerOption
```

to

```
type ExecMiddleware = func(ExecHandlerFunc) ExecHandlerFunc

func ExecHandlers(middlewares ...ExecMiddleware) RunnerOption
```

so in my opinion this is only worthwhile if the `func(T) T` type appears in multiple places. It might in the future, but it doesn't now, so I don't think it's worth it :)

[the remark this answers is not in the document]
mvdan: Adding godoc.

interp/api.go
mvdan: I've expanded the godoc.

interp/api.go
mvdan: See my other reply :)

interp/api.go
```diff
@@ -561,6 +587,18 @@ func (r *Runner) Reset() {
 		r.origStdin = r.stdin
 		r.origStdout = r.stdout
 		r.origStderr = r.stderr
+
+		if r.execHandler != nil && len(r.execMiddlewares) > 0 {
+			// ExecHandler should be replaced with ExecHandlers, not mixed.
+			panic("interp.ExecHandler and interp.ExecHandlers cannot be used together")
```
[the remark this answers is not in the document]
mvdan: Ha, good point. Joined.

mvdan:
Thanks all for the reviews!

PR #965, opened 2023-01-23
all: start using doc links

mvdan:
(see commit message)


mvdan:
Even filed two upstream bugs while trying out doc links... :sweat_smile:

mvdan:
You can see the result here, for example: https://pkg.go.dev/mvdan.cc/sh/v3@v3.6.1-0.20230123163603-8b4fb75c77da/interp#ExecHandlerFunc

I think the docs are much more usable with links.

riacataquian (approved):
> I think the docs are much more usable with links.

fully agreed! :)

PR #966, opened 2023-01-24
Add `pre-commit.com` support

taoufik07:
Now that `language_version` is supported for `golang` starting from [pre-commit v3.0.0](https://github.com/pre-commit/pre-commit/releases/tag/v3.0.0) and as this was the main [concern](https://github.com/mvdan/sh/pull/762#pullrequestreview-811904344), I think we can now fully support to `pre-commit`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,16 @@
+- id: shfmt
+  name: shfmt
+  description: A shell formatter.
+  minimum_pre_commit_version: 3.0.0
+  entry: shfmt
+  language: golang
+  types: [shell]
+  args: [-s, -w]
+
+- id: shfmt-system
+  name: shfmt system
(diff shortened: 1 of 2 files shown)


trallnag:
Advantage of pre-commit is that it installs Golang. So it is easier to use shfmt in a non Go project. But the workaround with a forked project works fine as well.

trallnag:
@taoufik07 the alternative is to mirror the project manually like I did here: https://github.com/trallnag/pre-commit-mirror-gofumpt

Freed-Wu:
Comes from #976. Since now, pre-commit-mirror-maker doesn't support go. See <https://github.com/pre-commit/pre-commit-mirror-maker/issues/164>.

Freed-Wu:
BTW, shfmt doesn't support zsh. So the following is needed:
```yaml
  exclude_types:
    - zsh
```

PR #971, opened 2023-02-11
Fix link to Google's Shell Style Guide

teefill:
The Google Shell Style Guide moved at some point, just updating the link.


mvdan (approved):
Thank you! I always found it a bit weird that the link had an XML extension.

PR #974, opened 2023-02-26
Revert "syntax: simplify unset-or-null with null default to just unset"

### LINKED ISSUE #970: Should always use e`${XXXX:-}`, not `${XXXX-}`
I noticed in a recent update (where "recent" means "recent to me" - the update could just as well be 2 years ago) that `${XXXX:-}` is auto-corrected to `${XXXX-}`.

Although the current change _does_ optimize for _slighty_ faster shell runtimes and _1-byte_ per instance smaller script size, this is a mistake.

It should optimize for **readability** and **change**, not file size or microticks.

If I ever go to refactor `${XXXX:-}` to be `${XXXX:-default}`, this will work wonderfully.

If I go to refactor `${XXXX-}` to `${XXXX-default}` and I'm not a shell expert, little do I know that I've just introduced a subtle bug into my program. And even if I am a shell expert, if I don't have my maximum awareness, and my screen isn't squeaky-clean from any dust specs (which is a real and practical problem I've come across on more than one occasion), I'll still introduce a bug into my script.

mvdan:
(see commit message)

Fixes #970.

### DIFF
--- a/syntax/simplify.go
+++ b/syntax/simplify.go
@@ -16,7 +16,6 @@ import "bytes"
 //	Remove redundant quotes                  [[ "$var" == str ]]
 //	Merge negations with unary operators     [[ ! -n $var ]]
 //	Use single quotes to shorten literals    "\$foo"
-//	Remove redundant param expansion colons  ${foo:-}
 func Simplify(n Node) bool {
 	s := simplifier{}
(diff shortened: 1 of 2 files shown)


riacataquian (approved):
LGTM

PR #983, opened 2023-03-18
Fix interp build tags to allow WebAssembly builds

diamondburned:
This PR changes package interp's build tags and filenames around to allow building on WebAssembly (and potentially Plan 9).


mvdan (approved):
thanks!

PR #1000, opened 2023-05-09
interp: Fix (*Runner).errf missing new lines

diamondburned:
Some of the `errf` calls were missing new lines, so I added them. One call is left where new lines are not wanted.


diamondburned:
(For future reference) I've confirmed that the test changes were needed:

screenshot of output of `break` outside loop

PR #1010, opened 2023-06-08
Adding a .pre-commit-hooks.yaml

djgoku:
@scop is there a compelling reason not to move these pre-commit hooks into this repository?

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,20 @@
+- id: shfmt
+  name: shfmt
+  description: Shell source code formatter (native install)
+  language: golang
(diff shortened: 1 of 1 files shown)


mvdan:
Thanks for your pull request. There have been quite a few of these, and I still don't understand the point of such a YAML file beyond stating "yes, you can run this with pre-commit". Anyone who wants to run `shfmt` can hopefully already do that with ease. See https://github.com/mvdan/sh/pull/966 for more details.

scop:
Some related info over at #554, too. I understand the concerns, and I'm fine with maintaining the configs separately, and also here should the stance change sometime.

PR #1016, opened 2023-07-07
chore: remove refs to deprecated io/ioutil

testwill:


interp/handler.go
```diff
@@ -318,7 +318,19 @@ type ReadDirHandlerFunc func(ctx context.Context, path string) ([]os.FileInfo, e
 // It makes use of [ioutil.ReadDir].
```
mvdan, line 318: fix this reference as well

mvdan (approved):

mvdan (approved):

mvdan:
Hm, this appears to break the tests. I guess in some edge cases the behavior is subtly different. I honestly don't think it's worth spending the time to figure out why; per the TODO, I will swap out the API in v4 to use `fs.DirEntry` directly. Thanks again for the PR.

PR #1017, opened 2023-07-12
cmd/shfmt: remove inline EditorConfig comments

slabstone:
EditorConfig 0.15.0 does not allow inline comments

https://spec.editorconfig.org/#no-inline-comments


mvdan (approved):
Interesting. I can understand why they made that change, but I admit it's a shame to not be able to use inline comments.

PR #1027, opened 2023-08-23
syntax: Allow redirects before all arguments

### LINKED ISSUE #942: syntax: allow printing redirections before all arguments
Sometimes, to improve readability, I put logfile redirections at the beginning of the line:

```sh
>>$LOGFILE echo "[label]" 1 2 3
>>$LOGFILE some_command 4 5 6
```

This is currently, as of shfmt v3.5.1, incorrectly reformatted as follows:

```sh
echo >> $LOGFILE "[label]" 1 2 3
some_command >> $LOGFILE 4 5 6
```

## Research

Putting redirects at various places in-between arguments and at the end seems to have been supported from the very beginning of `sh` in 1971. Putting redirects at the beginning was not supported in UNIX V1, but emerged sometime later. I haven't found out when.

UNIX V1 manual (1971-03-11) (and [code](https://www.tuhs.org/cgi-bin/utree.pl?file=V1/sh.s)):
> Each command is a sequence of non-blank command arguments separated by blanks.  [...] Two characters [i.e., "<" or ">"] cause the immediately following string to be interpreted as a special argument to the shell itself, not passed to the command. An argument of the form <arg causes the file arg to be used as the standard input file of the given command; an argument of the form “>arg” causes file “arg” to be used as the standard output file for the given command.

[Bash documentation, section 3.6](https://www.gnu.org/software/bash/manual/bash.html#Redirections) (2022-11-13):
> The following redirection operators may _**precede**_ or appear anywhere within a simple command or may follow a command. Redirections are processed in the order they appear, from left to right. [...]

[Bash documentation, section 3.7.1](https://www.gnu.org/software/bash/manual/bash.html#Simple-Command-Expansion) (2022-11-13):
> When a simple command is executed, the shell performs the following expansions, assignments, and redirections, from left to right, in the following order: 1. The words that the parser has marked as variable assignments (those preceding the command name) and **redirections** are saved for later processing. [...]

MaienM:
Fixes #942.

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1058,36 +1058,44 @@ func (p *Printer) stmt(s *Stmt) {
 	p.decLevel()
 }
 
+func (p *Printer) printRedirsUntil(redirs []*Redirect, startRedirs int, pos Pos) int {
+	for _, r := range redirs[startRedirs:] {
+		if r.Pos().After(pos) || r.Op == Hdoc || r.Op == DashHdoc {
+			break
+		}
+		if p.wantSpace == spaceRequired {
+			p.spacePad(r.Pos())
+		}
+		if r.N != nil {
+			p.writeLit(r.N.Value)
+		}
+		p.WriteString(r.Op.String())
+		if p.spaceRedirects && (r.Op != DplIn && r.Op != DplOut) {
+			p.space()
+		} else {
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Thanks!

Issue #171, opened 2017-11-07
interp: add a single process mode


mvdan:
Some shell features require the interpreter to own the process. To give some examples:

* The `trap` builtin to handle signals in a custom way (unimplemented now)
* The `exec` builtin (we fork a sub-process via `os/exec` now instead of doing a real exec)

None of these can work if the interpreter is simply bound to a goroutine, and not to the entire process. We can't assume that the interpreter owns the process by default, as most programs wouldn't need these features anyway.

I propose that we add a mode to the interpreter that tells it that it does own the process. That is, that no other interpreters are running in other goroutines and that it's up to the developer to make sure that these `trap` and `exec` actions won't break the rest of the software running.

For example, `cmd/gosh` would use this. The sole purpose of the tool is to run one shell program at a time, so it's perfectly reasonable.

If the mode isn't enabled and any of those are used, either a regular or a fatal error could be given. (A regular one just sets the exit code, a fatal one stops the interpreter completely)

/cc @andreynering

mvdan:
Another use case: `[[ -t 0 ]]` tests if stdin is a terminal. This should only query the real file descriptor 0 if we are in "single process" mode.

PR #1034, opened 2023-09-27
interp: fix `cd` builtin: call a system function to determine access instead of reverse-engineering it ourselves

### LINKED ISSUE #1033: `cd` builtin doesn't work if your group doesn't match the directory's group
The `cd` builtin checks whether your user ID can read a directory, or whether your actual group ID can read a directory, or whether "others" can read a directory, but it doesn't check the other groups your ID may be a member of.

```bash
% id
uid=1001(lmc) gid=1001(lmc) groups=1001(lmc),27(sudo),998(vboxsf)

% ls -ld host
drwxrwx--- 1 root vboxsf 640 Sep 16 03:48 host/

% cd host
=> error 1
```

It looks like you need to check user.User.GroupIds() (https://pkg.go.dev/os/user#User.GroupIds) too/instead.

theclapp:
Fixes #1033.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -963,7 +963,7 @@ func (r *Runner) changeDir(ctx context.Context, path string) int {
 	if err != nil || !info.IsDir() {
 		return 1
 	}
-	if !hasPermissionToDir(info) {
+	if !hasPermissionToDir(path) {
 		return 1
 	}
 	r.Dir = path
--- a/interp/os_notunix.go
+++ b/interp/os_notunix.go
@@ -7,7 +7,6 @@ package interp
 
 import (
 	"fmt"
-	"os"
 )
 
 func mkfifo(path string, mode uint32) error {
--- a/interp/os_notunix.go
+++ b/interp/os_notunix.go
@@ -15,6 +15,6 @@ func mkfifo(path string, mode uint32) error {
 }
 
 // hasPermissionToDir is a no-op on Windows.
-func hasPermissionToDir(info os.FileInfo) bool {
+func hasPermissionToDir(string) bool {
 	return true
 }
--- a/interp/os_unix.go
+++ b/interp/os_unix.go
@@ -18,8 +18,8 @@ func mkfifo(path string, mode uint32) error {
 	return unix.Mkfifo(path, mode)
 }
 
-// hasPermissionToDir returns if the OS current user has execute permission
-// to the given directory
+// hasPermissionToDir returns true if the OS current user has execute
+// permission to the given directory
 func hasPermissionToDir(info os.FileInfo) bool {
 	user, err := user.Current()
 	if err != nil {
@@ -38,33 +38,37 @@ func hasPermissionToDir(info os.FileInfo) bool {
 		panic("unexpected info.Sys type")
 	}
 	perm := info.Mode().Perm()
+
 	// user (u)
-	if perm&0o100 != 0 && st.Uid == uint32(uid) {
-		return true
+	if st.Uid == uint32(uid) {
+		return perm&0o100 != 0
 	}
 
-	// Group perms only apply if you're not the owner of the directory.
-	gid, _ := strconv.Atoi(user.Gid)
-	inGroup := st.Gid == uint32(gid)
-	if st.Uid != uint32(uid) {
-		gids, _ := user.GroupIds()
-		for _, gid := range gids {
-			gid, _ := strconv.Atoi(gid)
-			if st.Gid == uint32(gid) {
-				// other users in group (g)
-				if perm&0o010 != 0 {
-					return true
-				}
-				inGroup = true
-			}
+	// group (g) -- check the users's actual group, and then all the other
+	// groups they're in.
+	gid, err := strconv.Atoi(user.Gid)
+	if err != nil {
+		return false // on POSIX systems, Gid should always be a decimal number
+	}
+	if st.Gid == uint32(gid) {
+		return perm&0o010 != 0
+	}
+	gids, err := user.GroupIds()
+	if err != nil {
+		// If we can't get the list of group IDs, we can't know if the group
+		// permissions, so default to false/no access.
+		return false
+	}
+	for _, gid := range gids {
+		gid, err := strconv.Atoi(gid)
+		if err != nil {
+			return false
+		}
+		if st.Gid == uint32(gid) {
+			return perm&0o010 != 0
 		}
 	}
 
-	// remaining users (o) -- only apply if you're not the owner and none of
-	// your groups match its group.
-	if perm&0o001 != 0 && st.Uid != uint32(uid) && !inGroup {
-		return true
-	}
-
-	return false
+	// remaining users (o)
+	return perm&0o001 != 0
 }
--- a/interp/os_unix.go
+++ b/interp/os_unix.go
@@ -6,11 +6,6 @@
 package interp
 
 import (
-	"os"
-	"os/user"
-	"strconv"
-	"syscall"
-
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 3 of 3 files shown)


theclapp:
It seems like if you are user id 1 & group id 2, and a directory is also user id 1 & group id 2, and the directory is 070 or 007, then you should be able to cd into the directory -- the "group" or "other" perms should let you in even if the "owner" perms don't, but you can't & they don't.

I don't know if this is "right" or not, but it is what both bash and zsh do on both macOS and Linux, so I'm sticking with it.

(To be clear: This package already did it that way, this comment was just to explain it a little, and this PR was just to make it look at all your groups, not just your "main" group.)

mvdan:
I'm starting to wonder if this is the wrong approach entirely - it feels like we're trying to reverse-engineer the Unix-like semantics. Which are probably pointless or more likely wrong on Windows or Plan9.

What if we were to open the directory and try to read from it? That should get the OS to tell us whether we can use the directory or not. Opening the directory alone might be enough to tell; if not, then we could call e.g. https://pkg.go.dev/os#File.Readdirnames with an argument `1` to get just the first directory entry name, to just perform the cheapest read operation possible.

theclapp:
I thought about that, and would prefer that approach.  I wasn't sure what operations to try to verity the exact permissions required.  It may take some experimentation.

There is `os.Chdir` (et al) but that doesn't seem right, either.

mvdan:
Actually performing `os.Chdir` would be correct if we can assume we own the entire process per https://github.com/mvdan/sh/issues/171, but in general we can't assume that, e.g. if a Go process is interpreting multiple scripts they can't be fighting each other over `os.Chdir` or other global state like `os.Setenv`.

theclapp:
> but in general we can't assume that, e.g. if a Go process is interpreting multiple scripts

Yes, exactly.  My app might be doing exactly that.  Now, that's all it does, so if sh used a lock surrounding the global Chdir, and exported it (or made some api public), then apps that used sh and were aware of the problem could cooperate.

That still seems suboptimal.  I'd rather figure out some other (more local/specific) system call (or combination thereof) that exactly mirrors the pass/fail semantics of chdir.

mvdan:
Doesn't open and a possible read afterwards mimic the same behavior, without actually changing the directory? I would bet that it does.

theclapp:
No.  For --x and -wx, `cd` should succeed, but Open/OpenFile fails (so there's nothing to Read).  For r-- and rw-, `cd` should fail, but Open and Readdirnames both succeed.  Ditto ReadDir and Readdir.  So Open and Read* look at the "read" bit.

I don't see any Go call that would specifically examine the "execute" bit of a directory.

theclapp:
It looks like golang.org/x/sys/unix.Access (https://pkg.go.dev/golang.org/x/sys@v0.12.0/unix#Access) does the trick!

theclapp:
id: uid lmc, gid staff, other guids: contains admin, but not daemon

test directory:
```
d---------  3 daemon  staff   96 Sep 28 10:59 g0/
d-----x---  3 daemon  staff   96 Sep 28 10:59 g1/
d----w----  3 daemon  staff   96 Sep 28 10:59 g2/
d----wx---  3 daemon  staff   96 Sep 28 10:59 g3/
d---r-----@ 3 daemon  staff   96 Sep 28 10:59 g4/
d---r-x---@ 3 daemon  staff   96 Sep 28 10:59 g5/
d---rw----@ 3 daemon  staff   96 Sep 28 10:59 g6/
d---rwx---@ 3 daemon  staff   96 Sep 28 10:59 g7/
d---------  2 daemon  admin   64 Sep 28 11:36 g_0/
d-----x---  2 daemon  admin   64 Sep 28 11:36 g_1/
d----w----  2 daemon  admin   64 Sep 28 11:36 g_2/
d----wx---  2 daemon  admin   64 Sep 28 11:36 g_3/
d---r-----@ 2 daemon  admin   64 Sep 28 11:36 g_4/
d---r-x---@ 2 daemon  admin   64 Sep 28 11:36 g_5/
d---rw----@ 2 daemon  admin   64 Sep 28 11:36 g_6/
d---rwx---@ 2 daemon  admin   64 Sep 28 11:36 g_7/
d---------  3 daemon  daemon  96 Sep 28 10:59 o0/
d--------x  3 daemon  daemon  96 Sep 28 10:59 o1/
d-------w-  3 daemon  daemon  96 Sep 28 10:59 o2/
d-------wx  3 daemon  daemon  96 Sep 28 10:59 o3/
d------r--@ 3 daemon  daemon  96 Sep 28 10:59 o4/
d------r-x@ 3 daemon  daemon  96 Sep 28 10:59 o5/
d------rw-@ 3 daemon  daemon  96 Sep 28 10:59 o6/
d------rwx@ 3 daemon  daemon  96 Sep 28 10:59 o7/
d---------  3 lmc     staff   96 Sep 28 10:59 u0/
d--x------  3 lmc     staff   96 Sep 28 10:59 u1/
d-w-------  3 lmc     staff   96 Sep 28 10:59 u2/
d-wx------  3 lmc     staff   96 Sep 28 10:59 u3/
dr--------@ 3 lmc     staff   96 Sep 28 10:59 u4/
dr-x------@ 3 lmc     staff   96 Sep 28 10:59 u5/
drw-------@ 3 lmc     staff   96 Sep 28 10:59 u6/
drwx------@ 3 lmc     staff   96 Sep 28 10:59 u7/
```

Test sh code:
```bash
for f in * ; do if cd $f >& /dev/null ; then echo $f worked ; cd .. ; else echo $f failed ; fi ; done
finished huckshrc
g0 failed
g1 worked
g2 failed
g3 worked
[...]
```

test Go code:
```go
func (r *Runner) changeDir(ctx context.Context, path string) int {
	if path == "" {
		path = "."
	}
	path = r.absPath(path)
	info, err := r.stat(ctx, path)
	if err != nil || !info.IsDir() {
		return 1
	}
	accessErr := unix.Access(path, unix.X_OK)
	hptd := hasPermissionToDir(info)
	if hptd != (accessErr == nil) {
		log.Printf("%s: hptd: %v, accessErr: %v", path, hptd, accessErr)
	}
	if !hptd {
		return 1
	}
	r.Dir = path
	r.setVarString("OLDPWD", r.envGet("PWD"))
	r.setVarString("PWD", path)
	return 0
}
```

So the Go code runs both unix.Access and our already-defined hasPermissionToDir and prints stuff when they differ, and they never did.

theclapp:
Finally got all checks to pass.  No code changes, I just kept re-running the checks.  I dunno if that was the right approach.  🤷‍♂️ 😆

mvdan (approved):
TIL about the access syscall:

> access() checks whether the calling process can access the file pathname.
> R_OK, W_OK, and X_OK test whether the file exists and grants read,  write,  and execute permissions, respectively.

I am also personally surprised that "execute" permission is enough to cd into a directory, and "read" is unnecessary, but you've clearly tested it :)

Either way, love to delete all this code. Depending on x/sys/unix is always a bit tricky with portability, but this is already `os_unix.go`.

theclapp:
It's better than it was, but is still wrong for set-uid programs.

> The check is done using the calling process's real UID and GID, rather than the effective IDs as is done when actually attempting an operation (e.g., [open(2)](https://manned.org/open.2)) on the file.

I saw that in the `access` manpage, but, to be honest, just didn't understand it till now.

So, for example, if I (user "lmc") run gosh as set-uid root, I should be able to cd into any directory with any of the execute bits (user, group, other) set (since I'm running as root), but since `Access` checks the real UID ("lmc") not the effective UID ("root"), I can't.  (I think.  I'm not sure I've built my test-case correctly.)

I think I need to use `Faccessat` with a mode including AT_EACCESS.

But as I said, it's still better than it was.

Issue #1037, opened 2023-10-13
Ignored files not ignored when passed by name


dcd-arnold:
Hi @mvdan,

thank you for `shfmt`. I [have read](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#examples) that a directive in the `.editorconfig` like this:

```ini
# Ignore the entire "third_party" directory.
[third_party/**]
ignore = true
```

... should *ignore* the directory. However, this seems only true for the usage of `--find`. If I call `shfmt third_party/script.sh`, `shfmt` formats the file. I expect `shfmt` to output the file without any alternation, since the file is exempt from formatting.

Is this a bug or intended? I am asking because I am building a setup for CI and multiple editors and I would like to exclude some files from formatting.

Have a nice weekend,
dcd-arnold

mvdan:
That is indeed expected. The "ignore" logic only applies when walking directories, like `shfmt -w .`. When giving filenames directly, the tool assumes that you do want to format that file no matter what.

In a way, why else would you be running that command? Or, put another way, what could the tool reasonably do? Failing, or printing no stdout rather than the formatted source, both seem counter-intuitive.

That said, this is not well documented - I will update the man page.

dcd-arnold:
Hi @mvdan,

thank you for getting back to me that quickly.

> That is indeed expected. The "ignore" logic only applies when walking directories, like `shfmt -w .`.

You do mean `-f`, right? Or am I not getting something?

> When giving filenames directly, the tool assumes that you do want to format that file no matter what. In a way, why else would you be running that command? 

I get the rational behind it. However, this places the burden to determine which file to call `shfmt` for on the caller. I have multiple use cases to address, e.g. CI, VSCode, neovim, and that on multiple repositories. Each caller needs some logic to determine which file to run on. Usually I stuff that in some config file and have the tool to "just work", agnostic to the caller. 

> Or, put another way, what could the tool reasonably do? Failing, or printing no stdout rather than the formatted source, both seem counter-intuitive.

I completely agree, *printing no stdout* and *failing* are both counter intuitive, and not very useful for that matter. I would like to propose to print the file *without alternation*, like `cat`. That way `shfmt` behaves very predictable (there is always the processed file on stdout). Which is also the way prettier behaves (I only checked now to make sure).

> That said, this is not well documented - I will update the man page.

That would be very appreciated. However, the current behavior is rather counter-intuitive.

mvdan:
[...]

PR #1038, opened 2023-10-13
cmd/shfmt: document how EditorConfig's "ignore" works

mvdan:
(see commit message)

For #1037.

### DIFF
--- a/cmd/shfmt/shfmt.1.scd
+++ b/cmd/shfmt/shfmt.1.scd
@@ -131,7 +131,10 @@ keep_padding       = true
 # --func-next-line
 function_next_line = true
 
-# Ignore the entire "third_party" directory.
+# Ignore the entire "third_party" directory when calling shfmt on directories,
+# such as "shfmt -l -w .".
+# When formatting regular files explicitly, like "shfmt -w third_party/foo.sh",
+# the files are always formatted regardless of this setting.
 [third_party/**]
 ignore = true
 ```


mvdan:
@dcd-arnold let me know if this is clearer now.

cmd/shfmt/shfmt.1.scd
```diff
@@ -131,7 +131,10 @@ keep_padding       = true
 # --func-next-line
 function_next_line = true
 
-# Ignore the entire "third_party" directory.
+# Ignore the entire "third_party" directory when formatting directories,
+# such as "shfmt -l -w .".
```
dcd-arnold, line 135:
```suggestion
# Ignore the entire "third_party" directory when operating on a directory,
# e.g. "shfmt -l -w ." or "shfmt -f some/dir".
```

dcd-arnold (commented):
This makes it clearer. However, this is still counter-intuitive.

cmd/shfmt/shfmt.1.scd
mvdan: not sure I follow - is the addition of a second example what makes it clearer? both examples are pretty much the same, what matters is that the argument is a directory. I'd prefer to keep the man page short in general.

dcd-arnold: > not sure I follow - is the addition of a second example what makes it clearer? both examples are pretty much the same, what matters is that the argument is a directory.

And that exactly needs to be pointed out more precisely. It ignores *the entire "third_party" directory*, **if and only if** a directory is passed to `shfmt`. It does nothing otherwise. It is not only working "when formatting" directories, but whenever the user operates on a folder. And this statement will be ignored if a file is passed directly. 

> I'd prefer to keep the man page short in general.

Me, too. 

Maybe something like this then?

```suggestion
# Ignore the entire "third_party" directory when calling shfmt on directories, e.g. "shfmt -l -w .". 
```

mvdan: Thanks, done.

Issue #1018, opened 2023-07-15
_js: need a new version released again


alvis:
Similar to #760, the current version is now long outdated again.
I've hit an issue with `Part` become `undefined` when parsing an arg with quote e.g. `command "some quoted arg"`

It may be fixed with the go version, but without a new js npm release I'm not sure.

mvdan:
I checked briefly, and I don't think the Go library has this bug. The bug might well be in the JS code I wrote to expose the interface without blocking calls.

GopherJS has gone mostly unmaintained again, unfortunately; their last release only supports Go 1.18, and at this point we've required Go 1.20 for some time, with 1.21 having been released months ago.

I think continuing to invest energy into a GopherJS-powered library is simply not worth it. It was a neat experiment and it proved useful to some extent to some people, and for that reason I'm inclined to leave it be for now. But I also am not able to continue updating it even if I have the time.

In the medium term I think providing a wasm library will be the more future-proof approach. I know that's not a satisfactory answer to those needing something that works today, but the reality is that GopherJS doesn't work today either.

mvdan:
You might also want to try https://github.com/un-ts/sh-syntax in the meantime.

alvis:
Thanks @mvdan Maintaining a lib is not easy and I understand completely. Thanks for the suggestion. Will have a try

PR #1049, opened 2023-12-29
Update JS package to use go1.18.10

esimkowitz:
I've updated the JS package to the latest version of gopherjs, which supports go1.18.10. I've tested that this builds and passes the tests.

I've updated the `build` script to prompt the user to install the Go SDK if it is not already present.


mvdan:
Interesting timing, not sure if you saw my recent comment at https://github.com/mvdan/sh/issues/1018#issuecomment-1871582275.

Sure, I can try doing one more release with an outdated version of Go and a GopherJS beta release. It might work okay, or it might not. It would probably be the last release that I do with GopherJS, so I'd probably add a warning at the top of the npm package description to be clear about that.

I'll merge a simplified version of this given that now we can rely on `GOTOOLCHAIN` to run a precise version of Go without needing `go install`. I'll do that and try doing a release when I have better internet.

Issue #1053, opened 2024-01-28
Inline formatting single file: Breaks symlink and creates a regular file instead of retaining symlink


mechanicker:
Before run:
```
$ ls -l shfmt.test*|sed -E 's#(dkrishnamurthy)|(staff)#dev#g'
-rw-r--r-- 1 dev dev 50 Jan 28 06:43 shfmt.test
lrwxr-xr-x 1 dev dev 10 Jan 28 06:14 shfmt.test.ln -> shfmt.test
```

Note: `shfmt.test.ln` is a symlink

After running `shfmt -w shfmt.test.ln`:
```
$ ls -l shfmt.test*|sed -E 's#(dkrishnamurthy)|(staff)#dev#g'
-rw-r--r-- 1 dev dev 50 Jan 28 06:43 shfmt.test
-rwxr-xr-x 1 dev dev 47 Jan 28 06:43 shfmt.test.ln*
```

Note: `shfmt.test.ln` is a formatted regular file

mvdan:
Duplicate of https://github.com/mvdan/sh/issues/843.

Issue #843, opened 2022-04-06
cmd/shfmt: replaces symlinks with regular files


benjamineskola:
If `shfmt` is run on a file that is a symbolic link, and writes to that file, it breaks the symbolic link and replaces it with a copy of the original file (formatted accordingly).

```
$ printf "#!/bin/sh\nif true; then\ntrue\nfi\n" > test.sh
$ ln -s test.sh test2.sh
$ ls -l
.rw-r--r-- 32 ben  6 Apr 17:31 test.sh
lrwxr-xr-x  7 ben  6 Apr 17:31 test2.sh -> test.sh
$ shfmt -i 2 -w test2.sh
$ ls -l
.rw-r--r-- 32 ben  6 Apr 17:31 test.sh
.rwxr-xr-x 34 ben  6 Apr 17:31 test2.sh
```

Expected behaviour would be to write to the underlying file, preserving the symbolic link.

mvdan:
I think the expected behavior should rather be to skip symbolic links entirely; otherwise `shfmt -w .` when there's a symbolic link from `foo.sh` to `bar.sh` could result in formatting the same file twice, or causing a data race due to writing to the same path twice concurrently (once shfmt becomes parallelized). There's also other nasty edge cases that symbolic links could cause, such as a symbolic link pointing to a parent directory, meaning that `shfmt -w .` could potentially overwrite files anywhere in the filesystem, not just under the current directory.

In general, if one runs shfmt over an entire source repository, I assume that the symlink targets would already get formatted, so I think skipping the links will generally be fine. Do you want to send a patch with a test?

benjamineskola:
I'm not sure about ignoring symbolic links _entirely_ — if a symbolic link was specified as an explicit parameter then I think it would be surprising for it to be ignored. However, I think you're right that it should be safe to ignore symbolic links when recursing. (I think this is the default for tools like `find`.)

However, this has made me realise the problem is not limited to symbolic links but also occurs with hard links:

```
$ printf "#!/bin/sh\nif true; then\ntrue\nfi\n" > test.sh
$ ln test.sh test2.sh
$ stat -f '%i %N' *.sh
58677185 test.sh
58677185 test2.sh
# %i is the inode number; note that these are identical
$ shfmt -w test2.sh
$ stat -f '%i %N' *.sh
58677185 test.sh
58677241 test2.sh
# inode numbers now different
```

[...]

PR #1054, opened 2024-01-28
Resolve symlink when formatting single file

mechanicker:
Fix for: https://github.com/mvdan/sh/issues/1053

Using regular `os.Open` on symlink, formatting and writing back the output breaks the link and creates a new file instead.

Following a symlink automatically is something to decide. Do we trust the use to have done due diligence when passing a symlink OR avoid surprise and return an error for symlinks?

This change resolves the symlink and operates on resolved path, updates the source path and retains the symlink.


mvdan:
Thanks, but I think the right fix would be what is already described in https://github.com/mvdan/sh/issues/843#issuecomment-1344137219. Also, whatever fix we end up with would need tests.

PR #1058, opened 2024-02-13
cmd/shfmt: document when --filename is needed for EditorConfig support

### LINKED ISSUE #1055: cmd/shfmt: document how --filename relates to EditorConfig support
I'm imagining something like the following:

```shell
shfmt --editorconfig-path=../relative/path/to/.editorconfig foo.sh
shfmt --editorconfig-path=/absolute/path/to/.editorconfig foo.sh
```

This becomes useful when integrating shfmt into other applications/plugins. Please see the following PR, if interested, for Gradle plugin integration: https://github.com/diffplug/spotless/pull/2031.

A different flag name can be used than what I provided, but this at least gives the gist of what I'm looking for.

mvdan:
(see commit message)

Fixes #1055.

### DIFF
--- a/cmd/shfmt/shfmt.1.scd
+++ b/cmd/shfmt/shfmt.1.scd
@@ -25,7 +25,7 @@ predictable. Some aspects of the format can be configured via printer flags.
 
 ## Generic flags
 
-*-version*
+*--version*
 	Show version and exit.
 
 *-l*, *--list*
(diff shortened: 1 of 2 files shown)


mvdan:
cc @tcrawford-figure

cmd/shfmt/shfmt.1.scd
```diff
@@ -73,6 +73,9 @@ predictable. Some aspects of the format can be configured via printer flags.
 *-filename* str
 	Provide a name for the standard input file.
 
+	Use of this flag is necessary for EditorConfig support to work,
```
tcrawford-figure, line 76: .."for EditorConfig support to work with standard input,"

I think that makes it a little clearer in this standalone sentence that this isn't necessary when a file path is provided as the last parameter.

mvdan: Hm, I actually had it written like that first, but the section already begins with "for the standard input file", so I thought it would be redundant. I agree that it's better to be explicit, on second thought.

PR #1060, opened 2024-02-16
cmd/shfmt: fix typo in shfmt.1.scd

eltociear:
overriden -> overridden


mvdan (approved):
Oops, thanks for spotting.

PR #1062, opened 2024-02-23
cmd/shfmt: Add support for minify and simplify via EditorConfig

### LINKED ISSUE #819: cmd/shfmt: support editorconfig properties for `-s` and `-mn`
Right now I have to keep my text editor's autoformatting code in sync with my project's Makefile to invoke `shfmt`. It would be nice to keep that version controlled on my project's `.editorconfig` file. Is that possible? If not, is it something you'd be open to a PR adding support for?

Thanks!

ColemanTom:
I've never used `go` before, so I this is a naive, but working approach to support `simplify` and `minify` in editorconfig files. I have not looked at adding tests yet, but will add tests if this code change approach is agreeable.

Manual testing below.

```bash
$ for opt in "" "simplify=true" "minify=true"; do echo $'[[shell]]\nindent=4\n'"$opt" > .editorconfig; echo "### $opt"; shfmt --diff f.sh; done
###
--- f.sh.orig
+++ f.sh
@@ -1,11 +1,11 @@
 #!/bin/bash

 ff() {
-local h=asdffdfd
+       local h=asdffdfd
 }
 g=a

-echo $(( 1 + 2 ));
+echo $((1 + 2))
 ff
 echo $h
 [[ "$HOME" == "$HELLO" ]]
### simplify=true
--- f.sh.orig
+++ f.sh
@@ -1,11 +1,11 @@
 #!/bin/bash

 ff() {
-local h=asdffdfd
+       local h=asdffdfd
 }
 g=a

-echo $(( 1 + 2 ));
+echo $((1 + 2))
 ff
 echo $h
-[[ "$HOME" == "$HELLO" ]]
+[[ $HOME == "$HELLO" ]]
### minify=true
--- f.sh.orig
+++ f.sh
@@ -1,11 +1,9 @@
 #!/bin/bash
-
-ff() {
+ff(){
 local h=asdffdfd
 }
 g=a
-
-echo $(( 1 + 2 ));
+echo $((1+2))
 ff
 echo $h
-[[ "$HOME" == "$HELLO" ]]
+[[ $HOME == "$HELLO" ]]
```

Adding, for e.g. `-s` or `-mn` to the `shfmt` call leads to consistent results (i.e. if `-s` is on, then for the `opts=`, `opts="simplify=true"`, we get the second result above, but the `opts="minify=true"` option stays consistent with `minify` respected).

Closes #819.

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -43,28 +43,32 @@ type multiFlag[T any] struct {
 }
 
 var (
+	// Generic flags.
 	versionFlag = &multiFlag[bool]{"", "version", false}
 	list        = &multiFlag[boolString]{"l", "list", "false"}
-
 	write       = &multiFlag[bool]{"w", "write", false}
-	simplify    = &multiFlag[bool]{"s", "simplify", false}
-	minify      = &multiFlag[bool]{"mn", "minify", false}
-	find        = &multiFlag[boolString]{"f", "find", "false"}
 	diff        = &multiFlag[bool]{"d", "diff", false}
 	applyIgnore = &multiFlag[bool]{"", "apply-ignore", false}
+	filename    = &multiFlag[string]{"", "filename", ""}
 
-	lang       = &multiFlag[syntax.LangVariant]{"ln", "language-dialect", syntax.LangAuto}
-	posix      = &multiFlag[bool]{"p", "posix", false}
-	filename   = &multiFlag[string]{"", "filename", ""}
+	// Parser flags.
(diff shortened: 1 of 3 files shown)


ColemanTom:
I have added some tests following the structure of other editorconfig tests.

mvdan:
Thanks! This looks broadly fine - I'll adjust the code to treat simplify and minify as formatting flags properly, for consistency with the rest of the flags, as otherwise mixing formatting flags and EditorConfig files gets very messy.

PR #1063, opened 2024-02-29
interp: add -s support for read builtin

davidhulick:
This implements the -s option for the read builtin using the ReadPassword function in the golang.org/x/term package. This is a less featureful (and as a result, simpler) implementation of the -s flag than is present in https://github.com/mvdan/sh/pull/865

Like PR865, I am also unsure of how to test the -s flag since it requires interacting with the terminal input.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -14,8 +14,10 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/muesli/cancelreader"
+	"golang.org/x/term"
 
 	"mvdan.cc/sh/v3/expand"
 	"mvdan.cc/sh/v3/syntax"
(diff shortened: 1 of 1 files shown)


interp/builtin.go
```diff
@@ -559,10 +561,15 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 		r.setErr(returnStatus(code))
 	case "read":
 		var prompt string
+		var line []byte
+		var err error
```
mvdan, line 565: keep these vars declared further below, to ont keep as much in scope at once.

interp/builtin.go
```diff
@@ -589,7 +596,12 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			r.out(prompt)
 		}
 
-		line, err := r.readLine(raw)
+		if silent {
+			line, err = term.ReadPassword(int(syscall.Stdin))
+		}else{
```
mvdan, line 601: did you run gofmt?

mvdan (approved):
Thanks! SGTM, just two nits. I agree testing this is hard and not really worth it, given that it's really three lines of simple code.

interp/builtin.go
davidhulick: Moved these variable declarations further down.

interp/builtin.go
davidhulick: My apologies! Fixed.

Issue #856, opened 2022-05-04
interp: Cancelling `Runner.Run`'s context doesn't abort `read` builtin


theclapp:
See title.

Is there some other way to signal a Runner (something analogous to SIGINT) that I'm missing?

mvdan:
One possible solution would be to have the interpreter constantly read from stdin in a background goroutine, and use a channel for builtins like `read`. We'd just have to worry about:

1) buffering stdin bytes until they are used
2) forwarding stdin to a command when it's executed

mvdan:
Thinking about this idea some more: this does mean that the interpreter will aggressively read before it technically needs to, but that might be a reasonable tradeoff to prevent reads from blocking potentially forever, until stdin or the entire program are closed.

theclapp:
Yeah, after some Googling the best I can find so far is https://benjamincongdon.me/blog/2020/04/23/Cancelable-Reads-in-Go/ , which presents a CancelableReader class.  It basically implements what you described above.

FWIW I don't think the interpreter should fork a reader goroutine until and unless somebody actually calls read.

mvdan:
> FWIW I don't think the interpreter should fork a reader goroutine until and unless somebody actually calls read.

That's fair, though I guess "we start reading early once `runner.Run` is first called" and "we start reading early once the first read is triggered" are both reasonable to me. I guess the former might be less inconsistent or surprising, but the latter is better in terms of not spawning the goroutine and doing work that might not be used at all - like when interpreting a small script that does not use stdin at all.

theclapp:
Building on the code from the blog I linked, this seems like it'd work.  (Lightly tested in the [playground](https://go.dev/play/p/uB1SOzy0S1t).)

```go
package ctxreader

import (
	"context"
	"io"
	"sync"
)

type CancelableReader struct {
	ctx  context.Context
	data chan []byte
	err  error
	r    io.Reader
	once sync.Once
}

func (c *CancelableReader) begin() {
	buf := make([]byte, 1024)
	for c.ctx.Err() == nil {
		n, err := c.r.Read(buf)
		if n > 0 {
			tmp := make([]byte, n)
			copy(tmp, buf[:n])
			select {
			case c.data <- tmp:
			case <-c.ctx.Done():
				return
			}
		}
		if err != nil {
			c.err = err
			close(c.data)
			return
		}
	}
}

func (c *CancelableReader) Read(p []byte) (int, error) {
	c.once.Do(func() { go c.begin() })
	select {
[...]

PR #1066, opened 2024-03-08
Cancellable reads; EOF on "read" sets var to ""

theclapp:
- Make reading cancellable. Not all input from stdin, just that done directly by the shell (i.e. the "read" builtin). Exec'ed programs still read directly from stdin's os.File and are not cancellable.
- If you press ^D (EOF) when reading into a shell variable, set the variable to "". This is consistent with bash & zsh.

Fixes https://github.com/mvdan/sh/issues/856.
See also https://github.com/mvdan/sh/pull/857 for my first try at this.


mvdan (commented):
I assume this is one of those that is hard to test properly, but I'd still like us to try, because the logic is non-trivial and feels fragile.

For example, with https://pkg.go.dev/os#Pipe to create a stdin file, and cancelling the runner context to mimic ^D, we could test a few edge cases with the `read` builtin:
* only newline (empty input)
* only ^D (empty input)
* some string and newline (non-empty input)
* some string and ^D (non-empty input)

mvdan:
Overall nothing against merging this feature though - it is one added dependency, but it seems like a light one, and I don't think we can possibly implement cancelling reads any way other than OS-specific code.

interp/builtin.go
```diff
@@ -606,6 +604,10 @@ func (r *Runner) builtinCode(ctx context.Context, pos syntax.Pos, name string, a
 			r.setVarString(name, val)
 		}
 
+		if err != nil {
+			return 1
+		}
```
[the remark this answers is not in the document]
theclapp: > can you document why we check the error after using the returned value from readLine now?

In retrospect, it's not really clear, but I mentioned it in the PR comment: so that if you press ^D (EOF) when reading into a shell variable, it sets the variable to "".

I'll be happy to add a comment to the actual code, of course. This is just so you know.

theclapp:
> I assume this is one of those that is hard to test properly, but I'd still like us to try, because the logic is non-trivial and feels fragile.

Sure, no worries.

theclapp:
> * only newline (empty input)
> * only ^D (empty input)
> * some string and newline (non-empty input)
> * some string and ^D (non-empty input)

It turns out the current code already works with the first, third, and fourth. I've added a test case with the second.

None of these cases actually test having a read in-progress and cancelling its context, so I wrote a separate test for that. It turns out it wasn't that difficult.

interp/interp_test.go
```diff
@@ -2982,6 +2982,27 @@ done <<< 2`,
...
+	// // some string and newline
+	// {
+	// 	"a=c; echo x | read a; echo -n $a",
+	// 	"x",
+	// },
+	// // some string and EOF
+	// {
+	// 	"a=d; echo -n y | read a; echo -n $a",
+	// 	"y",
+	// },
```
mvdan, line 3005: did you mean to leave most of these commented out?

interp/interp_test.go
```diff
@@ -3943,6 +3964,49 @@ func TestRunnerContext(t *testing.T) {
...
+	} else {
+		siR, siW, err := os.Pipe()
+		if err != nil {
+			t.Fatalf("Error calling os.Pipe: %v", err)
+		}
+		defer func() {
+			siW.Close()
+			siR.Close()
+		}()
+		r, _ = interp.New(interp.StdIO(siR, nil, nil))
```
mvdan, line 3990: nit: move interp.New and StdIO outside the if/else, you can just have each branch set the stdin file var

interp/builtin.go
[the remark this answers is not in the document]
mvdan: LGTM

mvdan (commented):
FYI CI appears broken

interp/interp_test.go
theclapp: Yes. Sort of. I mentioned in the commit:

- Add some test cases. Daniel suggested four, three of which already
  pass without this PR. I left them in but commented-out, in case
  somebody wants to look at them.

Should uncomment them or remove them entirely? Wasn't sure how you'd feel about that either way.

mvdan: Sorry, maybe I'm being slow as it's a bit late. If these test cases already behaved the way we want in master, why not enable them as regression tests, even if they're not directly related to this fix?

theclapp: Sure, I'm fine with that.

theclapp:
As noted in the commit: TestCancelreader failing intermittently under Linux (on a Debian VM), even though the read context is definitely cancelled. Not sure what's going on. Committing what I've got, for posterity.

theclapp:
> TestCancelreader failing intermittently under Linux

At a guess, this is because of, or at least related to, the data race reported in the CI tests. Will look more tomorrow.

theclapp:
@ghostsquad Wanna take a look?

mvdan (approved):
SGTM, thanks! If anyone else has thoughts or a follow-up review, I'm happy to get more PRs after this lands.

mvdan:
@theclapp do you want to squash your commits into one and write a single commit message? I could squash myself, only keeping the original commit message, but I suspect you might want to keep some of the text from the other commit messages, e.g. about the test.

theclapp:
Squashed 'em all.

mvdan (approved):

PR #1073, opened 2024-04-16
Allow case-insensitive filename globbing; add nocaseglob

theclapp:
- Add a FoldCase flag to the expand.Config struct, which is used to configure how expand.Fields behaves.
- Similarly, add a FoldCase Mode to pattern.Regexp. Setting the flag adds (?i) to the front of the generated regular expression.
- Use the former to set the latter when calling expand.Fields.
- Use all of the above to implement the "nocaseglob" shopt.

Rationale: I use expand.Fields to do filename expansion, for which I'd like to do case-insensitive matching. E.g. I'd like "c*" to match both "cmd" and "CHANGELOG.md"

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -68,6 +68,10 @@ type Config struct {
 	// "**".
 	GlobStar bool
 
+	// NoCaseGlob corresponds to the shell option that causes case-insensitive
+	// pattern matching in pathname expansion.
+	NoCaseGlob bool
+
 	// NullGlob corresponds to the shell option that allows globbing
 	// patterns which match nothing to result in zero fields.
 	NullGlob bool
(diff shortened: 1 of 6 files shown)


mvdan:
> Rationale: I use expand.Fields to do filename expansion, for which I'd like to do case-insensitive matching. E.g. I'd like "c*" to match both "cmd" and "CHANGELOG.md"

Can you give some context in terms of when is this useful for a shell? I see `nocaseglob` in `man bash`, which alone is probably enough to need this Go API, but I wonder if your need comes from elsewhere.

theclapp:
I have an terminal/shell app that uses this package as its shell interpreter. It also, as mentioned in this PR, uses expand.Fields to do filename expansion, like bash does when you press tab. So I'd like c<tab> to expand to both "cmd" and "CHANGELOG.md". My app translates "c<tab>" to "c*" and calls expand.Fields.

This is similar to if you set `shopt -s nocaseglob` in bash and type "c*<tab>".

Would you like me to add the "nocaseglob" shopt and make it set this flag?

mvdan:
OK gotcha, makes sense. If you want to wire it up to the interpreter that would be nice as a way to make this change a bit more well rounded and get the easy win, but it's not a must. I'll review later.

theclapp:
I added the `nocaseglob` shopt, force-pushed, and updated the title and description of the PR.

mvdan (approved):
Thank you! Appreciate the added comments too.

PR #1075, opened 2024-04-18
DebugPrint: Print the string value of tokens

theclapp:
In DebugPrint, if a value implements fmt.Stringer, and isn't a zero value, display its string value in addition to the default %#v format.

For the most part, this prints the string value of "tokens", e.g. for a syntax.Redirect.Op, instead of "Op: 0x3b", you get "Op: 0x3b (>&)".


mvdan (approved):
Thanks!

PR #1078, opened 2024-05-28
Allow building with Tinygo

diamondburned:
This PR contains commits that allow `mvdan.cc/sh/v3` to be built using Tinygo. The commit messages will contain more information.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,34 @@ jobs:
     - name: Test as root, without cgo, and with busybox
       run: docker run -v="$PWD:/pwd" -w=/pwd -e=CGO_ENABLED=0 golang:1.22.0-alpine go test ./...
 
+  test-tinygo:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install Tinygo
+      uses: acifani/setup-tinygo@v2
+      with:
+        tinygo-version: 0.32.0
+
+    - name: Create temporary main package
+      run: |
+        mkdir _tinygo && cat<<EOF > _tinygo/main.go
+        package main
+
+        import (
+          _ "mvdan.cc/sh/v3/expand"
+          _ "mvdan.cc/sh/v3/interp"
+          _ "mvdan.cc/sh/v3/syntax"
+        )
+
+        func main() {}
+        EOF
+
+    - name: Build with Tinygo natively
+      run: tinygo build -o /dev/null ./_tinygo
+    - name: Build with Tinygo on WASM
+      run: tinygo build -o /dev/null -target wasm ./_tinygo
+
   docker:
     name: Build and test Docker images
     # Only deploy if previous stages pass.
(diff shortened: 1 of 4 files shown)


mvdan:
Thanks, I'm fine with supporting tinygo builds and adding it to CI.

I'm not super pumped about adding tinygo build tags, though. Ideally, tags like `cgo` or `purego` would be enough for the code on master to already be portable. For example, with either `CGO_ENABLED=0` or `-tags=purego`, `os/user.Lookup` on Linux should be a stub that returns nothing, why is that not enough for Tinygo?

https://tinygo.org/docs/reference/lang-support/stdlib/#osuser does show a failure, but that seems like a bug that should be tracked/fixed upstream.

diamondburned:
> https://tinygo.org/docs/reference/lang-support/stdlib/#osuser does show a failure, but that seems like a bug that should be tracked/fixed upstream.

I made https://github.com/tinygo-org/tinygo/pull/4278 to address this.

mvdan:
With https://github.com/tinygo-org/tinygo/pull/4278 merged a while back, I assume we can now get rid of the build tags here?

diamondburned:
I got rid of a few stubs, but I still have to stub `os.SameFile` for Tinygo + WASM to work :(

mvdan:
That's very odd. Upstream Go doesn't have SameFile behind any build tags, even though they support js/wasm. It seems to me like TinyGo should at the very least publish the stub themselves; the APIs should compile on all platforms, even if all they do is always fail on the unsupported platforms.

dolmen:
Apparently tinygo got support for SameFile on wasip2 with tinygo-org/tinygo#4027 (see [types_unix.go](https://github.com/tinygo-org/tinygo/commit/9cb263479c4b98f2d28889ca1acc297454e0d875#diff-fdefea98f0ad54e65a918f4353fd1ed534effca6ecd30e05530c47fdf2ac62e9)). 

@diamondburned Is the SameFile exception still needed?

.github/workflows/test.yml
```diff
@@ -40,6 +40,34 @@ jobs:
...
+          _ "mvdan.cc/sh/v3/syntax"
+        )
+
+        func main() {}
+        EOF
+
+    - name: Build with Tinygo natively
+      run: tinygo build -o /dev/null ./_tinygo
+    - name: Build with Tinygo on WASM
+      run: tinygo build -o /dev/null -target wasm ./_tinygo
```
JounQin, line 69: It doesn't check whether the output behaviors correctly.

And I'm already using `tinygo` with this package at https://github.com/un-ts/sh-syntax for a long time, although the json `encoding`/`decoding` parts are levered by [`easyjson`](https://github.com/mailru/easyjson) at https://github.com/un-ts/sh-syntax/blob/main/processor/structs_easyjson.go

reteps:
I got this building in tinygo `0.38.0` without any workarounds.

mvdan:
I'm going to close this for now as it doesn't seem like it's needed anymore and there's no activity. But my comment above still applies in terms of being welcoming of future contributions.

PR #1080, opened 2024-06-27
Implement -O and -G tests

theclapp:
Implement -O (syntax.TsUsrOwn, "file exists and is owned by the effective user id") and -G (syntax.TsGrpOwn, "file exists and is owned by the effective group id") tests, for non-Windows only.

Under Windows, still panics, as before.

-O and -G are hard/not applicable for Windows, since it doesn't really have the concept of "file owners", only ACLs, and it's not clear how to translate one to the other.

### DIFF
--- a/interp/test.go
+++ b/interp/test.go
@@ -8,11 +8,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/user"
 	"regexp"
-	"runtime"
-	"strconv"
-	"syscall"
 
 	"golang.org/x/term"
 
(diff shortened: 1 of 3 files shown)


theclapp:
Oops, this won't even *build* under Windows. My bad. I'll get back to you.

interp/test.go
```diff
@@ -200,6 +204,24 @@ func (r *Runner) unTest(ctx context.Context, op syntax.UnTestOperator, x string)
...
+		}
+		fi, err := os.Stat(x)
+		if err != nil {
+			return false
+		}
+		u, err := user.Current()
+		if err != nil {
+			return false
+		}
+		if op == syntax.TsUsrOwn {
```
ccoVeille, line 219: What about this ?

```suggestion
stat, ok :=  fi.Sys().(*syscall.Stat_t)
if !ok {
      return false
}
   
if op == syntax.TsUsrOwn {
       return strconv.Itoa(uid) == stat.Uid
} 
return return strconv.Itoa(gid) == stat.Gid
```

theclapp: This wouldn't build either, though. The problem is that under Windows the `syscall.Stat_t` type doesn't even exist, not whether the type assertion would succeed or fail.

I think I thought/hoped that the `runtime.GOOS` check would be resolved statically at compile time and the compiler would ignore the rest of the `case`, but alas, that was not, uh, the case. :) In reality, I think the GOOS check *is* resolved statically at compile time, but the rest of the code still needs to compile.

ccoVeille: ah ok, the need was different than I thought

mvdan (approved):
Thank you!

PR #1089, opened 2024-08-15
shfmt: skip non-regular files

kolyshkin:
I had a named pipe in my source directory, and `shfmt .` suddenly stopped working.

It probably doesn't make sense to try reading non-regular files.


mvdan:
Thanks for the PR! This is a pretty silly but significant omission on my part. I think the fix should be in `fileutil`, and with a test, so I'll do a slightly different fix while thanking you in the commit message. And apologies for the slowness.

Issue #1096, opened 2024-09-20
cmd/shfmt: support null byte separators in --list and --find


sdavids:
@msanders

https://github.com/mvdan/sh/issues/288#issuecomment-579582229

> It might make sense to add a -z or -0 flag for shfmt -f to use safely with xargs -0. (See e.g. find -print0, git ls-files -z, grep --null, etc.)

[How can I find and safely handle file names containing newlines, spaces or both?](https://mywiki.wooledge.org/BashFAQ/020)

Also for `shfmt -l`.

---

Adding `--find0` and `--list0` would not lead to ambiguous/erroneous cases like `shfmt -w -0 .`.

It also mirrors `find`'s `-print`/`-print0` and `-fprint `/`fprint0`.

mvdan:
Thanks. How often does this realistically come up?

If we add this, I would rather not add new flags. We can tweak `--find` and `--list` so that they also accept an optional zero parameter like `--find=0` and `--list=0`. I adapted an existing boolean flag to take optional string values too in a different project, you're free to copy paste the idea:

https://github.com/mvdan/xurls/blob/c5b6e994e5db07c3f691bbb8c044b34dd20720f8/cmd/xurls/main.go#L36-L44

sdavids:
> Thanks. How often does this realistically come up?

Every time one wants to use `--list` with `xargs`.

sdavids:
```shell
-l[=0], --list[=0]  list files whose formatting differs from shfmt's;
                    paths are separated by a newline or a null character if -l=0
                    (corresponding to the -0 option of xargs)
-f[=0], --find[=0]  recursively find all shell files and print the paths;
                    paths are separated by a newline or a null character if -l=0
                    (corresponding to the -0 option of xargs)
```

⇑

Your preference?

I am not sure, if that is more readable/discoverable than:

```shell
-l,  --list    list files whose formatting differs from shfmt's;
               paths are separated by a newline
-l0, --list0   list files whose formatting differs from shfmt's;
               paths are separated by a null character
               corresponding to the -0 option of xargs
-f, --find     recursively find all shell files and print the paths;
               paths are separated by a newline
-f0, --find0   recursively find all shell files and print the paths;
               paths are separated by a null character
               corresponding to the -0 option of xargs
```


```shell
-l,  --list    list files whose formatting differs from shfmt's;
               paths are separated by a newline
-f, --find     recursively find all shell files and print the paths;
               paths are separated by a newline
[...]

PR #1097, opened 2024-09-20
Add --find0 and --list0 flags

sdavids:
#1096

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -27,6 +27,16 @@ import (
 	"mvdan.cc/sh/v3/syntax/typedjson"
 )
 
+type boolString string
+
+func (s *boolString) Set(val string) error {
+	*s = boolString(val)
+	return nil
+}
+func (s *boolString) Get() any       { return string(*s) }
+func (s *boolString) String() string { return string(*s) }
+func (*boolString) IsBoolFlag() bool { return true }
+
 type multiFlag[T any] struct {
 	short, long string
 	val         T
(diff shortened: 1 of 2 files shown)


sdavids:
One thing I noticed but I did not want to create a new issue or pollute this PR:

https://github.com/mvdan/sh/blob/57b7613429b5ac2918c3218f321107f567ba2165/_js/go.mod#L3

should it not be the same as:

https://github.com/mvdan/sh/blob/57b7613429b5ac2918c3218f321107f567ba2165/go.mod#L3

cmd/shfmt/main.go
```diff
@@ -420,9 +447,13 @@ func formatPath(path string, checkShebang bool) error {
 		}
 		readBuf.Write(copyBuf[:n])
 	}
-	if find.val {
+	if find.val == "true" {
 		fmt.Println(path)
 		return nil
+	} else if find.val == "0" {
```
mvdan, line 453: use a switch on find.val

cmd/shfmt/main.go
```diff
@@ -484,8 +515,11 @@ func formatBytes(src []byte, path string, fileLang syntax.LangVariant) error {
 	printer.Print(&writeBuf, node)
 	res := writeBuf.Bytes()
 	if !bytes.Equal(src, res) {
-		if list.val {
+		if list.val == "true" {
 			fmt.Println(path)
+		} else if list.val == "0" {
```
mvdan, line 520: use a switch on list.val

mvdan (approved):
thanks, just two nits

mvdan (approved):
Thanks!

mvdan:
FYI force-pushed here to resolve the conflict, squash the commits, and write a slightly longer commit message.

Issue #1106, opened 2024-10-31
Sequential Backslashes are removed during interpretation.


cfogrady:
Sequential Backslashes seem to be reduced to a single backslash. Is there a purposeful reason behind this or am I correct in believing this to be a bug?

Example:
`gosh -c 'echo "hello \\\\\\ world"'` yields `hello \ world`

Running the same thing in bash:
`bash -c 'echo "hello \\\\\\ world"'` yields `hello \\\ world` matching my expectations.

In code:
```
func TestMvdanShellPreservesSequentialBackslashes(t *testing.T) {
	parser := syntax.NewParser(syntax.Variant(syntax.LangBash))
	prog, _ := parser.Parse(strings.NewReader(`echo "Hello \\\\\\ World"`), "$0")
	writer := &bytes.Buffer{}
	bash, _ := interp.New(interp.StdIO(nil, writer, writer))
	bash.Run(context.Background(), prog)
	want := `Hello \\\ World`
	if writer.String() != want {
		t.Errorf(`echo "Hello \\\\\\ World" = %v, want %v`, writer.String(), want)
	}
}
```

I believe this is an issue in the interp as running a `syntax.DebugPrint` on `prog` yields:
```
Parts: []syntax.WordPart (len = 1) {
.  .  .  .  .  .  .  .  .  0: *syntax.Lit {
.  .  .  .  .  .  .  .  .  .  ValuePos: 1:7
.  .  .  .  .  .  .  .  .  .  ValueEnd: 1:25
.  .  .  .  .  .  .  .  .  .  Value: "Hello \\\\\\\\\\\\ World"
.  .  .  .  .  .  .  .  .  }
.  .  .  .  .  .  .  .  }
```

mvdan:
Thanks for reorting, this does seem like a bug.

Out of curiosity, how did you find this? Are you using the interpreter via the Go API?

cfogrady:
Yup. Using the interp Go API for some shell scripts with gnarly arguments including some regexps. Some of the regexps were off after going through my program and after debugging, I managed to localize the issue to the library as above.

cfogrady:
Further investigation seems to indicate this is a problem specific to double quoted literals.
`gosh -c "echo 'hello \\\\\\ world'"` yields what I would expect: `hello \\\ world`.

cfogrady:
I'll prep a PR after lunch.
[...]

PR #1107, opened 2024-11-04
expand: fix sequential '\' in double quote literals

cfogrady:
Fix for https://github.com/mvdan/sh/issues/1106


mvdan:
Thanks! I actually avoid unit tests where possible, and instead test the exposed API. In this case I would add integration tests in the interpreter, for example adding to these around line 400 in interp_test.go:

```
	// escaped chars
	{"echo a\\b", "ab\n"},
	{"echo a\\ b", "a b\n"},
	{"echo \\$a", "$a\n"},
	{"echo \"a\\b\"", "a\\b\n"},
	{"echo 'a\\b'", "a\\b\n"},
	{"echo \"a\\\nb\"", "ab\n"},
	{"echo 'a\\\nb'", "a\\\nb\n"},
	{`echo "\""`, "\"\n"},
	{`echo \\`, "\\\n"},
	{`echo \\\\`, "\\\\\n"},
	{`echo \`, "\n"},
```

I know that means there isn't test coverage in the same package, but I think proper coverage across the module is still fine. And, ultimately, most of these bugs in the expand package only matter for users of the interp package.

mvdan:
And avoiding unit tests also means that you don't need to add one more function to test.

cfogrady:
> And avoiding unit tests also means that you don't need to add one more function to test.

For clarity, do you prefer to keep the logic inlined? My default is to split out what I would consider to be tangential logic because I think it helps readability, but you're the maintainer.

mvdan:
Yes, please keep the code where it was - the func isn't particularly large or very indented, so I don't think it requires splitting up just yet.

cfogrady:
Function is inlined once again, and the (in my opinion) essential test cases have been to the interp_test.go test cases.

mvdan:
There was a minor gofmt issue which I fixed. I will squash-merge, as I prefer to keep the git history tidy. For future PRs, feel free to squash into a single commit yourself with a good commit message, and I'll merge as-is.

cfogrady:
Sounds good!

mvdan:
(as usual got sidetracked waiting for CI to pass, merging now)

PR #1113, opened 2024-12-02
pattern: fix outdated comment, and add godoc links

ccoVeille:
TranslatePattern and QuotePattern were renamed Regexp and QuoteMeta
in 2019 when code was moved out "syntax" package to "pattern" one.

See 2e72e3c7a085f23431db3c6cfb01da390fd5003c


mvdan (approved):
Thanks!

PR #1114, opened 2024-12-02
interp: improve method signature

ccoVeille:
this is a zero code changes. The signature was already the right one.
We are just using an alias, that will make things clearer.


mvdan (commented):
The signature change is nice, but there's no need to make the godoc longer. Please undo the doc changes.

ccoVeille:
I just did

mvdan:
The changes look the same to me.

ccoVeille:
My bad, it's OK now

PR #1123, opened 2025-01-20
shfmt: add -vb, --variable-braces printer option to put braces around variables

aneiosi:
As the name says, I've added the option to put braces around variables. The -vb option has no effect if -mn is used which overrides it.

Also updated the .1.scd file and the --help command to reflect the change.

I also took the liberty to align the --help command (OCD overwhelmed me xD) which forced me to modify --language-dialect to just --language. Feel free to remove this if it's overstepping.


mvdan:
Please don't spend more time on this PR. I haven't had time to look at it properly yet, but it's unlikely that we will add a flag like `--variable-braces`. Such a refactor or rewrite can very easily be written as an external tool, and `shfmt` has a very high bar for more options.

aneiosi:
Ok, thanks for the feedback

external tool it is, then :)

PR #1129, opened 2025-02-17
README: Add neoformat NeoVim / Vim plugin

offa:
Adds [neoformat](https://github.com/sbdchd/neoformat) to the list of related projects.


mvdan (approved):
Thanks!

PR #1135, opened 2025-03-10
3.11 regression (?) with nested expressions being expanded

jippi:
I noticed a possible regression while upgrading to 3.11 in https://github.com/jippi/dottie/pull/217 - this behavior was not present in 3.10 which i upgraded from

The expansion of `[${UNSET:-${BAR-defaultValue}}]` used to result in `[]` but now result in `[defaultValue]`

In zsh/bash running these echo's give the following output, which I think Expand should match

```
$ BAR=""; echo "[${UNSET:-${BAR-defaultValue}}]"
[]

$ BAR="set"; echo "[${UNSET:-${BAR-defaultValue}}]"
[set]
```

## NOTE

This is not a fix, but a reproducible test case to show-case the issue


mvdan:
You're right that there was a change in behavior here. It was intentional, even if potentially breaking for a few advanced users like yourself, because the original API could not describe unset variables which had a type, such as `declare -A foo` which is an associative array but still unset. See https://github.com/mvdan/sh/commit/7f87c18fc59b4a4342f7f006fcc330a523bd4b30.

I've just pushed https://github.com/mvdan/sh/commit/12b8c0a0254d7f874aa94dffb32972c9b07d7eca to clarify this further a bit.

In your case, your implementation should set `Variable.Set` to `true` when returning a set variable, i.e. with a value. See how I changed our implementations like `listEnviron` in the first commit above.

mvdan:
The test case you wrote here is a red herring, because `shell.Expand` takes `env func(string) string`, so `strEnviron("BAR=")` is the same as `strEnviron()` - in both cases, `env("BAR")` returns the empty string, so it is treated as unset. If you try similar tests with the full interpreter, which does distinguish empty from unset variables via https://pkg.go.dev/mvdan.cc/sh/v3/expand#Environ, then it does work as expected:

```
$ gosh
$ echo [${UNSET:-${BAR-defaultValue}}]
[defaultValue]
$ BAR=''; echo [${UNSET:-${BAR-defaultValue}}]
[]
$ BAR='set'; echo [${UNSET:-${BAR-defaultValue}}]
[set]
```

mvdan:
Closing this for now, per the above.

PR #1147, opened 2025-04-07
pattern: add NoGlobStar mode

paulhammond:
I'm working on some code that should support just POSIX compliant globs and had hoped to use the pattern package for this. Unfortunately that package currently has support for `**`, which is not part of the POSIX specification. This pull request adds a new mode which disables that functionality.

I think in an ideal world the mode would be called `GlobStar` and would enable `**` support, but I assume backwards compatibility is more important than API cleanliness.


mvdan (approved):
Thanks for the patch - you're absolutely right that we want to keep the default behavior as-is to not break users. In a future v4 we might want to rethink what the default behavior is.

PR #1152, opened 2025-04-12
docs: `prettier-plugin-sh` now uses `sh-syntax`

JounQin:


JounQin:
cc @mvdan 

No idea why CI is broken...

Issue #1149, opened 2025-04-07
pattern: wrong results for `/**foo`


paulhammond:
While working on #1147 I think I noticed that the regexp generated for `/**foo` is incorrect.

https://go.dev/play/p/XXoRQ4EdBwV is some go code that matches `/**foo` against a number of paths; as you can see all three match.

Here's a bash shell in a container testing the same three paths:
 
```
$ echo $BASH_VERSION
5.2.15(1)-release
$ touch /foo
$ mkdir -p /a/b/c
$ touch /a/b/cfoo
$ touch /a/b/c/foo
$ ls /**foo
/foo
```

Bash only matches `/foo`, the pattern package matches all three paths. Is this a bug or am I missing something?

mvdan:
We aim to emulate Bash behavior as closely as possible within reason, so this is likely a bug. The recursive globbing logic is surprisingly difficult to get exactly right. Happy to review a patch, otherwise I'll look at this sometime this week.

mvdan:
This should be fixed now. See the commit adding test cases; this wasn't a bug in the `interp` or `expand` packages. It was only a bug when using the `pattern` package directly.

paulhammond:
Thanks for working on this so quickly! This looks right to me...

PR #1153, opened 2025-04-14
expand: always use pattern.NoGlobStar option

paulhammond:
This is a followup to #1147 and #1149 to slightly simplify the logic and add some tests.

After adding `pattern.NoGlobStar` in #1147 I had a nagging doubt that I should have addded some code to use this new mode in the expand package, depending on whether `cfg.GlobStar` was enabled or not. I spent some time this afternoon checking that assumption, and I can't find a way to get the expand package to use the GlobStar support in the pattern package, even before the fixes you added for #1149 where the behavior was different.

But it's possible I missed something, and I think it's a good idea to disable the unused functionality just in case. Hence this PR.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -963,7 +963,7 @@ func (cfg *Config) glob(base, pat string) ([]string, error) {
 			}
 			continue
 		}
-		mode := pattern.Filenames | pattern.EntireString
+		mode := pattern.Filenames | pattern.EntireString | pattern.NoGlobStar
 		if cfg.NoCaseGlob {
 			mode |= pattern.NoGlobCase
 		}


paulhammond:
I'm a little confused by the Windows test failure, and don't have a Windows system available to debug it on, but it looks like the problem is that I'm using a `fs.FS` here (where "[paths are slash-separated on all systems, even Windows](https://pkg.go.dev/io/fs#ValidPath)") but the rest of the code is converting paths to windows separators before passing them to `ReadDir2()`.

I'm not sure how this should work. Given `ReadDir2` was added to work with `fs` I assumed it followed the patterns of that package, but that was just an assumption. Go itself isn't specific on this in places either (eg: [the docs for os.ReadDir](https://pkg.go.dev/os#ReadDir) don't specify slash style either). And as I said I don't have a windows system to test on.

I'd be OK giving up on this PR, or doing some work (maybe in another PR) to clarify how this should all work. Any thoughts?

mvdan:
Thanks for these issues and PRs, they are really appreciated.

Your thinking of `expand` and `NoGlobStar` is entirely correct. It also made me realise, while staring at the code, that the `pattern` package was entirely wrong in handling leading periods as well. That, once again, was something handled by `expand` on top, but not in `interp` itself. See https://github.com/mvdan/sh/commit/14ccc2e28cfacf2eb12e79742af26af460b5f6e7 and the handful of commits before it.

Could you rebase on the latest master with those changes? Just in case.

I think the change is correct. As for the tests, you're correct that the slash separators are the culprit. In short, we indeed use `os.ReadDir` semantics, which uses `filepath.Separator`, i.e. `/` on "unix" and `\` on windows. That is not compatible with `io/fs`, which always uses forward slashes. For better or worse we designed much of this module API before `io/fs` and https://pkg.go.dev/io/fs#ReadDirFS.

The expand docs do at least hint towards this, in the last godoc line:

```
	// ReadDir2 is used for file path globbing.
	// If nil, and [ReadDir] is nil as well, globbing is disabled.
	// Use [os.ReadDir] to use the filesystem directly.
	ReadDir2 func(string) ([]fs.DirEntry, error)
```

However, given that AFAIK `os.ReadDir` supports forward slashes on Windows too, perhaps we could wrap `expand.Config.ReadDir2` calls so that the given path string is always run through https://pkg.go.dev/path/filepath#ToSlash. This would mean that we would be compatible with `io/fs` for reading directories, which I think is useful, without losing native Windows compatibility with `os.ReadDir`. (Note that https://github.com/golang/go/issues/44279 would still be a problem, but that's not something we can easily resolve ourselves)

mvdan:
In short:
* Your fix is correct. We can do that in this PR without a test for now, and merge it.
* I can look at making `expand.Config.ReadDir2` compatible with `io/fs` via `filepath.ToSlash`.
* When that is done, we can then add your tests which currently fail on Windows.

mvdan:
It's also worth noting that the `expand` package is light on tests because the `interp` package already includes tons of integration tests, which cover nearly every edge case of what `expand` does. Including globstar.

paulhammond:
Thank you for all the hard work you do on sh (and many other go packages) and for taking the time to work on these changes!

As you suggested I've updated my commit to remove the test changes, and rebased against latest master. I think the first test run hit https://github.com/golang/go/issues/22315 so I pushed again to restart the tests which now seem to be passing.

mvdan:
Yeah, those ETXTBSY errors are annoying. Not all that much we can do other than add retries at some level.

PR #1154, opened 2025-04-16
Revert `C.UTF-8` on tests to fix tests on macOS

andreynering:
Revert "all: assume C.UTF-8 is available in the tests"

This reverts commit dfc1441b9cd87d9430893024743aa2197a8e3e8f.

---

That commits assumes `C.UTF-8` is available everywhere, but it is not really available on macOS, making several tests fail because this is prefixed on the expected outputs:

```
bash: warning: setlocale: LC_ALL: cannot change locale (C.UTF-8): Bad file descriptor
```

Also, FYI `TestRunnerRunConfirm` is still failing on macOS after this change. I may take a look soon.

```
ok  	mvdan.cc/sh/v3/cmd/gosh	(cached)
ok  	mvdan.cc/sh/v3/cmd/shfmt	(cached)
ok  	mvdan.cc/sh/v3/expand	(cached)
ok  	mvdan.cc/sh/v3/fileutil	(cached)
--- FAIL: TestRunnerRunConfirm (0.02s)
    --- FAIL: TestRunnerRunConfirm/#433 (0.01s)
        interp_test.go:4012: wrong bash output in "echo foo >&- 2>&-; :":
            want: ""
            got:  "bash: line 1: echo: write error: Bad file descriptor\n"
    --- FAIL: TestRunnerRunConfirm/#289 (0.01s)
        interp_test.go:4012: wrong bash output in "HOME='/*'; echo ~; echo \"$HOME\"":
            want: "/*\n/*\n"
            got:  "/Users/andrey\n/*\n"
    --- FAIL: TestRunnerRunConfirm/#948 (0.03s)
        interp_test.go:4012: wrong bash output in "HOME=/foo; echo ~ ~/ ~/'' ~'' ~\"\"":
            want: "/foo /foo/ /foo/ ~ ~\n"
            got:  "/Users/andrey /Users/andrey/ /Users/andrey/ ~ ~\n"
    --- FAIL: TestRunnerRunConfirm/#946 (0.02s)
        interp_test.go:4012: wrong bash output in "HOME=/foo; rel=/bar; echo ~/bar ~/'bar' ~/\"bar\" ~/$rel ~/\"$rel\"":
            want: "/foo/bar /foo/bar /foo/bar /foo//bar /foo//bar\n"
            got:  "/Users/andrey/bar /Users/andrey/bar /Users/andrey/bar /Users/andrey//bar /Users/andrey//bar\n"
FAIL
FAIL	mvdan.cc/sh/v3/interp	2.405s
ok  	mvdan.cc/sh/v3/pattern	(cached)
ok  	mvdan.cc/sh/v3/shell	(cached)
ok  	mvdan.cc/sh/v3/syntax	2.628s
ok  	mvdan.cc/sh/v3/syntax/typedjson	(cached)
FAIL
```

### DIFF
--- a/syntax/parser_test.go
+++ b/syntax/parser_test.go
@@ -157,10 +157,8 @@ func TestParseBats(t *testing.T) {
 
 func TestMain(m *testing.M) {
 	// Set the locale to computer-friendly English and UTF-8.
-	// Some systems like Arch miss C.UTF8, so fall back to the US English locale.
-	//
-	// TODO: glibc 2.35 was released with C.UTF-8 in February 2022.
-	// This locale is now becoming a standard, so remove the workaround in 2023.
+	// Some systems like macOS miss C.UTF8, so fall back to the US English
+	// locale.
 	if out, _ := exec.Command("locale", "-a").Output(); strings.Contains(
 		strings.ToLower(string(out)), "c.utf",
 	) {
--- a/syntax/parser_test.go
+++ b/syntax/parser_test.go
@@ -156,10 +156,20 @@ func TestParseBats(t *testing.T) {
 }
 
 func TestMain(m *testing.M) {
-	// Set the locale to computer-friendly English and UTF-8, C.UTF-8,
-	// which started shipping with glibc 2.35 in February 2022.
-	os.Setenv("LANGUAGE", "C.UTF-8")
-	os.Setenv("LC_ALL", "C.UTF-8")
+	// Set the locale to computer-friendly English and UTF-8.
+	// Some systems like Arch miss C.UTF8, so fall back to the US English locale.
+	//
+	// TODO: glibc 2.35 was released with C.UTF-8 in February 2022.
+	// This locale is now becoming a standard, so remove the workaround in 2023.
+	if out, _ := exec.Command("locale", "-a").Output(); strings.Contains(
+		strings.ToLower(string(out)), "c.utf",
+	) {
+		os.Setenv("LANGUAGE", "C.UTF-8")
+		os.Setenv("LC_ALL", "C.UTF-8")
+	} else {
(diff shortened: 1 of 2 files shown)


syntax/parser_test.go
```diff
@@ -156,10 +156,20 @@ func TestParseBats(t *testing.T) {
...
 func TestMain(m *testing.M) {
-	// Set the locale to computer-friendly English and UTF-8, C.UTF-8,
-	// which started shipping with glibc 2.35 in February 2022.
-	os.Setenv("LANGUAGE", "C.UTF-8")
-	os.Setenv("LC_ALL", "C.UTF-8")
+	// Set the locale to computer-friendly English and UTF-8.
+	// Some systems like Arch miss C.UTF8, so fall back to the US English locale.
+	//
+	// TODO: glibc 2.35 was released with C.UTF-8 in February 2022.
+	// This locale is now becoming a standard, so remove the workaround in 2023.
```
mvdan, line 163: this TODO is wrong now then, as we're in 2025. it should probably mention that macOS still does not ship the C locale.

andreynering: I just updated the comment to mention macOS.

andreynering:
@mvdan I'm on macOS Sequoia:

Screenshot 2025-04-22 at 11 52 24

It seems that the reason on why CI passes without this is that we're only running `TestRunnerRunConfirm` tests on Ubuntu:

https://github.com/mvdan/sh/blob/1aa0626b04a999460c22ea417a4c421437ad8daa/.github/workflows/test.yml#L23-L27

mvdan:
Ah of course. I wasn't thinking that these are the "confirm" tets with the real Bash.

mvdan (approved):

PR #1157, opened 2025-04-17
fix(interp): allow multiple options on `declare` (`-Ag`)

andreynering:
We were getting the following error on a given script:

    declare: invalid option "-Ag"


mvdan:
Thanks - the change looks fine, but it needs tests.

andreynering:
Test added

mvdan (approved):
Thanks!

PR #1159, opened 2025-04-22
pattern: allow * to match non-leading dots

paulhammond:
I was checking your recent changes to how pattern/expand support `*` and `.`, and I think there's a mistake in commit ce872e7abef6d1be48efbc0ec290a703c23ea418. The rules that prevent `*` from matching dots in filenames only apply to the first character. In other words "foo*" should match "foo.suffix". This PR fixes that bug.


mvdan:
Dang, well spotted.

mvdan:
I think the same fix needs to happen for `foo**`, as it effectively behaves like `foo*`. We might need to refactor the code a bit to allow reusing some of these regex pieces.

mvdan (approved):

mvdan:
No need, I'll be sending some changes to improve the tests and refactor the code a bit.

mvdan:
I'm done with those changes as of https://github.com/mvdan/sh/commit/93f7ac212390b77f75b360440e35a22a9a09dc48. There is no duplication for handling wildcards now, and the highest amount of indentation in that code is down from 7 tabs to 5. The lexer added a bit of verbosity to the code, but I thought it would make the translator a bit more principled and maintainable in the long run.

paulhammond:
Looks good! Thanks again for all the hard work here

PR #1167, opened 2025-06-11
README.md: add another example to caveats

kolyshkin:
```sh
$ echo '${array[weird!key]}' | shfmt
<standard input>:1:8: reached ! without matching [ with ]
```

I was not aware of this limitation but saw it in README just before filing a bug. Let's add another example, mostly for the sake of discoverability.

While at it, add <standard input> to an existing example, as this is how the output looks like now (using v3.11).


mvdan (approved):
Thanks!

PR #1170, opened 2025-06-25
syntax: add balanceCase for balancing case parentheses

### LINKED ISSUE #1168: syntax: add option for balancing case parentheses
shfmt currently always removes the optional left parenthesis on `case`, which is part of the base POSIX sh language. Although nobody likes to type it on the command-line, the balanced form of case does have the benefit of, well, being balanced and not confusing generic parenthesis-highlight programs. I want to use it.

Artoria2e5:
This should fix #1168

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -60,6 +60,7 @@ var (
 
 	// Printer flags.
 	indent      = &multiFlag[uint]{"i", "indent", 0}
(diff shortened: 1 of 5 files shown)


mvdan:
I appreciate that you sent a patch for this, but see my response to the issue you raised a few days ago - I don't think we should add an option for this.

Issue #1176, opened 2025-07-11
Allow `${|stmts;}` and `{ stmts;}` in Bash


xPMo:
Bash 5.3 now supports the mksh `${ stmts;}` and `${|stmts;}` syntaxes.

```bash
#!/bin/bash
# test.sh
foo(){
    REPLY=x
}
echo "${|foo;}"
```
```
$ shfmt test.sh
test.sh:6:7: "${|stmts;}" is a mksh feature; tried parsing as bash (parsed as bash via -ln=auto)
```

```bash
#!/bin/bash
# test2.sh
foo(){
    echo x
}
echo "${ foo;}"
```
```bash
$ shfmt test2.sh
test2.sh:6:7: "${ stmts;}" is a mksh feature; tried parsing as bash (parsed as bash via -ln=auto)
```

jansorg:
https://www.gnu.org/software/bash/manual/bash.html#Command-Substitution mentions that \t and \n are valid as first char too (tab and newline).

> The character c following the open brace must be a space, tab, newline, or ‘|’, and the close brace must be in a position where a reserved word may appear (i.e., preceded by a command terminator such as semicolon). Bash allows the close brace to be joined to the remaining characters in the word without being followed by a shell metacharacter as a reserved word would usually require.

xPMo:
That's exactly what shfmt is already checking for when parsing for mksh:

https://github.com/mvdan/sh/blob/078753e5fd9efcb5139237b38f3b53d2f887d788/syntax/parser.go#L1065

amas127:
```bash
$ shfmt --version
v3.13.1

$ echo 'HI=${ foo; }' | shfmt -ln bash
HI=${ foo; }

$ echo 'HI="${ foo; }"' | shfmt -ln bash
<standard input>:1:14: reached EOF without closing quote `"`
```

Hi, guys! Sorry for reporting in a closed issue in advance. `foo="${ stmts;}"` is not supported while `foo=${ stmts;}` is supported.

PR #1178, opened 2025-07-24
syntax: support alternate command substitution of Bash 5.3

jansorg:
Closes https://github.com/mvdan/sh/issues/1176

Bash 5.3 introduced support for ${ command; } , ${| command; } , ${\n command; } and ${\t command; }

ToDo:
- [ ]  Bash 5.3 supports parsing of `prefix${ echo hello world; }suffix`, but shfmt does not currently support parsing the closing `}` without a terminating characters after it.
  Bash website says:
  > Bash allows the close brace to be joined to the remaining characters in the word without being followed by a shell metacharacter as a reserved word would usually require.

  In my tests, `mksh` works like Bash 5.3, too. It's correctly handling, but shfmt currently does not support the `prefix${ echo hello world; }suffix` syntax.

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -1951,29 +1951,29 @@ var fileTests = []testCase{
 		)))),
 	},
 	{
-		Strs: []string{"${ foo;}", "${\n\tfoo; }", "${\tfoo;}"},
-		mksh: &CmdSubst{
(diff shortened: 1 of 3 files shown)


mvdan:
This would need the positive tests for these constructs to apply for bash too. The lines in filetests_test.go like 1969 can be changed from `mksh` to `bsmk` (old me trying to abbreviate "bash and mksh").

As for the word continuations, I'd try to tackle that as a follow-up change, because it may be more involved.

PR #1179, opened 2025-07-24
moreinterp/coreutils: introduce exec middleware with first set of core utils

andreynering:


moreinterp/go.mod
```diff
@@ -1,3 +1,17 @@
 module mvdan.cc/sh/moreinterp
 
-go 1.23
+go 1.24.0
+
+toolchain go1.24.4
+
+require (
+	github.com/u-root/u-root v0.14.1-0.20250722142936-bf4e78a90dfc
```
andreynering, line 8: I'm importing from this branch for now because it has some important fixed:

* https://github.com/u-root/u-root/pull/3421

moreinterp/coreutils/coreutils.go
```diff
@@ -1 +1,70 @@
...
+	"find":  func() core.Command { return find.New() },
+	"ls":    func() core.Command { return ls.New() },
+	"mkdir": func() core.Command { return mkdir.New() },
+	"mv":    func() core.Command { return mv.New() },
+	"rm":    func() core.Command { return rm.New() },
+	"touch": func() core.Command { return touch.New() },
+	"xargs": func() core.Command { return xargs.New() },
+}
+
+// ExecHandler returns an ExecHandlerFunc middleware that handles core utils
```
mvdan, line 40:
```suggestion
// ExecHandler returns an [interp.ExecHandlerFunc] middleware that handles core utils
```

moreinterp/coreutils/coreutils.go
```diff
@@ -1 +1,70 @@
...
+	"github.com/u-root/u-root/pkg/core/ls"
+	"github.com/u-root/u-root/pkg/core/mkdir"
+	"github.com/u-root/u-root/pkg/core/mv"
+	"github.com/u-root/u-root/pkg/core/rm"
+	"github.com/u-root/u-root/pkg/core/touch"
+	"github.com/u-root/u-root/pkg/core/xargs"
+	"mvdan.cc/sh/v3/interp"
+)
+
+var coreUtils = map[string]func() core.Command{
```
mvdan, line 27: perhaps call this `commands`, given that we're already in a package called coreutils. or even better, to be clearer that these are constructors, something like `commandBuilders`.

moreinterp/coreutils/coreutils.go
```diff
@@ -1 +1,70 @@
...
+// commands.
+//
+// Keep in mind that this middleware has priority over the core utils available
+// by the system. You may want to use only on Windows to ensure that the system
+// core utils are used on other platforms, like macOS and Linux.
+func ExecHandler(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
+	return func(ctx context.Context, args []string) error {
+		if len(args) == 0 {
+			return next(ctx, args)
+		}
```
mvdan, line 50: you can remove this; args will never be empty. I've clarified this here: dd3c54b1703554f3dabc2c003b60d76093784207

moreinterp/coreutils/coreutils_test.go
```diff
@@ -0,0 +1,76 @@
...
+				}
+			case "rm":
+				if err.Error() != "rm [-Rrvif] file..." {
+					t.Errorf("expected %q output, got: %q", tt.cmd, err)
+				}
+			case "xargs":
+				t.Skip()
+			default:
+				t.Errorf("unimplemented test for core util: %q: %q, %q", tt.coreUtil, err, out.String())
+			}
```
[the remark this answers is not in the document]
mvdan: How about testing them all via a bad flag, like `cat --badoption`, `touch --badoption` and so on? It should make the input shell more consistent, and the expected result should presumably be predictable as well, as we wouldn't have to e.g. test what is the real behavior of `ls` and `cat`.

moreinterp/go.mod
```diff
@@ -1,3 +1,16 @@
 module mvdan.cc/sh/moreinterp
 
-go 1.23
+go 1.24.0
+
+toolchain go1.24.4
```
mvdan, line 5: rather surprising and a bit worrying that u-root already requires Go 1.24 or later; as a project we support two Go versions, so ideally I'd like to keep the same for the coreutils module. Perhaps we can ask them to lower this.

at the very least I think we can remove the toolchain line?

mvdan (approved):
SGTM; I'd just use a bad/unknown flag to test the commands, which should be simpler and more consistent.

Also, please add copyright headers to the Go files, like the others.

moreinterp/go.mod
andreynering: Opened https://github.com/u-root/u-root/pull/3428 with the proposal.

In the meantime, I removed the toolchain line.

mvdan (approved):
Thanks!

PR #1189, opened 2025-08-28
fix: output error when changing into a dir that doesn't exist

caarlos0:
Right now it merely exits 1, which IMHO lacks context.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -243,7 +243,7 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		default:
 			return failf(2, "usage: cd [dir]\n")
 		}
-		exit.code = r.changeDir(ctx, path)
+		exit.code = r.changeDir(ctx, "cd", path)
 	case "wait":
 		fp := flagParser{remaining: args}
 		for fp.more() {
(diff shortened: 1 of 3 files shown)


interp/builtin.go
```diff
@@ -984,14 +984,16 @@ func (r *Runner) readLine(ctx context.Context, raw bool) ([]byte, error) {
 	}
 }
 
-func (r *Runner) changeDir(ctx context.Context, path string) uint8 {
+func (r *Runner) changeDir(ctx context.Context, cmd, path string) uint8 {
 	path = cmp.Or(path, ".")
 	path = r.absPath(path)
 	info, err := r.stat(ctx, path)
 	if err != nil || !info.IsDir() {
+		r.errf("%s: the directory '%s' does not exist\n", cmd, path)
```
mvdan, line 992: use `%q` for the path, here and in the line below - that way we get consistently good quoting. it doesn't matter that we're not like bash - the tests already don't expect that because of `#JUSTERR`

interp/interp_test.go
```diff
@@ -3780,7 +3780,7 @@ func TestRunnerRun(t *testing.T) {
 			if err := r.Run(ctx, file); err != nil {
 				cb.WriteString(err.Error())
 			}
-			want := c.want
+			want := strings.ReplaceAll(c.want, "#JUSTDIR", tdir)
```
mvdan, line 3783: rather than adding `#JUSTDIR` to the test logic, what I'd do is report the error based on what the original path was, not the absolute path. that's more helpful to the user in most cases, and given that the tests don't really use absolute paths in these cases, it also avoids adding extra test logic here.

mvdan (approved):
Thanks! SGTM with two nits.

interp/builtin.go
andreynering, line 992: Even we don't need to be 100% compatible, I'd still try to have the same error format as it's easy enough.

```suggestion
		r.errf("%s: no such file or directory: %s\n", cmd, path)
```

andreynering (approved):

mvdan:
I still see CI failures; let me know when you're ready with a final review. Although I'd still argue against adding the JUSTDIR logic.

caarlos0:
regarding JUSTDIR, do you have any suggestions? It seems that `pushd some/dir` is fine, but if you run `pushd` again with no args, it'll be the absolute path

caarlos0:
Seems like build is passing now - except the one job that was canceled for some reason 🙏

mvdan:
With your permission I've done the tweaks outlined above, then I'll merge.

caarlos0:
🎉

PR #1190, opened 2025-08-28
interp: print output error if `cd` fails

andreynering:
`cd` currently just exits with status 1 without printing any output. This can be confusing when using inside `gosh` for example, because it's not clear that an error happened as there is no feedback for non-zero statuses, and not other error was print.

This patch replicates the behavior of Bash, outputing "no such file or directory" or "permission denied".

---

@mvdan I don't know if there is any specific reason on why there was no output at all for failed `cd` calls. Let me know if I'm missing anything.


andreynering:
@mvdan We're coworkers and were looking at the same issue. Lack of communication between us.

No worries, we can keep the other PR.

Issue #1223, opened 2025-11-21
(optionally) have `shfmt` exit nonzero when using `--list`


thejcannon:
Title says most of it!

For [some automation](https://hk.jdx.dev/) I'm using being able to know if files aren't formatted correctly is great, but unfortunately most other tools exit nonzero if there's files that are listed.

There's some prior art as well...
```
# Prettier
$ prettier --list-different bad.json; echo $?
bad.json
1
$ prettier --list-different good.json; echo $?
0

# Rustfmt
$ rustfmt --check --files-with-diff bad.rs; echo $?
/abs/path/to/bad.rs
1
$ rustfmt --check --files-with-diff good.rs; echo $?
0

# OpenTofu
$ tofu fmt --check bad.tf; echo $?
bad.tf
3
$ tofu fmt --check good.rs; echo $?
0
```

mvdan:
Previously: https://github.com/mvdan/sh/issues/1093

That time, I was a bit on the fence about it, mainly as I didn't see precedent with it. But it's also true that I only use gofmt regularly, I wasn't aware of other formatters behaving this way already.

This sounds fine to me on the surface. There's always a small chance that this may break some existing users who rely on the current exit status code behavior, but they could always adjust their scripts accordingly without too much effort, like `shfmt -l . || true`. It's still possible to tell if the command caused an error by checking if there was any stderr output.

PR #1224, opened 2025-11-21
cmd/shfmt: Exit nonzero for `-l` and file unformatted

thejcannon:
Fix: https://github.com/mvdan/sh/issues/1223

This change makes it so `shfmt` errors 1 if there are formatting errors with `-l`/`--list`.


mvdan (approved):
SGTM, we'd just need to adjust the help text and man page as well. for example:

```
  -l[=0], --list[=0]  error with a list of files whose formatting differs from shfmt;
                      paths are separated by a newline or a null character if -l=0
```

PR #1241, opened 2026-01-18
coreutils: add specific error type

andreynering:
We have the need to know if an error was returned by the core utilities. Since the u-root project doesn't have specific error types, that looks like a good fit for the middleware.

Related:

* https://github.com/go-task/task/issues/2466
* https://github.com/go-task/task/pull/2619

### DIFF
--- a/moreinterp/coreutils/error.go
+++ b/moreinterp/coreutils/error.go
@@ -1,0 +1,21 @@
+package coreutils
+
+import "fmt"
+
+// Error wraps any error returned from the core utilities.
+type Error struct {
+	err error
+}
+
+var (
+	_ error                       = &Error{}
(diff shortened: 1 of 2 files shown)


moreinterp/coreutils/error.go
```diff
@@ -0,0 +1,18 @@
...
+type Error struct {
+	err error
+}
+
+// Error implements the error interface.
+func (err *Error) Error() string {
+	return fmt.Sprintf("coreutils: %v", err.err)
+}
+
+// Unwrap implements the unwrap interface.
```
mvdan: I'd replace the doc comments with interface implementation asserts like `var _ error = ...`

Issue #1252, opened 2026-02-13
interp: bash compatibility gaps affecting Gentoo ebuild interpretation


kolkov:
Hi Daniel,

As mentioned in #1239, we're using `mvdan.cc/sh` as the bash interpreter in [GRPM](https://github.com/grpmsoft/grpm), our Go-based package manager for Gentoo Linux. Since then the project has grown to 70K+ LOC with 2K+ tests, and we've been running it against the full Gentoo package tree (~31K packages).

Along the way we've hit several interpreter gaps that we've been [working around in our codebase](https://github.com/grpmsoft/grpm/blob/main/internal/ebuild/executor.go) (text preprocessing, function body stripping, `BASH_VERSINFO` manipulation, etc.), but we've reached a point where we need to decide our interpreter strategy: continue investing in `mvdan.cc/sh`, or build a custom interpreter tailored for ebuild semantics.

Before making that decision, we'd like to understand which of the gaps below are realistic to address upstream and which are out of scope by design. We're not expecting all of these to be fixed — just looking for your perspective so we can plan accordingly.

### 1. `declare -f` / `declare -p` return error instead of function/variable info

```bash
$ bash -c 'f() { :; }; declare -f f >/dev/null && echo exists'
exists

$ gosh -c 'f() { :; }; declare -f f >/dev/null && echo exists'
declare: invalid option "-f"
```

**Impact:** Widely used across Gentoo eclasses (30+ occurrences). `distutils-r1.eclass` alone uses `declare -f func_name` 11 times to check whether phase functions were defined by the ebuild. `declare -p VAR` is used to inspect variable attributes (e.g., check if array).

**Source:** `interp/runner.go` DeclClause flag parser only accepts `-x`, `-r`, `-a`, `-A`, `-n`, `-g`.

### 2. `${var@a}` and `${var@A}` panic instead of returning attributes

```bash
$ bash -c 'declare -a arr=(1 2 3); echo "${arr@a}"'
a

$ gosh -c 'declare -a arr=(1 2 3); echo "${arr@a}"'
panic: unhandled @a param expansion
```

`@Q` and `@E` work correctly. `@a`, `@A`, and `@P` trigger `panic()` in `expand/param.go:337`.

[...]

PR #1254, opened 2026-02-14
expand: fix formatInto panic and unset var field count

kolkov:
Fix two bugs in `expand/expand.go`, both found while running the interp against Gentoo ebuilds (ref #1252).

### 1. formatInto panic on truncated escape sequences

`readDigits` in `formatInto` reads past the end of the format string when an escape like `\x`, `\u`, or `\U` appears at the end with no hex digits following. A trailing backslash also causes an out-of-bounds access.

```
$ printf '\0'   # OK
$ printf '\x'   # panic: index out of range
$ printf 'a\0'  # panic: index out of range
```

Fix: add `i+j < len(format)` bound in `readDigits` loop, and guard the trailing backslash case.

### 2. Unset var `[@]` produces wrong field count

`"${unset_var[@]}"` produces one empty field instead of zero fields:

```bash
count() { echo $#; }
count "${unset_var[@]}"
# expected: 0
# got: 1
```

Fix: handle the `Unknown` + unset case in `quotedElemFields` to return an empty slice.

Both fixes include tests.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -792,11 +792,18 @@ func (cfg *Config) quotedElemFields(pe *syntax.ParamExp) []string {
 	}
 	switch nodeLit(pe.Index) {
 	case "@": // "${name[@]}"
-		switch vr := cfg.Env.Get(name); vr.Kind {
+		vr := cfg.Env.Get(name)
+		switch vr.Kind {
 		case Indexed:
 			return vr.List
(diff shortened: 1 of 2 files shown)


kolkov:
Fixed, `printf '\x'` test now uses `#IGNORE` as bash prints a warning to stderr that gosh doesn't emit.

PR #1255, opened 2026-02-14
expand,interp: implement ${var@a}/@A/@P, declare -f/-p, type -P, read -a

kolkov:
Implement six unimplemented features from #1252 (items 1-4 and 6), split into four commits.

### Commit 1: `${var@a}`, `${var@A}`, `${var@P}`

Adds a `Variable.Flags` method that returns attribute flags in bash order (a/A/n, r, x), then uses it to implement the three remaining `@` operators:

- `${var@a}` returns attribute flags
- `${var@A}` returns a declare statement that recreates the variable
- `${var@P}` returns the value as-is (prompt expansion not yet supported)

Previously all three triggered a panic.

### Commit 2: `declare -f` and `declare -p`

- `declare -f name` prints the function definition, or silently returns exit 1 if not found
- `declare -p name` prints `declare -FLAGS name="value"` matching bash output

Common eclass pattern: `declare -f funcname >/dev/null`.

### Commit 3: `type -P`

`type -P` forces a PATH lookup, sharing the existing `-p` code path. Two-line change.

### Commit 4: `read -a`

`read -a` splits input by IFS into an indexed array. Supports combined flags (`read -ra`). Empty input produces an empty array.

All commits include tests. All tests verified against bash 5.2.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -665,6 +665,7 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		var prompt string
 		raw := false
 		silent := false
+		readArray := false
 		fp := flagParser{remaining: args}
 		for fp.more() {
 			switch flag := fp.flag(); flag {
@@ -672,6 +673,8 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 				silent = true
 			case "-r":
 				raw = true
+			case "-a":
+				readArray = true
 			case "-p":
 				prompt = fp.value()
 				if prompt == "" {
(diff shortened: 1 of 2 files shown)


expand/param.go
```diff
@@ -334,8 +334,27 @@ func (cfg *Config) paramExp(pe *syntax.ParamExp) (string, error) {
...
+				// We use orig (before nameref resolve) for the attributes.
+				str = orig.Flags()
+			case "A":
+				// ${var@A} returns a declare statement that recreates the variable.
+				// Plain vars with no flags: name='value' (no declare prefix).
+				// Vars with flags: declare -FLAGS name='value'.
+				flags := orig.Flags()
+				// Bash @A always wraps values in single quotes,
+				// escaping embedded single quotes as '\''.
+				quoted := "'" + strings.ReplaceAll(str, "'", `'\''`) + "'"
```
mvdan: we have a `syntax.Quote` API just for this purpose; use it.

moreover, this manual quoting you did here is incorrect; single-quoted strings in shell do not support escaping at all.

expand/param.go
```diff
@@ -334,8 +334,27 @@ func (cfg *Config) paramExp(pe *syntax.ParamExp) (string, error) {
...
+				if flags == "" {
+					str = name + "=" + quoted
+				} else {
+					str = "declare -" + flags + " " + name + "=" + quoted
+				}
+			case "P":
+				// ${var@P} performs prompt expansion.
+				// Minimal implementation: return the value as-is.
+				// Full prompt expansion (\u, \h, \w, etc.) is not yet supported.
```
mvdan: rewrite this last line as a `// TODO: ...`

kolkov:
Thanks for the review! Applied all suggestions:

- `syntax.Quote` instead of manual quoting, with `fmt.Sprintf` for string building
- `${var@P}` comment rewritten as `// TODO`
- All commit messages now use `Updates #1252`

Regarding splitting into separate PRs: commits 1 and 2 share `Variable.Flags()` (added in commit 1, used by `declare -p` in commit 2), so they need to land together. I kept type -P and read -a in the same PR since they're small and all six features were grouped together in #1252, but happy to split them out if you prefer.

interp/runner.go
```diff
@@ -709,6 +712,59 @@ func (r *Runner) cmd(ctx context.Context, cm syntax.Command) {
...
+					r.outf("%s\n", buf.String())
+				} else {
+					r.exit.code = 1
+				}
+				continue
+			}
+			if declQuery == "-p" {
+				// declare -p name: print variable with attributes.
+				vr := r.lookupVar(name)
+				if !vr.IsSet() && !vr.Declared() {
```
mvdan: checking both seems wrong; when `vr.Declared()` is false, `vr.IsSet()` must also be false.

PR #1256, opened 2026-02-15
interp,pattern: implement !(pattern) extglob negation

kolkov:
Implement `!(pattern-list)` extglob negation for `case` and `[[ ]]` pattern matching, as suggested in #1252 (item 7).

Go's `regexp` package does not support negative lookahead, so `!(pattern)` cannot be compiled into a single regex. Instead, `pattern.Regexp` now returns a `NegExtglobError` sentinel, and the interpreter handles negation at the caller level:

- Simple case (`!(inner)` is the entire pattern): compile the inner pattern and negate `MatchString`
- Mixed case (`prefix!(inner)suffix`): try all split points of the name string where prefix and suffix match, and check that the middle does NOT match the inner pattern

Standalone `!` without `(` is now correctly treated as a literal character, fixing a minor regression where `!` in non-extglob context would error.

Pathname expansion (globbing) still returns the existing error for `!(...)` patterns, as glob matching requires different handling with `Filenames` mode and directory scanning.

Only a single `!(...)` group per pattern is supported; concatenated groups like `!(a)!(b)` are not. The empty pattern `!()` is not supported either, as the parser treats `)` as the end of the case pattern rather than the extglob closing paren.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3151,6 +3151,64 @@ done <<< 2`,
 		// Bash would print "a123z a12z az\n"
 		"extglob operator !(: Go's regexp package does not support negative lookahead\n #IGNORE",
 	},
+	// !(pattern) extglob negation in case and [[ ]] matching
+	{
+		"shopt -s extglob\ncase \"bar\" in !(foo)) echo match;; esac",
+		"match\n",
+	},
+	{
+		"shopt -s extglob\ncase \"foo\" in !(foo)) echo match;; esac",
+		"",
+	},
+	{
+		"shopt -s extglob\ncase \"\" in !(foo)) echo match;; esac",
+		"match\n",
+	},
+	{
+		"shopt -s extglob\ncase \"baz\" in !(foo|bar)) echo match;; esac",
+		"match\n",
+	},
+	{
+		"shopt -s extglob\ncase \"file.tar.gz\" in !(*.sig)) echo match;; esac",
+		"match\n",
+	},
+	{
+		"shopt -s extglob\ncase \"file.sig\" in !(*.sig)) echo match;; esac",
+		"",
+	},
+	{
+		"shopt -s extglob\ncase \"foo_xxx_baz\" in foo_!(bar)_baz) echo match;; esac",
+		"match\n",
+	},
+	{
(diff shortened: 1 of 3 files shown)


interp/interp_test.go
```diff
@@ -3151,6 +3151,47 @@ done <<< 2`,
...
+		"shopt -s extglob\ncase \"file.sig\" in !(*.sig)) echo match;; *) echo no;; esac",
+		"no\n",
+	},
+	{
+		"shopt -s extglob\ncase \"foo_xxx_baz\" in foo_!(bar)_baz) echo match;; *) echo no;; esac",
+		"match\n",
+	},
+	{
+		"shopt -s extglob\ncase \"foo_bar_baz\" in foo_!(bar)_baz) echo match;; *) echo no;; esac",
+		"no\n",
```
mvdan: remove all default cases with "echo no"; the output being empty is clear enough. and this way we also keep the inputs shorter.

interp/runner.go
```diff
@@ -869,12 +869,133 @@ func match(pat, name string) bool {
...
+	prefixExpr = strings.TrimPrefix(prefixExpr, "(?s)")
+	suffixExpr = strings.TrimPrefix(suffixExpr, "(?s)")
+	prefixRx := regexp.MustCompile("(?s)^" + prefixExpr + "$")
+	suffixRx := regexp.MustCompile("(?s)^" + suffixExpr + "$")
+	// Try all possible prefix lengths, then all suffix splits within each.
+	for prefixEnd := 0; prefixEnd <= len(name); prefixEnd++ {
+		if !prefixRx.MatchString(name[:prefixEnd]) {
+			continue
+		}
+		for mid := prefixEnd; mid <= len(name); mid++ {
```
mvdan: this double loop is rather worrying. this is exactly the kind of slowness that we should be very careful about.

some quick googling around shows me that the majority of uses of `!(expr)` are rather simple, for example without any prefix nor suffix, or with either a prefix or suffix without any pattern meta-characters, i.e. just fixed strings. All the test cases that you added conform to this limited form too.

So I would implement that as a first step, which should not require any looping at all:
* match the fixed prefix, if any
* match the fixed suffix, if any
* negate-match a single `!(expr)` in the middle

PR #1258, opened 2026-02-16
expand: apply array slicing in quoted parameter expansions

### LINKED ISSUE #1239: interp: `local -a` without assignment doesn't create indexed array
## Summary

When using `local -a varname` (or `declare -a varname`) without an assignment, the variable is not properly initialized as an indexed array. Subsequent array operations like `+=` or index assignment don't work as expected.

## Reproduction

```bash
f() {
    local -a arr
    arr+=("a" "b")
    echo "${arr[@]}"
}
f
```

**Expected output:** `a b`
**Actual output:** (empty or incorrect)

Similarly:

```bash
f() {
    declare -a arr
    arr[0]="x"
    arr[1]="y"
    echo "${arr[@]}"
}
f
```

**Expected output:** `x y`
**Actual output:** (empty or incorrect)

## Analysis

Looking at `interp/runner.go` around line 711-720, when processing a naked declaration with `-a` flag:

```go
if as.Naked {
    if valType == "-A" {
        vr.Kind = expand.Associative
    } else {
        vr.Kind = expand.KeepValue
    }
}
```

The code correctly handles `-A` (associative array) by setting `vr.Kind = expand.Associative`, but `-a` (indexed array) falls through to the `else` branch which sets `vr.Kind = expand.KeepValue` instead of `expand.Indexed`.

## Expected Fix

```go
if as.Naked {
    switch valType {
    case "-A":
        vr.Kind = expand.Associative
    case "-a":
        vr.Kind = expand.Indexed
    default:
        vr.Kind = expand.KeepValue
    }
}
```

## Real-World Impact

This bug affects Gentoo's `ver_cut` function from `/usr/lib/portage/bin/version-functions.sh`, which uses `local -a comp` to parse version strings. We discovered this while implementing a Go-based package manager that uses mvdan.cc/sh for ebuild evaluation.

## Environment

- mvdan.cc/sh version: v3.12.0 (latest)
- Go version: 1.25.4

kolkov:
When `"${a[@]:offset:length}"` was the sole content of a double-quoted word, `quotedElemFields` returned the full array without applying the slice. This happened because the fast path for quoted array expansions did not account for `pe.Slice`.

### Reproduction

```bash
$ bash -c 'a=(1 2 3 4 5); echo "${a[@]:2:2}"'
3 4

$ gosh -c 'a=(1 2 3 4 5); echo "${a[@]:2:2}"'
1 2 3 4 5
```

### Fix

Add a `sliceElems` helper that applies offset and length to the element list, and call it from the `[@]` and `[*]` return points in `quotedElemFields`.

### Tests

6 new test cases covering `[@]`, `[*]`, array assignment, offset-only, negative offset, and negative-beyond-bounds.

Fixes #1239.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -796,7 +796,7 @@ func (cfg *Config) quotedElemFields(pe *syntax.ParamExp) []string {
 		vr := cfg.Env.Get(name)
 		switch vr.Kind {
 		case Indexed:
-			return vr.List
+			return cfg.sliceElems(pe, vr.List)
 		case Associative:
 			return slices.Collect(maps.Values(vr.Map))
 		case Unknown:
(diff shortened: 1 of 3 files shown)


kolkov:
Note: while working on this fix I noticed a pre-existing issue with positional parameter slicing. `${@:N}` uses 1-based offsets in bash (and offset 0 includes `$0`), but gosh treats the offset as 0-based:

```bash
$ bash -c 'f() { echo "${@:2:2}"; }; f a b c d e'
b c

$ gosh -c 'f() { echo "${@:2:2}"; }; f a b c d e'
c d
```

The root cause is in `paramExp`: `vr.List` is set to `r.Params` which stores `$1`+ as a 0-indexed Go slice, so `elems[2:]` gives `$3` instead of `$2`. This affects both quoted and unquoted paths.

I intentionally kept this PR scoped to indexed arrays only. Want me to file a separate issue for the positional parameter case?

PR #1259, opened 2026-02-16
expand: fix positional parameter slicing offsets

kolkov:
## Summary

Two bugs in positional parameter slicing (`$@` and `$*`):

1. **Quoted `"${@:N:M}"` as sole double-quoted content** — `quotedElemFields` returned all positional parameters, ignoring the slice entirely.

2. **0-based instead of 1-based offsets** — In bash, `${@:N}` uses 1-based offsets and offset 0 includes `$0`. The interpreter used 0-based offsets because `r.Params` stores `$1+` as a 0-indexed Go slice without `$0`.

## Fix

Prepend `$0` to the element list before calling `sliceElems` (from #1258) at each call site that handles positional parameters. This makes standard 0-based Go slicing produce correct 1-based results. No separate helper function — reuses `sliceElems` directly.

## Changes

- `expand/expand.go` — `quotedElemFields`: handle `"${*:N:M}"` and `"${@:N:M}"` by prepending `$0` and calling `sliceElems`
- `expand/param.go` — `paramExp`: prepend `$0` before `sliceElems` call when `name` is `@` or `*`
- `interp/interp_test.go` — 12 test cases covering both code paths

Updates #1252.

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -786,17 +786,17 @@ func (cfg *Config) quotedElemFields(pe *syntax.ParamExp) []string {
 		return nil
 	}
 	switch name {
-	case "*": // "${*}"
-		return []string{cfg.ifsJoin(cfg.Env.Get(name).List)}
-	case "@": // "${@}"
-		return cfg.Env.Get(name).List
+	case "*": // "${*}" or "${*:offset:length}"
+		return []string{cfg.ifsJoin(cfg.sliceElems(pe, cfg.Env.Get(name).List, true))}
+	case "@": // "${@}" or "${@:offset:length}"
+		return cfg.sliceElems(pe, cfg.Env.Get(name).List, true)
 	}
 	switch nodeLit(pe.Index) {
 	case "@": // "${name[@]}"
 		vr := cfg.Env.Get(name)
 		switch vr.Kind {
 		case Indexed:
-			return cfg.sliceElems(pe, vr.List)
+			return cfg.sliceElems(pe, vr.List, false)
 		case Associative:
 			return slices.Collect(maps.Values(vr.Map))
 		case Unknown:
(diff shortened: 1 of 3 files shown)


expand/expand.go
```diff
@@ -814,6 +822,40 @@ func (cfg *Config) quotedElemFields(pe *syntax.ParamExp) []string {
...
+		}
+		elems = elems[slicePos(offset):]
+	}
+	if pe.Slice.Length != nil {
+		length, err := Arithm(cfg, pe.Slice.Length)
+		if err != nil {
+			return elems
+		}
+		elems = elems[:slicePos(length)]
+	}
```
mvdan: this duplicates the logic we already have elsewhere to handle parameter expansion slicing. can we reuse the code instead?

kolkov: Sure! If I understand correctly, `sliceElems` from #1258 already handles the element slicing logic. I can rebase this PR on #1258 and reuse it — `slicePositionalParams` would then just prepend `$0` and call `sliceElems`:

```go
case "@":
    elems := cfg.Env.Get(name).List
    if pe.Slice != nil {
        elems = cfg.sliceElems(pe, append([]string{cfg.Env.Get("0").Str}, elems...))
    }
    return elems
```

Would you prefer to merge #1258 first so I can rebase on it, or should I take a different approach?

expand/expand.go
```diff
@@ -786,10 +786,18 @@ func (cfg *Config) quotedElemFields(pe *syntax.ParamExp) []string {
...
 	}
 	switch name {
-	case "*": // "${*}"
-		return []string{cfg.ifsJoin(cfg.Env.Get(name).List)}
-	case "@": // "${@}"
-		return cfg.Env.Get(name).List
+	case "*": // "${*}" or "${*:offset:length}"
+		elems := cfg.Env.Get(name).List
+		if pe.Slice != nil {
+			elems = cfg.sliceElems(pe, append([]string{cfg.Env.Get("0").Str}, elems...))
```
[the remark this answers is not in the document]
kolkov: You're right that centralizing this in `sliceElems` is cleaner. I'll add a `positional bool` parameter.

One clarification on the semantics: `$0` isn't being thrown away — it's a real element that bash returns in the output when the offset reaches it:

```
$ set -- a b c; echo "${@:0:2}"
bash a
$ set -- a b c; echo "${@: -4}"
bash a b c
```

Bash models positional parameters as `[$0, $1, ..., $N]` for slicing purposes, and negative offsets count from `$# + 1` to include $0. Prepending $0 constructs this virtual array so the existing zero-based slicing math works for all cases.

I'll move this into `sliceElems` so `Get("0")` appears in one place, and call sites just pass the boolean. Does that match what you had in mind?

mvdan (approved):

PR #1281, opened 2026-03-08
shfmt: insert space after Not

### LINKED ISSUE #987: syntax: printer breaks certain arithmetic expressions
By default, shfmt removes brackets within parentheses, but this breaks certain kinds of expressions.

For example:

```bash
(( ! 0 )) && echo true || echo false # true
```
shfmt will format this into:

```bash
((!0)) && echo true || echo false # bash: !0: event not found
```

kristiandueholm:
**Purpose**

Attempts to solve #987 regarding history expansion triggered by missing whitespace after `!`.

**Changes**

I simply added a call to the printer's space function when the unary operator is a Not.

**Questions**

* Should I add a test of some sort?
* Should I add a check for compact, and omit the space if true? It "breaks" if the space is omitted, but I really am not sure if compact is considered "dangerous".

Thanks!

Closes #987

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -3088,7 +3088,7 @@ var fileTests = []fileTestCase{
 		langFile(arithmExp(&UnaryArithm{Op: Dec, X: litWord("i")})),
 	),
 	fileTest(
-		[]string{`$((!i))`},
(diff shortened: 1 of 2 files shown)


kristiandueholm:
I should have run tests before pushing. My bad.

PR #1313, opened 2026-04-02
Add  KillProcessGroup  option to handle process groups

mkaaad:
Reverted the previous change that unconditionally killed the entire process group (main.go and main_test.go were not reverted).
 
To prevent interactive commands from failing, I add an explicit switch  KillProcessGroup  to the Runner. It defaults to  false  to keep the original behavior. Set it to  true  when user want to ensure all child processes are terminated.
 
This maintains backward compatibility while solving the orphaned process problem.
```go
// ensure all child processes are terminated
runner.New(interp.KillProcessGroup(true))
```

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -362,7 +362,8 @@ func Interactive(enabled bool) RunnerOption {
 // process group via setpgid, and interrupt/kill signals are sent to the
 // group.
 //
-// Defaults to false. Note that this option only has an effect on Unix systems.
+// Defaults to false. On Windows, this uses CREATE_NEW_PROCESS_GROUP and
+// GenerateConsoleCtrlEvent to signal the process group.
 func KillProcessGroup(enabled bool) RunnerOption {
 	return func(r *Runner) error {
 		r.killProcessGroup = enabled
--- a/interp/api.go
+++ b/interp/api.go
@@ -126,6 +126,10 @@ type Runner struct {
 	// such as the condition in a [syntax.IfClause].
 	noErrExit bool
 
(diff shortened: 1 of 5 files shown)


mvdan:
Also cc @andreynering for thoughts on whether an option like this makes sense. My intuition is that an option here is unintuitive and should be avoided, but if we can't get reasonable behavior for all users without two modes, perhaps it's the best we can do.

mkaaad:
I'm actually trying to fix an issue I encountered with Crush:
https://github.com/charmbracelet/crush
 
When using this agent to work on a Go project, I ran into a problem where processes couldn’t be terminated.
For example, when I asked the LLM to write a simple server using Gin, it wrote the program and tested the endpoints correctly.
However, after testing, when the LLM tried to kill the process using the job kill tool, it just hung infinitely.
 
I found that Crush executes commands using:
 runner.Run(ctx, command) 
from:
https://github.com/charmbracelet/crush/blob/main/internal/shell/shell.go
 
But when a cancel signal is sent to the provided  ctx , the process is not actually killed.
 
I checked the issues and saw other users are facing the same problem:
https://github.com/charmbracelet/crush/issues/2490
His platform is Windows 11.
 
So I want to create a patch for users who run into this.
I tested the second Windows-specific patch on my own machine (Windows 10), and it successfully kills  go run  commands that previously couldn’t be terminated.
 
Since most people hitting this issue are users of the  interp  package,
I think adding an initialization option to the runner would be the most compatible approach.
It would also provide a way for users who really need to ensure child processes are fully killed.

PR #1317, opened 2026-04-05
Feat test n option

Mohamed-Moumni:
## Summary

* Implement the -N file test operator (TsModif), which was previously unsupported and returned an error
* Add interp/actime package with platform-specific GetAtime() functions (darwin, linux, windows) to retrieve file access time, as access time retrieval differs across platforms
* Tested GetAtime() across all supported platforms, including Windows

Implemented #1316

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -1977,10 +1977,6 @@ var runTests = []runTest{
 		"r\nw\nx\n",
 	},
 	{
-		"test -N a",
-		"unsupported unary test op: -N\nexit status 1 #IGNORE",
-	},
-	{
 		"test -? a",
 		// TODO: this error message should refer to `-?`
 		"1:1: not a valid test operator: `a`\n1:1: a must be followed by a word\nexit status 2 #JUSTERR",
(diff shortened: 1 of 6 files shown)


mvdan:
We can't just support the three main GOOSes; e.g. CI also covers js and plan9 because some users care about less widespread platforms. At least there should be a fallback file for other platforms that acts as a no-op by returning the mod time.

We need the type assertions to use the `_, ok` syntax and fall back to the mod time if the type doesn't match, because a custom handler may return an implementation of `fs.FileInfo` that doesn't contain this type in `Sys`.

Finally, name these files consistently with `os_unix.go`, i.e. `os_linux.go`, `os_darwin.go`, and so on.

Mohamed-Moumni:
Thanks for the review. I added `os_other.go` for unhandled platforms, using mod time as a fallback value, and fixed the type assertions to use the` _, ok` syntax with a mod time fallback.

### Changes Made

* `actime_*.go` → `os_*.go` 
* Added the type assertions with mod time as a fallback return

mvdan:
you're exposing public APIs. make these internal APIs in the interp package

Mohamed-Moumni:
I internalized the actime logic, so it's only accessible via the interp package.

mvdan:
I meant in the actual interp package. I'll tweak this myself and merge it later.

Mohamed-Moumni:
Sorry for the confusion, I misunderstood the req, It's fixed now.

mvdan:
The test fails on Windows. Not sure if the test is too flakey with the sleep, or if it's just incorrect on Windows. I'll have to investigate.

mvdan:
It looks like the real windows does not really support access times by default. I'm making the tests unix-only.

PR #1319, opened 2026-04-06
Add `AccessHandler` to handler file permissions

amjadjibon:
https://github.com/mvdan/sh/issues/1318

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -95,6 +95,10 @@ type Runner struct {
 	// statHandler is a function responsible for getting file stat. It must be non-nil.
 	statHandler StatHandlerFunc
 
+	// accessHandler is a function responsible for checking file permissions.
+	// If nil, a default implementation is used.
+	accessHandler AccessHandlerFunc
+
 	stdin  *os.File // e.g. the read end of a pipe
 	stdout io.Writer
 	stderr io.Writer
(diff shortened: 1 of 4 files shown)


mvdan (changes requested):
The fallback logic doesn't seem right. As the TODO says, we need this logic to unconditionally use a handler, rather than use a mix of the real OS and a handler. If we mix both, then the sandboxing is entirely pointless still.

amjadjibon:
> The fallback logic doesn't seem right. As the TODO says, we need this logic to unconditionally use a handler, rather than use a mix of the real OS and a handler. If we mix both, then the sandboxing is entirely pointless still.

 Hi, I'm looking at your feedback about the access handler fallback logic mixing real OS calls with handler calls.                           
                                                                                                                                              
 I'm wondering if adding a new `AccessHandlerFunc` to the handler interface (similar to `StatHandlerFunc`, `OpenHandlerFunc`, etc.) would be the right approach here? This would let users provide custom access handlers for virtual filesystems, similar to how they can already provide custom stat and open handlers.                                                                                                              

Or would you prefer a different direction? The TODO mentions v4, so I want to make sure I'm heading in the right direction before proceeding. 

Thanks!

mvdan:
Thanks. I'll have to take a look at this myself because designing API is not easy, and doing it over a code review does not make it easier.

Issue #339, opened 2018-12-18
base#number notation


rob-myers:
`gosh` doesn't support it e.g. `echo $(( 2#1010 ))` outputs `0` instead of `10`.

Could it be parsed as an arithmetic binary operator for `LangBash`?

mvdan:
The interpreter doesn't support this kind of number, you're right. That's definitely a bug. We can fix that without changing the syntax package.

A year and a half ago, when I first came across this numerical notation in bash and mksh, I simply modified the parser to allow `#` characters within numbers. But that's not a great solution. It's lazy parsing, leaving work to be done by the user, and also accepts incorrect expressions like `foo#123` or `1#2#3`.

I think the parser should definitely parse this as a node somehow. We already abuse `BinaryArithm` for comma operators, so using them here seems fine too.

mvdan:
I'm actually re-evaluating this right now. From `man bash`:

> Constants  with  a  leading 0 are interpreted as octal numbers.  A leading 0x or 0X denotes hexadecimal.  Otherwise, numbers take the form [base#]n, where the optional base is a decimal number between 2 and 64 representing the arithmetic base, and n is a number in that base.  If base# is omitted,  then  base  10  is used.  When specifying n, the digits greater than 9 are represented by the lowercase letters, the uppercase letters, @, and _, in that order.  If base is less than or equal to 36, lowercase and uppercase letters may be used interchangeably to represent numbers between 10 and 35.

So should the syntax nodes also tell you about `0xNNN` and `0NNN`? If we support `BASE#NNN`, seems to me like we should be consistent.

I'm starting to think that, as far as the nodes go, we can keep on leaving them as literals. Then, an interpreter or analyzer can simply have logic like:

```
if contains(lit, "#") {
    // base#number
} else if hasprefix(lit, "0x") {
    // 0xNNN
} else if hasprefix(lit, "0") {
    // 0NNN
} else {
    // NNN
}
```

Of course, the `expand` package would implement this, so one could reuse it too.

But of course, we'd still have to fix the parser to error on invalid input like `3#2#1`, and the interpreter to expand these number literals properly.

This is similar to how the syntax tree in Go itself handles literals - see https://golang.org/pkg/go/ast/#BasicLit. The string there can take one of many standard forms, and then one can easily "evaluate" that number via a public API.

rob-myers:
Fair enough.

[...]

PR #1328, opened 2026-04-24
expand: Support integer bases in arithmetic

stephen-fox:
The existing arithmetic logic is hard coded to always treat strings as base 10 integers. This commit updates that logic to support what bash supports based on its manual page. [1]

References

1. https://www.man7.org/linux/man-pages/man1/bash.1.html

### DIFF
--- a/expand/arith.go
+++ b/expand/arith.go
@@ -111,10 +111,35 @@ func oneIf(b bool) int {
 	return 0
 }
 
-// atoi is like [strconv.ParseInt](s, 10, 64), but it ignores errors and trims whitespace.
+// atoi is like [strconv.ParseInt](s, BASE, 64), but it handles integer
+// base prefixes according to bash-shell's rules, ignores errors, and
+// trims whitespace.
+//
+// For more information about bash's integer base handling syntax,
+// refer to the bash manual:
+// https://www.man7.org/linux/man-pages/man1/bash.1.html
 func atoi(s string) int64 {
 	s = strings.TrimSpace(s)
-	n, _ := strconv.ParseInt(s, 10, 64)
+	base := int64(10)
+	switch {
+	case strings.HasPrefix(s, "0x") || strings.HasPrefix(s, "0X"):
+		base = 16
+		s = s[2:]
+	case strings.HasPrefix(s, "0"):
+		base = 8
+		s = s[1:]
+	default:
+		baseStr, intStr, hasSep := strings.Cut(s, "#")
+		if hasSep {
+			var err error
+			base, err = strconv.ParseInt(baseStr, 10, 8)
+			if err != nil || base > 64 {
+				return 0
+			}
+			s = intStr
+		}
+	}
+	n, _ := strconv.ParseInt(s, int(base), 64)
 	return n
 }
 
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -2074,6 +2074,44 @@ var runTests = []runTest{
 		"2 3\n",
 	},
 	{
+		"echo $((255+1))",
+		"256\n",
+	},
+	{
+		"echo $((0xff+1))",
+		"256\n",
+	},
+	{
+		"echo $((0377+1))",
+		"256\n",
+	},
+	{
+		"echo $((10#255+1))",
+		"256\n",
+	},
+	{
+		"echo $((16#ff+1))",
+		"256\n",
+	},
+	{
+		"echo $((2#11111111+1))",
+		"256\n",
+	},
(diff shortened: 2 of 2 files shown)


mvdan:
What shell syntax does this follow? I don't see any mention of `0b` in `man bash` for example.

stephen-fox:
Hello @mdvan. The `0b` is something I picked up from Go. If we want to adhere strictly to bash, this is what the [manual page](https://www.man7.org/linux/man-pages/man1/bash.1.html) says:

> Integer constants follow the C language definition, without
> suffixes or character constants.  Constants with a leading 0 are
> interpreted as octal numbers.  A leading 0x or 0X denotes
> hexadecimal.  Otherwise, numbers take the form [base#]n, where the
> optional base is a decimal number between 2 and 64 representing
> the arithmetic base, and n is a number in that base.  If base# is
> omitted, then base 10 is used.  When specifying n, if a non-digit
> is required, the digits greater than 9 are represented by the
> lowercase letters, the uppercase letters, @, and _, in that order.
> If base is less than or equal to 36, lowercase and uppercase
> letters may be used interchangeably to represent numbers between
> 10 and 35.

... which I guess looks like this:

```console
bash-3.2$ echo $((0xff+1))
256
bash-3.2$ echo $((0Xff+1))
256
bash-3.2$ echo $((0377+1))
256
bash-3.2$ echo $((10#1+1))
2
bash-3.2$ echo $((16#ff+1))
256
bash-3.2$ echo $((2#11111111+1))
256
```

Happy to update the commit to match that syntax.

mvdan:
Yep, mimicking bash is good. Just add some tests near existing arithmetic ones.

stephen-fox:
@mvdan, I updated the commit to use bash's syntax for integer base handling. I am not a huge fan of the code I wrote, but I think it's a good starting point. I'm not sure if the test format is what you are looking for. But, the tests pass and the bash examples I posted earlier work now.

expand/arith.go
```diff
@@ -111,10 +111,48 @@ func oneIf(b bool) int {
...
+//	optional base is a decimal number between 2 and 64 representing
+//	the arithmetic base, and n is a number in that base. If base# is
+//	omitted, then base 10 is used. When specifying n, if a non-digit
+//	is required, the digits greater than 9 are represented by the
+//	lowercase letters, the uppercase letters, @, and _, in that order.
+//	If base is less than or equal to 36, lowercase and uppercase
+//	letters may be used interchangeably to represent numbers between
+//	10 and 35.
+//
+// - https://www.man7.org/linux/man-pages/man1/bash.1.html
```
mvdan: no need to quote the whole section of the bash man page; simply mention that this mimics the documented bash behavior.

expand/expand_test.go
```diff
@@ -133,6 +133,58 @@ func Test_glob(t *testing.T) {
 	}
 }
 
+func TestArith(t *testing.T) {
```
mvdan: no need to add an entirely new test; you can add test cases in `interp/interp_test.go` after the existing arithmetic ones with `echo $((...))`. An advantage of that route is that those get run against bash as well, so we catch any discrepancy.

mvdan (approved):
Thanks! SGTM, just two thoughts.

mvdan:
Also, please add "Fixes #339" to the commit message.

expand/arith.go
stephen-fox: I removed it.

expand/expand_test.go
stephen-fox: I moved the tests around as requested.

stephen-fox:
Hey @mvdan, I have updated the commit with the requested changes and ran the `interp` package's tests on a Debian machine and they passed.

mvdan (approved):
Thanks very much!

mvdan:
Hm, the tests failed on 32-bit.

stephen-fox:
Awesome. I can probably make the test pass by just using a smaller number... It does make me wonder about sh's handling of integer bit width and what the "correct" behavior would look like though.

mvdan:
All the arithmetic code should be on 64-bit integers; there might be a spot that isn't the case. This might be a latent bug and not introduced by your fix, but perhaps just surfaced by your test.

mvdan:
Ah, I see that `expand.Arithm` returns `int` rather than `int64`, and I have a TODO for that. Perhaps just comment out the test case you have here and add a smaller one for now. I'll deal with the int vs int64 issue at a later time.

stephen-fox:
I commented out that test and added another with a smaller number.

PR #1329, opened 2026-04-25
fix: restore terminal foreground control to support interactive commands

mkaaad:
# Problem
Previously, only the process group was modified, but **terminal foreground ownership was not transferred** to child processes.
This broke interactive programs such as `bash`, `vim`, and REPL environments:
- Terminal signals（SIGINT/Ctrl+C）could not be delivered to child processes
- Interactive input/output abnormal
- Interactive shells and editors could not work properly

# Solution
Add explicit foreground process group control via `TIOCSPGRP` ioctl:
1. Switch the child process group to the terminal foreground before waiting
2. Restore the original shell process group as foreground after the command exits
3. Fully deliver terminal keyboard signals to child processes

# Key Changes
- Add `setProcessForeground` / `restoreForeground` helper functions
- Switch foreground process group when starting interactive commands
- Restore shell foreground control by `defer`

# About ignore `SIGTTOU`
When a **background process** calls the `TIOCSPGRP` ioctl to modify the terminal foreground group:
1. The kernel sends `SIGTTOU` to the calling process by default
2. The default disposition of `SIGTTOU` is **stop the process**, which will freeze the program
3. Temporarily ignoring `SIGTTOU` avoids suspension during foreground switching
4. Restore the original signal handler immediately after the syscall to avoid side effects

### DIFF
--- a/cmd/gosh/main.go
+++ b/cmd/gosh/main.go
@@ -11,7 +11,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 
 	"golang.org/x/term"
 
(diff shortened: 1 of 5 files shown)


mvdan:
Could you clarify if this is indeed what shells like Bash do already?

mvdan:
cc @andreynering

mkaaad:
Based on the Bash source code (`jobs.c`), its terminal management mechanism works as follows:

When spawning a child process group, Bash transfers terminal ownership to the foreground process group: (jobs.c:3905-3918)
```c
/* Save the tty settings before we start the job in the foreground. */
if (foreground)
  {
#if defined (READLINE)
    /* Don't fetch the terminal attributes if we're doing this from a key
       binding or programmable completion. */
    if (RL_ISSTATE(RL_STATE_COMPLETING|RL_STATE_DISPATCHING|RL_STATE_TERMPREPPED) == 0)
#endif
    get_tty_state ();
    save_stty = shell_tty_info;
    jobs[job]->flags &= ~J_ASYNC;	/* no longer async */
    /* Give the terminal to this job. */
    if (IS_JOBCONTROL (job))
      give_terminal_to (jobs[job]->pgrp, 0);
```

When the foreground child process group exits, Bash restores terminal control back to the shell's own process group. (jobs.c:3250-3253)
```c
    if ((flags & JWAIT_NOTERM) == 0 && running_in_background == 0 &&
        (job == NO_JOB || IS_ASYNC (job) == 0 || IS_FOREGROUND (job)) &&
        (subshell_environment & (SUBSHELL_ASYNC|SUBSHELL_PIPE)) == 0)
      give_terminal_to (shell_pgrp, 0);
```

Reference source: https://cgit.git.savannah.gnu.org/cgit/bash.git/tree/jobs.c

PR #1331, opened 2026-05-04
interp: add stmt, subshell, builtin, func-call, and var-lookup handlers

saracen:
`ExecHandlers` already lets external code wrap external command execution. Apply the same middleware-chain pattern to five more dispatch points so the interpreter can be extended from outside without patching its internals.

`StmtHandlers` fires around every `syntax.Stmt` before dispatch. `SubshellHandlers` fires around each subshell creation site, with a `SubshellKind` to distinguish background, paren, cmd subst, proc subst, and pipeline. `BuiltinHandler` registers a name-keyed override for a builtin; `BuiltinHandlers` wraps all builtin dispatch with middleware. `FuncCallHandlers` wraps the body of a declared shell function. `LookupVarHandlers` wraps lookupVar on top of the runner's own special-variable and environment resolution, so middlewares can add or override variables without losing the existing behaviour.

Each chain follows `ExecHandlers`' shape: middlewares are chained first-to-last, the bottom of the chain is the runner's own behaviour, and a middleware may short-circuit by not calling next. Chain bottoms recover the active runner from ctx, so the chains are built once in `Reset` and inherited by subshell copies the same way `execHandler` is.

Also add `RunStmts`, a top-level helper for dispatching stmts on the runner attached to `ctx`, useful for handlers that need to execute shell code in the runner's current scope (e.g. a trap body from a Go-side signal handler).

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -85,6 +85,29 @@ type Runner struct {
 	// The slice is needed to preserve the relative order of middlewares.
 	execMiddlewares []func(ExecHandlerFunc) ExecHandlerFunc
 
+	// stmtHandler is built from stmtMiddlewares when Reset is first called.
+	stmtHandler     StmtHandlerFunc
+	stmtMiddlewares []func(StmtHandlerFunc) StmtHandlerFunc
+
+	// subshellHandler is built from subshellMiddlewares when Reset is first called.
+	subshellHandler     SubshellHandlerFunc
+	subshellMiddlewares []func(SubshellHandlerFunc) SubshellHandlerFunc
+
+	// builtinOverrides is the registry populated by [BuiltinHandler]. It may be nil.
+	builtinOverrides map[string]BuiltinHandlerFunc
+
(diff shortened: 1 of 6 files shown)


mvdan:
This is a large amount of new API surface. Do you really need all of it? And if so, what for?

I can't really review this as a single commit. At the very least, split this into multiple commits so that I can review one by one.

Some high level thoughts:

* Why do you need to customize subshells and func bodies? Those seem like very specific entrypoints. Will we also add custom handlers for pipes, redirects, arithmetic, for loops, and other shell features? Rather than adding 20 more API entrypoints, I would rather think about generic hooks.
* You already have CallHandler if you want to replace or bypass builtins. You can, for example, rewrite `echo` into `_echo` and then implement that via ExecHandler.
* A LookupVar handler seems odd to me. You can already supply an environment via Runner.Env. If this is not enough, then please elaborate further.

mvdan:
Friendly nudge @saracen

mvdan:
Closing for now. Happy to discuss if you open an issue like a feature request.

PR #1332, opened 2026-05-09
syntax: apply BinaryNextLine to test and arithmetic expressions

### LINKED ISSUE #813: syntax: make BinaryNextLine also apply to binary test and arithmetic expressions
When using -bn option, only the first of the two examples below is correct. In the second case below, the binary operators are not formatted to be at the beginning of the line. The second example even gets re-formatted, so that the operators, once corrected manually, appear at the end of line again.

```
#!/bin/bash
# ok, pipes in front of line
if [[ -z "${foo}" ]] \
    || [[ -z "${bar}" ]] \
    || [[ -z "${baz}" ]]; then
    echo "Hello world"
fi

# nok, pipes at the end of line
if [[ -z "${foo}" ||
    -z "${bar}" ||
    -z "${baz}" ]]; then
    echo "Hello world"
fi
```

kimjune01:
Fixes #813.

The `-bn` flag moves binary operators to the start of the next line for pipes and logical operators in commands, but didn't apply to `[[ ]]` test expressions or `$(( ))` arithmetic expressions. This meant `shfmt -bn` would reformat operators back to the end of the line in those contexts.

Handle `BinaryTest` (`&&`, `||`) and `BinaryArithm` (`+`, etc.) the same way `BinaryCmd` already works: when the operand spans multiple lines, move the operator to the beginning of the continuation line with proper indentation.

Test cases cover single-line (unchanged), two-operand, three-operand chaining, and the exact example from the issue.

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -886,6 +886,15 @@ func (p *Printer) arithmExprRecurse(expr ArithmExpr, compact, spacePlusMinus boo
 			p.arithmExprRecurse(expr.X, compact, spacePlusMinus)
 			p.w.WriteString(expr.Op.String())
 			p.arithmExprRecurse(expr.Y, compact, false)
+		} else if p.binNextLine && !p.singleLine && expr.Y.Pos().Line() > p.line {
+			p.arithmExprRecurse(expr.X, compact, spacePlusMinus)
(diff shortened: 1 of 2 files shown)


mvdan:

PR #1334, opened 2026-05-09
interp: dup pipe fds in pipelines so EOF and SIGPIPE propagate

### LINKED ISSUE #1142: pipelining and redirection are broken because gosh does not close file descriptors
# Versions

- go version: upstream `656b5b3abe25` ("internal/poll: don't skip empty writes on Windows", 2025-03-28)
- gosh version: upstream `bf4faa20c0ea` ("interp: clarify handler docs a bit", 2025-03-28)

# Description

For constructing a pipeline of N commands, a shell performs something like this:
1. call `pipe()` N-1 times
2. call `fork()` N times
3. in each child process:
   1. call `dup2()` to duplicate the read end of the pipe to the left (if any) over fd#0
   2. call `dup2()` to duplicate the write end of the pipe to the right (if any) over fd#1
   3. close all pipe fds from step 1
   4. `exec()`
4. in the parent, close all pipe fds from step 1

The bug is that `gosh` omits the last step, and it breaks pipeline behavior.

# Reproducer

~~~
gosh
$ dash -c 'exec >/dev/null; exec sleep 3600' | cat >/dev/null &

$ ls -g -o /proc/$(pidof sleep)/fd/1
l-wx------. 1 64 Mar 28 19:19 /proc/781245/fd/1 -> /dev/null

$ ls -g -o /proc/$(pidof cat)/fd/0
lr-x------. 1 64 Mar 28 19:19 /proc/781246/fd/0 -> 'pipe:[592425]'

$ ls -g -o /proc/$$/fd | grep -F 'pipe:[592425]'
lr-x------. 1 64 Mar 28 19:20 3 -> pipe:[592425]
l-wx------. 1 64 Mar 28 19:20 4 -> pipe:[592425]
~~~

Explanation:
- The first command in the pipeline drops its fd#1 reference to the pipe. At this point, there should be no writers left for the pipe.
- However, the second command in the pipeline does not see EOF on the pipe; it blocks on reading the pipe.
- Turns out that gosh itself keeps the pipe open for writing (fd#4), which is what prevents `cat` from seeing EOF.
- `cat` hangs uselessly, as a result, instead of exiting.
- The problem occurs because gosh does not close file descriptors 3 and 4 right after the pipeline is constructed (step 4 at the top).

# Expected behavior
~~~
dash
$ dash -c 'exec >/dev/null; exec sleep 3600' | cat >/dev/null &

$ ls -g -o /proc/$(pidof sleep)/fd/1
l-wx------. 1 64 Mar 28 19:25 /proc/781363/fd/1 -> /dev/null

$ ls -g -o /proc/$(pidof cat)/fd/0
ls: cannot access '/proc//fd/0': No such file or directory

$ ls -g -o /proc/$$/fd | grep -F 'pipe:'
(nothing)
~~~

Dash doesn't keep the pipe open (at all); thus, `cat` reads EOF (and exits), after the first command in the pipeline drops its reference -- the only reference -- to the write end of the pipe. This is the proper behavior.

# Alternative reproducer

This one uses redirection operators with a named pipe (FIFO), instead of a pipeline.

## Erroneous behavior

~~~
gosh
$ rm -f fifo
$ mkfifo fifo
$ dash -c 'exec >/dev/null; exec sleep 3600' >fifo &
$ cat <fifo >/dev/null &

$ ls -g -o /proc/$(pidof sleep)/fd/1
l-wx------. 1 64 Mar 28 19:35 /proc/781581/fd/1 -> /dev/null

$ ls -g -o /proc/$(pidof cat)/fd/0
lr-x------. 1 64 Mar 28 19:35 /proc/781580/fd/0 -> .../fifo

$ ls -g -o /proc/$$/fd | grep -F 'fifo'
l-wx------. 1 64 Mar 28 19:36 3 -> .../fifo
lr-x------. 1 64 Mar 28 19:36 4 -> .../fifo
~~~

## Expected behavior

~~~
dash
$ rm -f fifo
$ mkfifo fifo
$ dash -c 'exec >/dev/null; exec sleep 3600' >fifo &
$ cat <fifo >/dev/null &

$ ls -g -o /proc/$(pidof sleep)/fd/1
l-wx------. 1 64 Mar 28 19:35 /proc/781581/fd/1 -> /dev/null

$ ls -g -o /proc/$(pidof cat)/fd/0
ls: cannot access '/proc//fd/0': No such file or directory

$ ls -g -o /proc/$$/fd | grep -F 'fifo'
(nothing)
~~~

qiangli:
## Summary

When the parent Go process holds extra references to a pipeline's read or write end, EOF does not propagate to readers and SIGPIPE is not delivered to writers. Concretely, `yes | head -1` hangs forever instead of terminating after `head` closes its stdin.

## Approach

Duplicate both pipe ends with `syscall.Dup` before handing them to the goroutines that run the left and right sides, then immediately close the originals in the parent. The parent no longer holds any reference to the pipe; closing the duplicates in the goroutines is now sufficient to drive EOF/SIGPIPE through the pipeline.

- Duplicates are marked `CloseOnExec` so external commands spawned from inside the goroutines do not inherit them.
- Restoring `r.stdin` from `oldStdin` after `wg.Wait()` prevents the right-hand side's stdin replacement from leaking into subsequent statements in the enclosing block.
- A new helper `dupPipeFd` lives in `interp/os_unix.go` on unix platforms and as a no-op fallback in `interp/os_notunix.go`. On non-unix the previous behaviour is preserved (pipelines run but EOF/SIGPIPE propagation is best-effort).

Fixes #1142.

## Test plan

- [x] `go test -short ./interp/` passes.
- [x] Manual: `yes | head -1` exits promptly under the patched runner; previously hung indefinitely.

### DIFF
--- a/interp/os_notunix.go
+++ b/interp/os_notunix.go
@@ -8,12 +8,20 @@ package interp
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"mvdan.cc/sh/v3/syntax"
 )
 
 func mkfifo(path string, mode uint32) error {
 	return fmt.Errorf("unsupported")
+}
+
+// dupPipeFd is a no-op on non-unix platforms; the original pipe fd is
+// returned. Pipelines will still run, but EOF/SIGPIPE propagation is
+// best-effort because the parent retains the original fd reference.
+func dupPipeFd(f *os.File) (*os.File, error) {
+	return f, nil
 }
 
 // access attempts to emulate [unix.Access] on Windows.
--- a/interp/os_unix.go
+++ b/interp/os_unix.go
@@ -7,6 +7,7 @@ package interp
 
 import (
 	"context"
+	"os"
 	"os/user"
 	"strconv"
 	"syscall"
@@ -17,6 +18,20 @@ import (
 
 func mkfifo(path string, mode uint32) error {
 	return unix.Mkfifo(path, mode)
+}
+
+// dupPipeFd duplicates a pipe file descriptor, returning a new *os.File
+// that refers to the same underlying pipe endpoint. The caller can close
+// the original fd while the duplicate remains valid. This is used to
+// ensure the parent process does not hold extra pipe fd references during
+// pipeline execution, which would prevent EOF/SIGPIPE propagation.
+func dupPipeFd(f *os.File) (*os.File, error) {
+	newFd, err := syscall.Dup(int(f.Fd()))
+	if err != nil {
+		return nil, err
+	}
+	syscall.CloseOnExec(newFd)
+	return os.NewFile(uintptr(newFd), f.Name()+"-dup"), nil
 }
 
 // access is similar to checking the permission bits from [io/fs.FileInfo],
--- a/interp/runner.go
+++ b/interp/runner.go
@@ -487,23 +487,51 @@ func (r *Runner) cmd(ctx context.Context, cm syntax.Command) {
 				r.exit.fatal(err) // not being able to create a pipe is rare but critical
 				return
 			}
+			// Duplicate pipe fds for use by the goroutines, then close
+			// the originals. This ensures the parent process does not
+			// hold extra references to the pipe, so that:
+			//   - EOF propagates when the writer closes its end
+			//   - SIGPIPE is delivered when the reader closes its end
+			// Builtins and external commands in the goroutines use the
+			// duplicated fds, which remain valid until explicitly closed.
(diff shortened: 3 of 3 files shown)


mvdan:
Can we write a regression test for this that fails without the fix? Otherwise, the fix is going to be rather fragile.

mvdan:
@qiangli did you see my comment above?

qiangli:
Sorry, I did not see. I may create another pull request, closing this one for now.

PR #1336, opened 2026-05-18
interp: have Runner.access consult StatHandler before falling back to unix.Access

c-tonneslan:
Closes #1318.

On Unix, `Runner.access` went straight to `unix.Access` (host FS) without asking the configured `StatHandler` first. Interpreters wired up against a virtual filesystem therefore saw `cd` / `[ -x foo ]` / etc. fail for paths the `StatHandler` resolves fine. The non-Unix implementation already did the right thing - check the bits returned by `r.lstat`. Do the same on Unix and only fall through to `unix.Access` when `r.lstat` can't see the path.


mvdan:
No; see the TODO right above the code you're adding. And see the original reason why we stopped using "stat": https://github.com/mvdan/sh/issues/1033

c-tonneslan:
Sorry for the noise, makes sense.

PR #1337, opened 2026-05-19
Add gosh link to README.md

phuslu:
This adds `gosh` to the README's list of noteworthy integrations.

I create `gosh` co-works with AI copilot which is a compact Bash-style shell on top of `mvdan.cc/sh/v3`, it builds an interactive shell experience around it with readline support, history, completion, prompt rendering, and Bash-flavored compatibility helpers.

I think it fits this section as another user-facing project built directly on top of `mvdan.cc/sh`.


mvdan:
This conflicts with `./cmd/gosh`, so I'm afraid it would be rather confusing to add here. The name is also pretty overused, as there are over a dozen existing projects with that name. For our shell it's fine because it's mainly meant as a small proof of concept, but it's best if we don't confuse things further.

PR #1341, opened 2026-06-02
cmd/shfmt: add shell completion support

### LINKED ISSUE #914: [feature] Add shell completions
How about adding shell completions for bash/zsh/fish/posh/etc? Thanks.

```
❯ shfmt -<TAB>
option
--binary-next-line   binary ops like && and | may start a line
-bn                  binary ops like && and | may start a line
--case-indent        switch cases will be indented
-ci                  switch cases will be indented
...
```

toller892:
Add `--completion` flag that generates shell completion scripts for bash, zsh, and fish.

## Changes

- New file: `cmd/shfmt/completion.go` - completion script generation
- Modified: `cmd/shfmt/main.go` - add completion flag and handling
- Modified: `cmd/shfmt/shfmt.1.scd` - document the new flag
- Modified: `cmd/shfmt/testdata/script/flags.txtar` - add tests

## Usage

```bash
# Bash
source <(shfmt --completion=bash)

# Zsh
source <(shfmt --completion=zsh)

# Fish
shfmt --completion=fish > ~/.config/fish/completions/shfmt.fish
```

## Features

- Completes all flags (short and long forms)
- Provides value completions for `--language-dialect` (bash/posix/mksh/bats/zsh/auto)
- Provides value completions for `--indent` (0/2/4/8)
- Falls back to file completion for arguments
- Proper error message for unsupported shells

Fixes #914

### DIFF
--- a/cmd/shfmt/completion.go
+++ b/cmd/shfmt/completion.go
@@ -1,0 +1,142 @@
+// Copyright (c) 2016, Daniel Martí <mvdan@mvdan.cc>
+// See LICENSE for licensing information
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+)
+
+var completionFlag = flagVal("", "completion", "", flag.StringVar)
+
+func printCompletion(shell string) {
+	switch shell {
+	case "bash":
+		fmt.Print(bashCompletion)
+	case "zsh":
+		fmt.Print(zshCompletion)
+	case "fish":
+		fmt.Print(fishCompletion)
+	default:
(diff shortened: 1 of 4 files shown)


mvdan:
Thanks. This unfortunately means that I have to maintain these scripts by hand, because they duplicate all the flag information. Is there no way to do this with the `flag` package?

I know that CLI packages like `cobra` support completion generation, but they are also rather heavy. If there is a small library that supports completions with the standard `flag` package I'm happy to use it. If no such software exists, then I'm happy to not have completions for now.

mvdan:
Closing for now. I don't really intend to maintain this myself, even if it's AI generated.

Issue #1348, opened 2026-06-10
Single line functions without semicolon yield parser error: reached EOF without matching `{` with `}`


joshkehn:
As far as I'm aware, this isn't valid posix. So it would make sense for `shfmt` to fail to parse something like:

```
❯ cat ~/f.sh
f() { echo "foo" }
❯ shfmt -w ~/f.sh
f.sh:1:5: reached EOF without matching `{` with `}`
```

However, if I'm running in `zsh`, the *script* runs this just fine masking the problem away from me.

Can we handle this better in `shfmt`? Some options could be:

1. Throw a more readable error instead of a parse error, something like "expression not valid in language dialect `dialect`: ..."
2. If the shell is `zsh`, treat this as a `zsh` script and put a semicolon at the end.
3. If the shell is `zsh` handle this gracefully but still treat it as a `bash` script and put a semicolon there since it was probably intended and the formatted output is more portable/correct.
4. Handle this unconditionally for all `bash` scripts independent of current shell status.

While I'm probably not the first person to lazily write `f() { echo "blah" }` in a script and found it working despite not being valid for a `.sh` extension. 

There's also a related bug with the `zsh` dialect:
```
❯ cat ~/ff.sh
#!/usr/bin/env zsh
f() { echo foo}
❯ shfmt -w ~/ff.sh
ff.sh:2:5: reached EOF without matching `{` with `}`
```

mvdan:
duplicate of https://github.com/mvdan/sh/issues/1295

joshkehn:
nice! i had tried to search existing issues but didn't find that one

PR #1349, opened 2026-06-10
syntax: close zsh brace blocks on an attached '}'

joshkehn:
See #1348. Update syntax and allow `}` to close a brace block instead of treating it as part of the word.


mvdan:
Changes to the lexer are quite tricky; I'm afraid that this is not really in the right direction and I would rather tackle this myself directly.

joshkehn:
Totally fair. Thanks for ack'ing!

PR #1350, opened 2026-06-11
interp: respect StatHandler in Unix access fallback

VrtxOmega:
Fixes #1318.

## Summary

- keep `unix.Access` as the primary Unix permission check for real filesystem paths
- fall back to the runner's `StatHandler` only when the host path is missing, so virtual filesystems can authorize access by mode bits
- add a Unix regression test for `cd /virtual` with a custom `StatHandler`

## Validation

- `go test ./interp -run TestRunnerAccessStatHandlerVirtualDir -count=1`
- `go test ./interp -count=1`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/interp/os_unix.go
+++ b/interp/os_unix.go
@@ -7,6 +7,8 @@ package interp
 
 import (
 	"context"
+	"errors"
+	"io/fs"
 	"os/user"
 	"strconv"
 	"syscall"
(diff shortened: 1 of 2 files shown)


mvdan:
No. See the TODO.

PR #1355, opened 2026-07-04
syntax: parse arithmetic array index with # subscript key

### LINKED ISSUE #1285: syntax: errors and/or breaks code when arrays have keys that contain pound signs or variables
I'm using shfmt 3.12.0.

The following is valid Bash, and when used in the context of a larger script ([like this](https://github.com/openjck/dotfiles/blob/e422d7a2dc9188c39f8b9dd92d8f399ea3c14e58/bin/general/entr-fzf#L97-L100)), it works as expected:

```bash
#!/usr/bin/env bash

# ...

for ((I=0; I < ARGS[COMMAND,#]; I++)); do
  ENTR_ARGS+=("${ARGS[COMMAND,$I]}")
done
```

In the above code, the `ARGS` array has a key named `COMMAND,#` and also keys named `COMMAND,1`, `COMMAND,2`, etc. It might look a little unusual, but the code is valid, and that pattern is used by [`docopts`](https://github.com/docopt/docopts?tab=readme-ov-file#associative-array-mode).

When `shfmt` is run on that code, the following error occurs:

```
ef:4:28: , must be followed by an expression
```

If that issue is fixed, by quoting the name of the key, like this:

```bash
#!/usr/bin/env bash

# ...

for ((I=0; I < ARGS['COMMAND,#']; I++)); do
  ENTR_ARGS+=("${ARGS[COMMAND,$I]}")
done
```

...and shfmt is run on the file again, then the code is formatted to the following, which does not work:

```bash
#!/usr/bin/env bash

# ...

for ((I = 0; I < ARGS['COMMAND,#']; I++)); do
  ENTR_ARGS+=("${ARGS[COMMAND, I]}")
done
```

In the second-to-last line, the line beginning with `ENTR_ARGS+=`, notice that a space was inserted after the comma and the dollar sign was removed from the `I` variable.

As before, the workaround appears to be quoting the name of the key. The nested double quotes look weird, but it works as expected, so I assume it's valid Bash:

```bash
#!/usr/bin/env bash

# ...

for ((I = 0; I < ARGS['COMMAND,#']; I++)); do
  ENTR_ARGS+=("${ARGS["COMMAND,$I"]}")
done
```

momomuchu:
`shfmt` fails to parse the valid bash arithmetic array index `ARGS[COMMAND,#]` with "`,` must be followed by an expression", and on the quoted workaround it corrupts the code on re-format (#1285).

Two parts, both needed:
- `syntax/lexer.go`: in arithmetic mode, `#` was glued onto the preceding literal instead of breaking into its own token. It now breaks, except after a pure-digit prefix so base-N literals like `16#FF` and `3#20` keep working.
- `syntax/parser_arithm.go`: a bare `#` token is now accepted as a subscript key value in `arithmExprValue`.

Added a fixture to `syntax/filetests_test.go`, which also checks print idempotency. It fails without the fix (parse error across bash/mksh/zsh) and passes with it, with round-trip formatting stable. `go test ./syntax/...`, `go vet` and `gofmt` clean.

Closes #1285

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -3414,6 +3414,18 @@ var fileTests = []fileTestCase{
 		}), LangBash|LangMirBSDKorn|LangZsh),
 	),
 	fileTest(
+		[]string{`$((ARGS[COMMAND,#]))`},
+		langFile(arithmExp(word(&ParamExp{
+			Short: true,
+			Param: lit("ARGS"),
+			Index: &BinaryArithm{
+				Op: Comma,
+				X:  litWord("COMMAND"),
+				Y:  litWord("#"),
+			},
+		})), LangBash|LangMirBSDKorn|LangZsh),
+	),
+	fileTest(
 		[]string{"$((a += 2, b -= 3))"},
 		langFile(arithmExp(&BinaryArithm{
 			Op: Comma,
--- a/syntax/lexer.go
+++ b/syntax/lexer.go
@@ -17,6 +17,20 @@ func asciiDigit[T rune | byte](r T) bool {
 	return r >= '0' && r <= '9'
 }
 
+// allDigits reports whether bs is non-empty and consists entirely of ASCII
+// digits, as in the base of a base-N numeric literal like 16#FF.
+func allDigits(bs []byte) bool {
+	if len(bs) == 0 {
+		return false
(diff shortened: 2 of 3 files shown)


mvdan:

PR #1356, opened 2026-07-16
interp: don't permanently disable traps after a trap callback parse error

josharian:
When a trap callback failed to parse, r.handlingTrip didn't get reset.
As a result, all subsequent trap callbacks were silently
skipped for the rest of the runner's lifetime.


josharian:
(I have another feature-y PR coming after this one, but this was an indepdently useful/correct bug fix, so figured I'd get it dispatched first.)

mvdan (approved):
yikes, nice catch.

PR #1357, opened 2026-07-18
interp: provide HandlerContext to stat handlers

josharian:
StatHandlerFunc is supposed to get a HandlerContext.
It doesn't, and hasn't since the option was introduced.
Fix that.


josharian:
Pretty sure that CI failure is a flake. Don't see a way to re-run it myself.

mvdan (approved):
🫪🙈

PR #1358, opened 2026-07-18
syntax: preserve comments terminating continued commands

### LINKED ISSUE #1290: Undoes comment terminating multi-line lists or commands to avoid unnecessary diff modification
Hi!

To avoid unnecessary diff modifications, reduce potential accidents, and help with sortability in some cases, I've been using the following pattern on multi-line lists in commands:

```
#!/bin/sh

cmd_list \
  aa \
  bb \
  cc \
  # EOL

cmd_a \
  | cmd_b \
  | cmd_c \
  | cmd_d \
  # EOC

other
```

But running for example `shfmt -i=2 -bn`, disarms this pattern into:

```
#!/bin/sh

cmd_list \
  aa \
  bb \
  cc
# EOL

cmd_a \
  | cmd_b \
  | cmd_c \
  | cmd_d
# EOC

other
```

This pattern makes it so that adding, modifying or removing lines changes the minimal amount as possible (except if you need to insert something before `cmd_a`).

snowyukitty:
Fixes #1290.

Associate a terminal comment reached through an escaped newline with its
statement. When `BinaryNextLine` is enabled, the printer can then keep the
final continuation instead of moving the comment out of the command.

Ordinary following comments remain unchanged. Tests cover both issue examples,
variable whitespace, CRLF, following statements, multiple comments, and
word-internal or semicolon-terminated negative cases.

### DIFF
--- a/syntax/parser.go
+++ b/syntax/parser.go
@@ -2144,7 +2144,9 @@ func (p *Parser) getStmt(readEnd, binCmd, fnBody bool) *Stmt {
 	}
 	if len(p.accComs) > 0 && !binCmd && !fnBody {
 		c := p.accComs[0]
-		if c.Pos().Line() == s.End().Line() {
+		// A later-line comment can only be reached here through an escaped
+		// newline, as a regular newline ends getStmt first.
+		if c.Pos().Line() >= s.End().Line() {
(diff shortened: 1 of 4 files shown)


mvdan:

PR #1359, opened 2026-07-19
syntax: support zsh short form for loops

jsvk:
`zsh` supports short-form for loops like this:

```zsh
% arr=(1 2 3)
% for i ($arr) echo $i
1
2
3
```

Today, when i try to format a script that uses this, i get this error:

```
% shfmt aliases.sh
aliases.sh:1345:5: `for foo` must be followed by `in`, `do`, `;`, or a newline
```

this PR adds support for this syntax. The resulting output format is this:

```zsh
# file contents -------------------------------
% cat test.sh
#!/usr/bin/env zsh

set -euo pipefail

main() {
    for    e    (1 2 3) echo $e
      }

main "$@"
% # validate syntax  -------------------------------
% zsh test.sh
1
2
3
% # with this pr -------------------------------
% go run ./cmd/shfmt test.sh
#!/usr/bin/env zsh

set -euo pipefail

main() {
	for e (1 2 3) echo $e
}

main "$@"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+- **syntax**
+  - Support Zsh short `for` loops with a parenthesized word list, such as
+    `for i ($arr) echo $i` and `for i (1 2 3) { echo $i }`
+
 ## [3.13.1] - 2026-03-09
 
 - **cmd/shfmt**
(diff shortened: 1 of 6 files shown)


mvdan:
Thanks. This is new API, so I need to think carefully about it, and hence implement it myself.

jsvk:
sure, sounds good

PR #1360, opened 2026-07-19
interp: add HandlerContext.LastExitStatus

josharian:
I've been doing some error handling work,
and have been missing the ability for handlers
to observe the exit status of the previous command.

Expose that value (i.e. what "$?" would expand to)
as HandlerContext.LastExitStatus.

Happily, doing that fixes two related bugs:

Inside a trap callback, "$?" now holds the status of the command
that triggered the trap, rather than its predecessor.

Run now records the exit status of a bare Command,
so an exit trap and subsequent incremental calls to Run see it in "$?".

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -926,6 +926,8 @@ func (r *Runner) Run(ctx context.Context, node syntax.Node) error {
 	default:
 		return fmt.Errorf("node can only be File, Stmt, or Command: %T", node)
 	}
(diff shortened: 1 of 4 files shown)


josharian:
The actual feature addition I was heading towards. :) I'm not particular about naming or anything. Feel free to Just Change It if the general approach is acceptable.

mvdan (approved):
Thanks!

PR #1362, opened 2026-07-21
interp: run ENOEXEC scripts via /bin/sh

### LINKED ISSUE #1065: interp: bash runs scripts if kernel returns ENOEXEC
make a file like such, named `foo`:
```
echo "bar"
```

```
$ chmod +x foo
$ bash -c "./foo"
hi
```

when I run this with `strace -f bash -c ./foo` I see that 

```
execve("./testdata/foo", ["./testdata/foo"], 0x56158959fca0 /* 48 vars */) = -1 ENOEXEC (Exec format error)
...            
openat(AT_FDCWD, "./testdata/foo", O_RDONLY) = 3                                                                                                  
newfstatat(AT_FDCWD, "./testdata/foo", {st_mode=S_IFREG|0750, st_size=11, ...}, 0) = 0                                                            
read(3, "echo \"bar\"\n", 80)           = 11                                                                                                      
lseek(3, 0, SEEK_SET)                   = 0                                                                                                       
dup2(3, 255)                            = 255                                                                                                     
close(3)                                = 0                                                                                                       
read(255, "echo \"bar\"\n", 11)         = 11                                                                                                      
write(1, "bar\n", 4)                    = 4                                    
```

Looks to me like bash will try to execute as a script files with execute permission that get ENOEXEC from the kernel.

This looks like something that could be implemented as an `ExecHandlerFunc`.

l46983284-cpu:
#### Description
When the kernel rejects an executable text script with `ENOEXEC` (commonly a missing shebang), bash retries the script via `/bin/sh`. `interp.DefaultExecHandler` previously surfaced the fork/exec error instead.

This matches bash for that case:
- detect `ENOEXEC` from `cmd.Start`
- re-run via `sh` from `PATH`, falling back to `/bin/sh` when `PATH` is empty
- keep Windows behavior unchanged

Includes regression tests for relative scripts with and without a working directory change.

Closes #1065

#### Checklist
- [x] tests: `go test ./interp`

### DIFF
--- a/interp/handler.go
+++ b/interp/handler.go
@@ -5,6 +5,7 @@ package interp
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
@@ -14,6 +15,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
 
 	"mvdan.cc/sh/v3/expand"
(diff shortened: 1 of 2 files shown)


mvdan:
Thanks. This has no tests and the logic in question is rather important, so I would rather do this myself.

PR #1364, opened 2026-07-25
syntax: allow chained zsh parameter subscripts

### LINKED ISSUE #1361: zsh: multiple parameter subscripts fail to parse
The following valid Zsh syntax fails to parse:

```zsh
var=(foo)
echo ${var[1][2]}
```

```console
$ shfmt --version
v3.13.2-0.20260718060058-20744fb2d3e9

$ printf '%s\n' 'var=(foo)' 'echo ${var[1][2]}' | shfmt -ln=zsh
<standard input>:2:14: not a valid parameter expansion operator: `[`
```

Zsh accepts the same input and prints `o`.

See [Zsh docs: Parameter subscripting](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Rules), which uses `${var[1][2]}` as an example.

l46983284-cpu:
## Summary
Parse and print chained zsh subscripts like `${var[1][2]}` (and short `$var[1][2]`) under `LangZsh`.

Previously the second `[` was rejected as an invalid parameter expansion operator.

Closes #1361

## Test plan
- [x] `go test ./syntax/ -run TestZshMultiSubscript`
- [x] `go test ./syntax/`
- [x] `printf 'echo ${var[1][2]}\n' | go run ./cmd/shfmt -ln=zsh`

### DIFF
--- a/syntax/nodes.go
+++ b/syntax/nodes.go
@@ -660,6 +660,9 @@ type ParamExp struct {
 	// TODO(v4): rename Index to Subscript, which better matches bash and zsh terminology
 
 	Index ArithmExpr // ${a[i]}, ${a["k"]}, or a ${a[i,j]} slice with [LangZsh]
+	// ExtraIndexes holds additional consecutive subscripts like ${a[i][j]} with [LangZsh].
+	// Index is always the first subscript when any are present.
+	ExtraIndexes []ArithmExpr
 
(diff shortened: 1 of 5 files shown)


mvdan:
I understand that you have AI tokens to spare, but please stop sending PRs. This doesn't follow the way we typically write test cases. And any new API needs careful thought.

PR #1366, opened 2026-07-26
syntax: keep multiline commands out of inline bodies

### LINKED ISSUE #720: syntax: don't allow multiline commands to be "inline"
We allow complex commands like `for` to be a single line, as long as their inner command also fits in a single line:

```
$ echo 'for i in 1 2 3; do if true; then foo; fi; done' | shfmt
for i in 1 2 3; do if true; then foo; fi; done
```

However, we have broken edge cases. For example, if the inner `if` turns into multiple lines due to having two commands in its body, we still allow the whole `if` to be "inline", even though it doesn't take a single line:

```
$ echo 'for i in 1 2 3; do if true; then foo; bar; fi; done' | shfmt
for i in 1 2 3; do if true; then
	foo
	bar
fi; done
```

The correct behavior should be:

```
$ echo 'for i in 1 2 3; do if true; then foo; bar; fi; done' | shfmt
for i in 1 2 3; do
	if true; then
		foo
		bar
	fi
done
```

deepakganesh78:
Fixes #720

A body containing one compound command was considered inline solely from the number and source positions of its direct statements. When that compound command formatted its own nested body across multiple lines, the outer body stayed inline, producing incorrect indentation.

Detect whether the sole nested compound command will itself expand to multiple lines and force the containing body onto separate lines. Single-line compound commands remain compact.

Tests:
- `go test ./syntax -run '^TestPrintTable$' -count=1`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1487,6 +1487,10 @@ func (p *Printer) nestedStmts(stmts []*Stmt, last []Comment, closing Pos) {
 		// Force a newline if we find:
 		//     { stmt; stmt; }
 		p.wantNewline = true
+	case stmtsWillBeMultiline(stmts):
+		// Force a newline if the only statement expands while printing:
+		//     { if foo; then stmt; stmt; fi; }
+		p.wantNewline = true
 	case closing.Line() > p.line && len(stmts) > 0 &&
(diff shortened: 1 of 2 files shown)


mvdan:

PR #1369, opened 2026-08-07
expand: give printf numbers a leading-quote character value

sueun-dev:
POSIX printf takes a numeric argument beginning with a single or double
quote as the value of the character right after the quote, so
`printf %d "'A"` prints 65. We passed the argument straight to
strconv.ParseInt, so any such argument became 0.

formatNum now handles the quote prefix for the numeric verbs (d, i, u, o,
x), taking the value of the first rune after the quote and ignoring the
rest:

    $ printf '%d %x %o\n' "'A" "'A" "'A"
    65 41 101

That matches bash, dash, and ksh for ASCII. I decode a rune rather than a
byte, matching the "// TODO: use a rune?" note by the %c test; the two
only differ for multi-byte input, where those shells disagree with each
other anyway.

Added interp printf cases for both quote forms, '0 (48, not 0), a
trailing character after the first, and a lone quote (still 0).

### DIFF
--- a/expand/expand.go
+++ b/expand/expand.go
@@ -378,7 +378,7 @@ func formatInto(sb *strings.Builder, format string, args []string) (int, error) 
 						return 0, err
 					}
 				} else if c != 's' {
-					n, _ := strconv.ParseInt(arg, 0, 0)
+					n := formatNum(arg)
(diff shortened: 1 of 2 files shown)


mvdan:

PR #1374, opened 2026-08-09
pattern: support mixed POSIX character classes

roeyazroel:
## Summary

Teach `pattern.Regexp` to recognize POSIX character classes as elements inside a larger bracket expression.

This fixes valid patterns such as `*[![:space:]]*`, while preserving surrounding negation, literals, ranges, and multiple POSIX classes.

## Root cause

The bracket parser previously treated the `]` in `[:space:]` as the end of the outer bracket expression. For this input:

```text
*[![:space:]]*
```

it generated the invalid Go regular expression:

```text
(?s)^.*[^[:space:]\].*$
```

`pattern.Regexp` documents that successful output is safe to pass to `regexp.MustCompile`. The interpreter does exactly that, so the invalid expression causes a process-wide panic.

One real-world reproducer is NVM 0.40.6. Its alias reader uses this case pattern to ignore whitespace-only alias lines, and `nvm use --lts` crashes applications embedding the interpreter.

## Fix

- Parse POSIX classes within the surrounding bracket-expression loop.
- Consume the inner `:]` without closing the outer expression.
- Preserve validation for malformed character classes and unsupported collating/equivalence forms.
- Cover negation, literals, ranges, multiple classes, malformed inputs, and matching semantics.

## Verification

- `go test ./pattern -count=1`
- `go test ./...`
- `git diff --check`

The full suite passes: 9,061 tests across 10 packages.


mvdan:
Thanks - this looks mostly right, but it seems like significantly more code than is needed. I'll do a pass over this and update the PR myself.

mvdan:
As mentioned, I reworked the patch a bit before merging and force-pushed, keeping your authorship. The end result is the same, but the diff is roughly half the size:

- Instead of pre-scanning bracket expressions with `bracketEnd`, which duplicated the bracket grammar that `regexpNext` then re-parsed, the main loop now does a single pass: it builds the expression in a temporary buffer, tracks whether a slash was seen, and defers errors until we know the bracket isn't literal.
- That also let me merge `posixClassLen` and `bracketElementLen` back into `charClass`, which now returns the element length alongside an error.
- I trimmed the new test cases down to the ones covering distinct code paths, and squashed the result into a single commit.

roeyazroel:
Thank you! @mvdan

PR #1377, opened 2026-08-10
pattern: treat an unmatched "[" as a literal

### LINKED ISSUE #1372: pattern: unmatched opening bracket should be treated as a literal

The public `pattern.Regexp` API rejects an unmatched opening bracket, while Bash treats it as a literal character in shell pattern positions.

To reproduce, create an empty directory and save this as `main.go`:

```go
package main

import (
	"fmt"

	"mvdan.cc/sh/v3/pattern"
)

func main() {
	regexp, err := pattern.Regexp("[", pattern.EntireString)
	fmt.Printf("regexp=%q error=%T %v\n", regexp, err, err)
}
```

Run it with:

```console
go get mvdan.cc/sh/v3@v3.13.1
go run .
```

Actual output:

```text
regexp="" error=*pattern.SyntaxError [ was not matched with a closing ]
```

The corresponding Bash pattern matches the literal opening bracket:

```bash
case "[" in
	[) echo match ;;
	*) echo miss ;;
esac
```

Bash output:

```text
match
```

Would it be appropriate for shell-pattern compilation to treat an unmatched `[` as a literal, while retaining errors for malformed constructs that Bash also rejects?

Pastalikek65:
Fixes #1372

## Summary

`pattern.Regexp` rejected an unmatched opening bracket `[` with a `SyntaxError`, but Bash treats a `[` that never gets its closing `]` as a **literal character**, reparsing the rest of the pattern normally.

The impact is real, not just cosmetic: the interp's `case`-statement matching (`internal.ExtendedPatternMatcher` → `pattern.Regexp`) discards the parse error, so a valid Bash pattern like `case "[" in [) ...` silently never matched.

## Root cause

In `regexpNext` (`pattern/pattern.go`), the `case '[':` block returned a `SyntaxError` at four sites when the bracket expression never saw a closing `]`:

- right after `[`
- after `[!` / `[^`
- after `[]`
- in the general class loop at EOF

The Filenames mode already had the correct behaviour for one sub-case (a bracket expression containing `/`): emit `[` literally, rewind the lexer, and reparse the rest.

## Fix

Replace all four error returns with a single `reparseBracket` helper that does what the Filenames fallback did:

```go
func reparseBracket(sb *strings.Builder, sl *stringLexer, lit int) {
	sl.i = lit
	sb.WriteString(`\[`)
}
```

The unmatched `[` is emitted as a literal and the lexer rewinds so the remainder of the pattern is parsed normally — exactly what Bash does:

| pattern | before | after |
|---|---|---|
| `[` | error | `(?s)\[` (matches `[`) |
| `[a` | error | `(?s)\[a` |
| `[a*` | error | `(?s)\[a.*` (literal `[` + glob) |
| `[]` | error | `(?s)\[\]` |
| `[^]` | error | `(?s)\[\^\]` |
| `[[:digit` | error | `(?s)\[\[:digit` |

Closed-but-invalid brackets are unchanged: `[z-a]` still reports `invalid range`, and `[[:wrong:]]` / `[[=x=]]` / `[[.x.]]` still report `charClass invalid`. A trailing `\` in `[\` still reports `\ at end of pattern`, consistent with the existing rule for a trailing backslash.

## Tests

- Replaced the nine `[ was not matched with a closing ]` error expectations with literal `want` + `mustMatch`/`mustNotMatch` cases, including the `[a*` glob-reparse case.
- Verified against real Bash: `case "[" in [)` prints `match-literal-[` in both Bash and the fixed interp (previously the interp printed `miss`).

## Verification

- `go test ./pattern/ ./syntax/ ./expand/` pass
- `gofmt -l` and `git diff --check` clean; `go vet ./pattern/` clean

Note: `TestRunnerRun`'s `chmod 0xxx; cd` cases fail in this container because tests run as root (pre-existing, also fail on `master`); the race tests cannot run here due to a ThreadSanitizer VMA restriction (also pre-existing).


mvdan:

PR #1378, opened 2026-08-10
expand: short-circuit && and || in arithmetic expressions

### LINKED ISSUE #1371: expand: arithmetic logical operators evaluate the right operand eagerly

The public `expand.Arithm` API evaluates the right operand of arithmetic `&&` and `||` even when the left operand already determines the result.

To reproduce, create an empty directory and save this as `main.go`:

```go
package main

import (
	"fmt"
	"strings"

	"mvdan.cc/sh/v3/expand"
	"mvdan.cc/sh/v3/syntax"
)

type environment map[string]string

func (e environment) Get(name string) expand.Variable {
	value, ok := e[name]
	if !ok {
		return expand.Variable{}
	}
	return expand.Variable{Set: true, Kind: expand.String, Str: value}
}

func (e environment) Each(yield func(string, expand.Variable) bool) {
	for name := range e {
		if !yield(name, e.Get(name)) {
			return
		}
	}
}

func (e environment) Set(name string, variable expand.Variable) error {
	e[name] = variable.String()
	return nil
}

func evaluate(env environment, source string) {
	expr, err := syntax.NewParser().Arithmetic(strings.NewReader(source))
	if err != nil {
		panic(err)
	}

	value, err := expand.Arithm(&expand.Config{Env: env}, expr)
	fmt.Printf("%s: value=%d x=%s error=%v\n", source, value, env["x"], err)
}

func main() {
	env := environment{"x": "0"}
	evaluate(env, "0 && (x = 1)")

	env["x"] = "0"
	evaluate(env, "1 || (x = 1)")
}
```

Run it with:

```console
go get mvdan.cc/sh/v3@v3.13.1
go run .
```

Actual output:

```text
0 && (x = 1): value=0 x=1 error=<nil>
1 || (x = 1): value=1 x=1 error=<nil>
```

Bash short-circuits both expressions, so the equivalent commands leave `x=0`:

```bash
x=0
: $((0 && (x = 1)))
printf 'and: x=%s\n' "$x"

x=0
: $((1 || (x = 1)))
printf 'or: x=%s\n' "$x"
```

Expected output:

```text
and: x=0
or: x=0
```

Would a change that evaluates arithmetic `&&` and `||` lazily in `expand.Arithm`, with tests covering skipped assignments and increments, be accepted?

Pastalikek65:
Fixes #1371

## Summary

`expand.Arithm` evaluated both operands of arithmetic `&&` and `||` eagerly, so `0 && (x = 1)` performed the assignment `x = 1` even though the result of the whole expression was already known to be false. Bash short-circuits: `0 && (x = 1)` and `1 || (x = 1)` leave `x` unchanged.

## Fix

In `Arithm` (`expand/arith.go`), after evaluating the left operand and before evaluating the right:

- `AndArit` (`&&`): if `left == 0`, return `0` immediately.
- `OrArit` (`||`): if `left != 0`, return `1` immediately.

This mirrors the existing conditional evaluation in the ternary `TernQuest` branch. Because assignments in arithmetic write through `Config.Env.Set`, skipping the right-hand side correctly avoids the side effect.

## Tests

- New `expand/arith_test.go`: `TestArithmShortCircuit` table test with a minimal `WriteEnviron` map, asserting both the returned value and the resulting value of `x` for `&&`/`||` (including that the non-short-circuiting `0 + (x = 5)` still evaluates both sides).
- Three interp `runTests` entries in `interp/interp_test.go` covering `0 &&`, `1 ||`, and `1 &&`.

## Verification

- `go test ./expand/ ./pattern/ ./syntax/` pass.
- Cross-checked against real Bash: the fixed interp and Bash produce identical output for `0 &&`, `1 ||`, `1 &&`, `0 ||`, and `2 &&` with assignments.
- `gofmt -l`, `git diff --check` clean; `go vet` clean.

Pre-existing failures, unrelated to this change (also fail on `master`): the `TestRunnerRun` cases for `[` builtin / `chmod 0xxx; cd` running as root, and `go test -race` (ThreadSanitizer VMA restriction in this container).


mvdan:

PR #1379, opened 2026-08-10
syntax: parse ${ command; } inside double quotes (#1368)

### LINKED ISSUE #1368: syntax: support alternate command substitution of Bash 5.3 inside double quotes
PR https://github.com/mvdan/sh/pull/1178 added the initial support for the "subshell-less" command substitutions introduced in Bash 5.3: `${ command; }`  and `${| command; }`.

However, when these command substitutions are put inside double quotes, they cause errors.

```bash
#!/usr/bin/env bash
# test.sh
x="${ echo 1;}"
echo "$x"
```

```
$ shfmt test.sh
test.sh:4:9: reached EOF without closing quote `"`
```

```bash
#!/usr/bin/env bash
# test2.sh
x="${| REPLY=1;}"
echo "$x"
```

```
$ shfmt test.sh
test2.sh:4:9: reached EOF without closing quote `"`
```

Both scripts are valid and print `1` to the output.

Pastalikek65:
Fixes #1368.

In zsh, `${ command; }` works inside double quotes (e.g. `"${ echo hi; }"`). In mksh too (Debian #700161), and bash 4.4+ (since bash 4.4, ${ var; } is an alternative command substitution syntax). mksh/bash restrict it to inside double quotes; POSIX doesn't have it.

Currently the syntax parser fails on `${ command; }` inside double quotes with a hard error. This change parses it as an alternative command substitution.


Pastalikek65:
Closing this myself — understood the project isn't interested in drive-by patches. Apologies for the noise.

PR #1380, opened 2026-08-10
syntax: preserve trailing backslashes at EOF

### LINKED ISSUE #1354: syntax: printer drops a trailing backslash at EOF, making shfmt non-idempotent

Boulea7:
Fixes #1354.

An odd run of unquoted backslashes at EOF is literal, but the printer's final newline turned its last backslash into a line continuation. Pair the final backslash before writing that newline so formatting preserves the argument value and reaches a fixed point.

Tests cover all five parser variants with 1–4 trailing backslashes, quoted and unquoted, with and without a source newline (80 matrix cases). I also ran the full test suite with the race detector, 386 tests, and the CI cross-builds for plan9/amd64 and js/wasm.

Developed with AI assistance and independently reviewed; all changes and test results were manually verified.

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -123,6 +123,7 @@ func fullProg(v any) *File {
 
 type fileTestCase struct {
 	inputs []string // input sources; the first is the canonical formatting
+	print  string   // optional canonical formatting when it differs from inputs[0]
 
 	// Each language in [langResolvedVariants] has an entry:
 	// - nil:    nothing to test
@@ -132,6 +133,10 @@ type fileTestCase struct {
 
 	// The real shells where testing the input succeeds or fails in the opposite way.
 	flipConfirmSet LangVariant
+}
+
+func printAs(want string) func(*fileTestCase) {
+	return func(c *fileTestCase) { c.print = want }
 }
 
 func flipConfirm2(langSet LangVariant) func(*fileTestCase) {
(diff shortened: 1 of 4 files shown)


mvdan:

Boulea7:
Thanks for clarifying. I understand and will respect the project’s contribution preference.

PR #1381, opened 2026-08-14
interp: implement the help builtin, plus times

0pcom:
Adds three builtins the interpreter reported as unimplemented.

**`help`** — bash's own format: `help` lists the builtins in columns, `help NAME` prints that builtin's usage and description, `help -s NAME` just the usage line, and an unknown name exits 1 with `help: no help topics match`, as bash does. The table lives in `interp/help.go` so entries sit next to each other and stay comparable.

**`times`** — reports the shell's user/system time, then the children's, in the `0m0.000s` form bash prints.

**`umask`** — reads and sets the mask, including `-S` for the symbolic form.

`interp/help_test.go` covers the listing, per-topic output, `-s`, and the unknown-topic exit status.

Written while building an interactive shell on top of this package, where a missing `help` is one of the first things a user notices.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -155,6 +155,8 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 	case ":", "true":
 	case "false":
 		exit.code = 1
+	case "help":
+		return r.runHelp(args)
 	case "exit":
 		switch len(args) {
 		case 0:
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -157,6 +157,14 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		exit.code = 1
 	case "help":
 		return r.runHelp(args)
+	case "times":
+		// js/wasm has no per-process CPU accounting at all, so report zeros
+		// in bash's format — shell user/sys, then children user/sys — rather
+		// than fail.
+		// TODO: report real times on platforms that can, via os/exec's
+		// ProcessState or syscall.Getrusage. Not urgent; nothing depends on
+		// the values being non-zero.
+		r.out("0m0.000s 0m0.000s\n0m0.000s 0m0.000s\n")
 	case "exit":
 		switch len(args) {
 		case 0:
--- a/interp/help.go
+++ b/interp/help.go
@@ -1,0 +1,214 @@
+// Copyright (c) 2017, Daniel Martí <mvdan@mvdan.cc>
+// See LICENSE for licensing information
+
+package interp
+
+import (
+	"maps"
+	"path"
+	"slices"
+	"strings"
+)
+
+// The `help` builtin, following bash: `help` lists every builtin's synopsis in
+// two columns, `help NAME` prints the synopsis and description of one, and
+// NAME may be a glob (`help ex*`). -d, -s and -m select short, synopsis-only
+// and man-page-ish output.
+//
+// A builtin the interpreter recognizes but does not implement is marked with a
+// star, the way bash marks disabled builtins — so `help` doubles as an honest
+// statement of what this shell can do.
+
+// builtinHelp is the synopsis and description of one builtin.
+type builtinHelp struct {
+	synopsis string
+	desc     string
+}
+
+// helpTable is keyed by builtin name. Synopses follow bash 5's wording so the
+// output is familiar; descriptions are ours, and say where this shell differs.
+var helpTable = map[string]builtinHelp{
+	":":       {": ", "Null command; expands arguments and returns success."},
+	".":       {". filename [arguments]", "Execute commands from a file in the current shell."},
+	"[":       {"[ arg... ]", "Evaluate a conditional expression; a synonym for `test' requiring a closing `]'."},
+	"alias":   {"alias [-p] [name[=value] ... ]", "Define or display aliases. With no arguments, print the alias list."},
+	"bg":      {"bg [job_spec ...]", "Resume jobs in the background."},
+	"bind":    {"bind [-lpsvPSVX] [-m keymap] [keyseq:readline-function]", "Set readline key bindings and variables."},
+	"break":   {"break [n]", "Exit the enclosing for, while or until loop; with n, exit n levels."},
+	"builtin": {"builtin [shell-builtin [arg ...]]", "Run a shell builtin, bypassing a function of the same name."},
+	"caller":  {"caller [expr]", "Return the context of the current subroutine call."},
+	"cd":      {"cd [-L|[-P [-e]]] [dir]", "Change the shell working directory."},
+	"command": {"command [-pVv] command [arg ...]", "Run a command, bypassing shell function lookup."},
+	"compgen": {"compgen [-abcdefgjksuv] [-o option] [word]", "Display possible completions for a word."},
+	"complete": {"complete [-abcdefgjksuv] [-pr] [-o option] [name ...]",
+		"Specify how arguments are to be completed by readline."},
+	"compopt":  {"compopt [-o|+o option] [name ...]", "Modify or display completion options."},
+	"continue": {"continue [n]", "Resume the next iteration of the enclosing loop; with n, of the nth enclosing loop."},
+	"declare": {"declare [-aAfFgiIlnrtux] [name[=value] ...]",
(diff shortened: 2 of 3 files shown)


interp/help.go
```diff
@@ -0,0 +1,244 @@
...
+
+// helpNames returns every documented builtin name, sorted.
+func helpNames() []string {
+	names := make([]string, 0, len(helpTable))
+	for name := range helpTable {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	return names
+}
```
mvdan: this can be just `slices.Sorted(maps.Keys(helpTable))`. inline.

interp/help_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	f, err := syntax.NewParser().Parse(strings.NewReader(src), "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	_ = r.Run(context.Background(), f)
+	return out.String()
+}
+
+func TestHelpBuiltin(t *testing.T) {
```
mvdan: we avoid unit tests where possible. just add a few test cases to the existing table-driven test.

interp/help.go
```diff
@@ -0,0 +1,244 @@
...
+			r.errf("help: %s: invalid option\n", rest[0])
+			r.errf("help: usage: %s\n", helpTable["help"].synopsis)
+			return exitStatus{code: 2}
+		}
+		rest = rest[1:]
+	}
+patterns:
+
+	if len(rest) == 0 {
+		r.out("A shell built on mvdan.cc/sh. These shell commands are defined internally.\n")
```
mvdan: if you're going to name the module, at least use the full path, i.e. mvdan.cc/sh/v3.

interp/builtin.go
```diff
@@ -155,6 +155,41 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args
 	case ":", "true":
 	case "false":
 		exit.code = 1
+	case "help":
+		return r.runHelp(args)
+	case "times":
+		// No per-process CPU accounting on the targets this fork serves
+		// (js/wasm has none at all), so report zeros in bash's format
+		// rather than fail: shell user/sys, then children user/sys.
```
mvdan: this comment refers to a fork, which is not relevant at all. you should just summarize that this is not doable on js/wasm. and perhaps add a TODO to actually implement it on other platforms in the future, which is not urgent.

interp/help.go
```diff
@@ -0,0 +1,244 @@
...
+				r.out("\n")
+			}
+		}
+	}
+	return exit
+}
+
+// umaskSymbolic renders a mask the way `umask -S` does: the permissions that
+// remain, not the ones masked off.
+func umaskSymbolic(mask uint32) string {
```
mvdan: why is a umask function in help.go?

mvdan (commented):
SGTM but split each of the builtins into a separate commit.

0pcom:
All addressed, and split into three commits — `help`, `times`, `umask`.

- `slices.Sorted(maps.Keys(helpTable))` inlined at both call sites; `helpNames` and the `sort` import are gone.
- `help_test.go` deleted, cases folded into `runTests`.
- Full module path in the `help` banner.
- The `times` comment no longer mentions a fork. It now says js/wasm has no per-process CPU accounting, with a TODO to report real times where the platform can.
- `umaskSymbolic` moved out of `help.go` into `builtin.go`, next to the `umask` case it serves. It was only there because that is where I happened to write it.

On the tests: folding them in was more useful than I expected, because `TestRunnerRunConfirm` runs every case against real bash — which my standalone test did not. Three umask cases (`umask`, `umask -S`, `umask 077; umask -S`) now pass against bash, so that behaviour is confirmed rather than merely asserted.

Where this shell deliberately differs I used ` #IGNORE` with a reason, following the existing entries: bash suggests `man`/`info` on a help miss, prints a keyword header for a glob, prefixes errors with `bash: line N:`, and implements `jobs`. The starred-unimplemented output is ours by design, so it is ignored there too.

`go test ./interp/` passes, including the bash confirmation pass.

interp/api.go
```diff
@@ -58,6 +58,11 @@ type Runner struct {
 	// tempDir is either $TMPDIR from [Runner.Env], or [os.TempDir].
 	tempDir string
 
+	// umask is the file mode creation mask reported and set by the umask
+	// builtin. It is bookkeeping only: this interpreter creates files
+	// through the open handler, which owns permissions.
```
mvdan: not entirely sure this is correct. we have a default open handler, which should surely use this umask?

mvdan (approved):

0pcom:
Filed #1401 covering the unimplemented builtins, per the contributing guidelines, to accompany this and #1382.

0pcom:
You were right, and it was a real bug rather than just a wrong comment.

Redirections hardcoded `0o644`. That is `0666 &^ 022` — the result for the *default* umask — so it looked correct while silently ignoring whatever the script had set. `umask 077; echo hi >f` produced a 0644 file where bash gives 0600.

Redirections now mask `0666` with the runner's umask, as bash does. No behaviour change at the default: `0666 &^ 022` is still `0644`.

Added `TestRunnerUmask` covering 022, 077, 002 and 000. I checked it fails without the fix rather than assuming — with the old hardcoded value the 077 case reports `file mode is 0644, want no bits beyond 0600`.

It is a Go test rather than a `runTests` case, which I know cuts against what you said on the other file. Printing a file mode needs an external command and interp has no builtin that reports one, so the table cannot express it. Happy to move it if you can see a way.

Two things I decided rather than guessed, so correct me if either is wrong:

- The check is `got &^ want != 0` rather than equality, because on a real OS the *process* umask applies on top — the kernel enforces it in `os.OpenFile`, not `DefaultOpenHandler`. In bash those are the same value; here they are separate, so the test asserts no extra bits rather than an exact mode.
- The field doc now says the same thing, since a custom open handler over an in-memory filesystem gets no kernel masking at all and has to honour this itself. That was the case I had in mind when I wrongly called it bookkeeping.

0pcom:
Five of the six are already pushed; the sixth is a design question I would rather you decide than guess at.

Done: `slices.Sorted(maps.Keys(helpTable))` inlined at both sites; `help_test.go` deleted and the cases folded into the table-driven test in `interp_test.go`; the module named in full as `mvdan.cc/sh/v3`; the `times` comment rewritten to summarize that per-process CPU accounting is not available on js/wasm, with a TODO to report real values elsewhere, and no mention of a fork; and no umask function in `help.go` — only its one-line help entry, alongside every other builtin.

**On the umask and the default open handler.** You were right to be unsure. I measured it:

```
$ umask                      # process umask
0022
$ gosh -c 'umask 000; echo x > f000; umask 077; echo x > f077'
-rw-r--r--  f000     # bash gives 0666 here
-rw-------  f077
```

So it can only ever make permissions tighter than the process umask, never looser. That is the real gap.

The reason the mask is applied at the redirection site rather than inside `DefaultOpenHandler` is that a custom open handler — an in-memory filesystem, say — would otherwise never get it applied at all, since `umask` is unexported. But that has a cost I should have called out: `perm` in `OpenHandlerFunc` nominally means the pre-umask mode, the same as `os.OpenFile`, and we are handing it a post-umask value. A custom handler that applies a umask itself would double-apply.

What we cannot do is what bash does — bash's umask *is* the process umask, and calling `syscall.Umask` from a library would reach out and change the host program's global state.

Three ways I can see:

1. Leave it, and document on `OpenHandlerFunc` that `perm` arrives already masked.
2. Pass the raw `0o666`, apply the mask in `DefaultOpenHandler`, and expose the umask on `HandlerContext` so custom handlers can honour it. Cleanest semantics, but it is an API addition and therefore yours to approve.
3. Drop `umask` from this PR so `help` and `times` can land, and take it up separately.

Happy to do any of them — 3 if you would rather not decide the API question under a PR that is mostly about `help`.

mvdan:
> Drop `umask` from this PR so `help` and `times` can land, and take it up separately.

please do this, because `umask` feels subtle and orthogonal to the others.

0pcom:
Done — `umask` is dropped, so this is `help` and `times` only, and the diff no longer touches `interp/api.go` or `interp/runner.go`.

I will open `umask` separately once this lands.

PR #1382, opened 2026-08-14
interp: implement the job control builtins (jobs, kill, disown, fg, bg)

0pcom:
A job here is a goroutine running a subshell rather than a process, so the builtins are shaped by what that can honestly mean: `jobs` lists them, `kill` cancels a job's context (which is what any fatal signal would have amounted to), `disown` drops it from the table and out of a bare `wait`, `fg` waits for the job since there is no terminal to hand it, and `bg` reports on a job that is already running. Nothing is ever stopped, so SIGSTOP/SIGCONT are rejected rather than faked. This is also what makes them work on js/wasm, where there are neither processes nor signals.

The second commit detaches background jobs from the statement's context, because a job outlives the command line that started it: an interactive shell cancels a line's context when the line is done, and in bash the interrupt that ends a foreground job leaves the background ones running. What ends a job is `kill`, the shell exiting, or the new `Runner.StopJobs`.

`help` gains a small notes table for builtins that cannot mean quite the same thing without processes or a controlling terminal.

Split out of the earlier version of this PR as requested: umask is #1411 and enable/compgen/history/logout are #1412, each independent against master.

Part of #1401.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -302,6 +302,36 @@ type bgProc struct {
 	// the process ID when the background statement started exactly one
 	// external program, and zero otherwise.
 	started chan int
+
+	// cmd is the source text of the backgrounded statement, as the jobs
+	// builtin prints it.
+	cmd string
+
+	// substitution marks the shells behind process substitutions, which bash
+	// does not list as jobs either. They have no job number, cannot be named
+	// by a job spec, and a bare wait does not wait for them.
+	//
+	// A separate field rather than testing cmd == "": an empty command string
+	// is a plausible thing to have for other reasons, and every place that
+	// asked "is this a real job" was really asking this.
+	substitution bool
+
+	// cancel stops the background shell. A job here is a goroutine rather
+	// than an operating system process, so the kill builtin cancels its
+	// context instead of sending a signal.
+	cancel context.CancelFunc
+
+	// signal is the name of the signal kill delivered, so that jobs can
+	// report Terminated rather than Done.
+	signal string
+
+	// disowned jobs are hidden from jobs and are not waited for by a bare
+	// wait, as after bash's disown.
+	disowned bool
+
+	// notified records that this job's completion has already been reported
+	// by `jobs -n`.
+	notified bool
 }
 
 // newBgProc returns a background job with the next fake PID.
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -162,6 +162,16 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		exit.code = 1
 	case "help":
 		return r.runHelp(args)
+	case "jobs":
+		return r.runJobs(args)
+	case "kill":
+		return r.runKill(args)
+	case "disown":
+		return r.runDisown(args)
+	case "fg":
+		return r.runFg(ctx, args)
+	case "bg":
+		return r.runBg(args)
 	case "times":
 		// js/wasm has no per-process CPU accounting at all, so report zeros
 		// in bash's format — shell user/sys, then children user/sys — rather
@@ -349,6 +359,9 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		if len(args) == 0 {
 			// Note that "wait" without arguments always returns exit status zero.
 			for _, bg := range r.bgProcs {
+				if bg.disowned {
+					continue
+				}
 				<-bg.done
 			}
 			break
--- a/interp/handler.go
+++ b/interp/handler.go
@@ -108,6 +108,12 @@ type HandlerContext struct {
 // Shell builtins touch on many internals of the Runner, after all.
 //
 // Returning a non-nil error will halt the [Runner] and will be returned via the API.
+//
+// Note that this is the only handler which sees the kill builtin. Given a PID
+// that is not one of the runner's own jobs, kill signals that process on the
+// host directly, without going through [ExecHandlerFunc] as `/bin/kill` once
+// would have. An embedder sandboxing what a script may do to the machine has
+// to intercept it here.
 type CallHandlerFunc func(ctx context.Context, args []string) ([]string, error)
 
 // TODO(v4): consistently treat handler errors as non-fatal by default,
--- a/interp/help.go
+++ b/interp/help.go
@@ -25,6 +25,17 @@ type builtinHelp struct {
 	desc     string
 }
 
+// helpNotes says how a builtin differs from bash's, for the ones that cannot
+// mean quite the same thing without processes or a controlling terminal.
+// `help NAME` prints the note after the description.
+var helpNotes = map[string]string{
+	"bg":    "background jobs already run in the background here, so this reports on the job rather than resuming it",
+	"fg":    "there is no controlling terminal here, so this waits for the job and returns its exit status",
+	"jobs":  "nothing is ever stopped without a controlling terminal, so -s lists nothing",
+	"kill":  "a job is a goroutine, not a process: terminating signals cancel it, and stop/continue are rejected; a PID that is not a job is signalled like any process",
+	"times": "there is no per-process CPU accounting on every target this shell runs on, so the times are zero",
+}
+
 // helpTable is keyed by builtin name. Synopses follow bash 5's wording so the
 // output is familiar; descriptions are ours, and say where this shell differs.
 var helpTable = map[string]builtinHelp{
@@ -102,10 +113,9 @@ var helpTable = map[string]builtinHelp{
 // helpUnsupported are recognized by this shell but not implemented; `help`
 // stars them the way bash stars disabled builtins.
 var helpUnsupported = map[string]bool{
-	"bg": true, "bind": true, "caller": true, "compgen": true, "complete": true,
-	"compopt": true, "disown": true, "enable": true, "fc": true, "fg": true,
-	"history": true, "jobs": true, "kill": true, "logout": true, "newgrp": true,
-	"suspend": true, "ulimit": true,
+	"bind": true, "caller": true, "compgen": true, "complete": true,
+	"compopt": true, "enable": true, "fc": true, "history": true,
+	"logout": true, "newgrp": true, "suspend": true, "ulimit": true,
 }
 
 // helpMatch returns the documented names matching a bash-style pattern. An
@@ -205,6 +215,8 @@ patterns:
 				r.outf("    %s\n", h.desc)
 				if helpUnsupported[name] {
 					r.out("    (recognized by this shell but not implemented)\n")
+				} else if note := helpNotes[name]; note != "" {
+					r.outf("    (%s)\n", note)
 				}
 				r.out("\n")
 			}
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -241,8 +241,25 @@ var runTests = []runTest{
 	{"help -q", "help: -q: invalid option\nhelp: usage: help [-dms] [pattern ...]\nexit status 2 #IGNORE bash prefixes its errors with `bash: line N:'"},
 	// a builtin we recognize but do not implement is starred, so that help is
 	// an honest statement of what this shell can do
-	{"help -s jobs", "jobs: jobs [-lnprs] [jobspec ...]\n #IGNORE bash implements jobs, and its synopsis differs"},
-	{"help jobs", "jobs: jobs [-lnprs] [jobspec ...]\n    Display status of jobs.\n    (recognized by this shell but not implemented)\n\n #IGNORE bash implements jobs"},
+	{"help -s jobs", "jobs: jobs [-lnprs] [jobspec ...]\n #IGNORE bash's synopsis differs"},
+	{"help -s ulimit", "ulimit: ulimit [-SHabcdefiklmnpqrstuvxPRT] [limit]\n #IGNORE bash implements ulimit"},
+
+	// job control builtins; the cases needing a job that is still running are
+	// in jobs_test.go, since here the exec middleware backgrounds goroutines
+	// that cancellation cannot interrupt
+	{"jobs", ""},
+	{"kill -l 9", "KILL\n"},
+	{"kill -l TERM", "15\n"},
+	{"kill -l NOPE", "kill: NOPE: invalid signal specification\nexit status 1 #IGNORE bash prefixes its errors with `bash: line N:'"},
+	{"kill", "kill: usage: kill [-s sigspec | -n signum] pid | jobspec ...\nexit status 2 #IGNORE bash prefixes its errors with `bash: line N:'"},
+	{"kill %9", "kill: %9: no such job\nexit status 1 #IGNORE bash prefixes its errors with `bash: line N:'"},
+	{"jobs -Z", "jobs: -Z: invalid option\njobs: usage: jobs [-lnprs] [jobspec ...]\nexit status 2 #IGNORE bash prefixes its errors with `bash: line N:'"},
+	{"sleep 0 & wait; jobs", "[1]+  Done                    sleep 0\n #IGNORE bash formats the job listing differently"},
+	{"(exit 3) & wait; jobs", "[1]+  Exit 3                  (exit 3)\n #IGNORE bash formats the job listing differently"},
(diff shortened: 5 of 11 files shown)


0pcom:
Rebased on the reworked #1381 — this PR is stacked on it, so its first three commits are that one.

I applied your #1381 review here too, without waiting to be told twice:

- Three files carried a `Copyright (c) 2026, 0magnet fork authors` header. Replaced with the project's. Same class of thing as the `times` comment you flagged: fork context that has no business upstream.
- `helpNames` is gone, so the two call sites in `bashbuiltin.go` now use `slices.Sorted(maps.Keys(helpTable))` inline.
- The `help` test cases in `runTests` are updated for this PR's effect: implementing job control moves `jobs` from starred-unimplemented to implemented-with-a-note, so `ulimit` becomes the starred example instead.

One thing I did **not** do, and would rather ask than guess. `jobs_test.go` is still a unit test file. Your "we avoid unit tests where possible" applied cleanly to `help_test.go` — those were exact-output checks that belong in `runTests`, and folding them in was strictly better, since `TestRunnerRunConfirm` then validated them against real bash.

These do not convert as cleanly. Of the 41 cases, most assert with `strings.Contains` rather than exact output, because job state and the order two background shells finish in are not fixed — `sleep 30 & jobs` has to tolerate whatever else is on the line. And nearly every one would need ` #IGNORE`, since bash's job control genuinely differs here: a job is a goroutine, nothing is ever stopped, and there is no controlling terminal. A table where every entry is ignored seems worse than a unit test that says why.

So `jobs_test.go` keeps its own small `runSrc` helper, with a comment explaining the choice. Happy to convert the subset that is exact-output if you would prefer that split, or to convert the lot — say which.

`go test ./interp/` passes, bash confirmation included.

0pcom:
Filed #1401 covering the unimplemented builtins, per the contributing guidelines, to accompany this and #1381.

0pcom:
Rebased on current master. The interesting merge was with the new $! handling from master: bgProc now carries both the fake-PID/started-report fields from master and this PR's job-control fields, background statements get both the started-program report and the detached cancellable context, and the wait builtin uses master's lookupBgProc together with the interruptible await from this PR.

mvdan:
Thanks - please split this again, as it seems to me like job control has little to do with some of the non-trivial builtins you are adding here, like compgen and history. Ideally I don't want to be reviewing more than a few hundred lines per PR.

0pcom:
Done — split three ways, each independent against master:

- **this PR** is now job control only: `jobs`/`kill`/`disown`/`fg`/`bg` plus the detached-context commit (838 lines, ~340 of which are tests)
- **#1411** — `umask` (63 lines)
- **#1412** — `enable`, `compgen`, `history`, `logout` (549 lines, ~180 tests)

The one seam between them: `compgen -A job` and `-A signal` in #1412 yield nothing until this PR lands, noted there. Each branch passes `go test ./...` on its own.

mvdan:
Thanks for splitting it. Four things before I look closer:

**`kill $!` doesn't work.** Since the `$!` change on master, `$!` is the real PID whenever the background statement is exactly one external program, but `jobSpec` only knows `%n`, `%str` and the `gN` ids:

```
$ gosh -c 'sleep 30 & echo pid=$!; kill $!; echo st=$?; jobs'
pid=65252
kill: 65252: no such job
st=1
[1]+  Running                 sleep 30
```

That's the common form in scripts. Inverted, too: a bare `kill 1` cancels job 1, where bash signals PID 1. The id/PID forms should go through `lookupBgProc`, and bare integers should mean PIDs.

**Detaching from the caller's context changes behavior for every embedder, not just interactive ones.** Running `sleep 5 & echo started` under a 100ms context: master reaps the child on cancel, this branch leaves it running. Anyone bounding a script with a timeout now leaks processes and goroutines. And `Reset` clears `bgProcs`, so after a reset those jobs are both unreachable and unstoppable — their cancel funcs are gone. I'd want this gated on `Interactive(true)` or an explicit option, with `Reset` stopping jobs, and I'd rather review it as its own PR: it's the risky part, the builtins are the easy part.

**`StopJobs` hangs.** It waits on every `bgProc`, including the process-substitution shells, whose `cancel` is nil. `echo <(sleep 20) >/dev/null` then `StopJobs()` blocks for the full 20s, which isn't what its doc promises. It also takes no context, so a caller can't bound the wait — for frozen v3 API I'd want `StopJobs(ctx)`. No in-repo caller either; `cmd/gosh` is the motivating case and isn't updated.

**The tests leak.** `go test -run TestDisownFgBg ./interp` leaves four `sleep 30` processes alive after the binary exits — the disown cases never kill their jobs and nothing calls `StopJobs`. `runSrc` also shares one `strings.Builder` between stdout and stderr and reads it while detached job goroutines can still write; you already wrote `syncBuffer` for the last test, so `-race` only stays quiet because `sleep` prints nothing.

On the unit-test question: split it. The deterministic cases (`kill -l 9`, `kill -l TERM`, the invalid-option and invalid-signal errors, `sleep 0 & wait; jobs`, `(exit 4) & fg`, `disown; jobs`, `bg` on a finished job) have fixed output and belong in `runTests`; `#IGNORE` on most of them is fine. Keep a Go test only for what's genuinely concurrent.

0pcom:
All four addressed. The context detachment is now its own PR — #1418, stacked on this branch — and this PR is builtins-only again.

**`kill $!` and bare integers.** `jobSpec` now treats anything without a leading `%` as a PID, resolved backwards through the job ids the way `lookupBgProc` does — the real PID when the statement was exactly one external program, `gN` otherwise — so `kill $!` works in both shapes, and a disowned job's PID still resolves, since bash's kill also still reaches a disowned job's process. Your repro:

```
$ gosh -c 'sleep 30 & echo pid=$!; kill $!; echo st=$?; wait; jobs'
pid=2740006
st=0
[1]+  Terminated              sleep 30
```

A bare integer that matches no job is signalled as a real process where the platform can (unix; an error on windows/plan9/js): before this PR those specs reached the system's kill program, and `kill $(cat pidfile)` shouldn't regress just because kill became a builtin. The inverted case:

```
$ gosh -c 'sleep 30 & kill 1; echo st=$?; jobs; kill %1; wait'
kill: (1) - operation not permitted
st=1
[1]+  Running                 sleep 30
```

`jobs -p` and `jobs -l` print the same ids $! reports, so they also show real PIDs now.

**Detachment** — moved to #1418, gated on `Interactive(true)`, with `Reset` stopping jobs first; details and tests there.

**`StopJobs`** — also #1418: it takes a context so the wait can be bounded, skips the process-substitution shells (`echo <(sleep 20) >/dev/null` no longer blocks it), and `cmd/gosh` calls it when the interactive loop exits.

**Tests** — the deterministic cases moved into `runTests` (`kill -l 9` and `kill -l TERM` confirm against bash; the rest carry `#IGNORE`), and the Go tests keep only genuinely concurrent state. Every script that starts a sleep now ends it — kill + wait, by PID for the disowned ones — and `runSrc` writes through a locked buffer. `go test -race ./interp/` passes with no sleep processes left behind.

mvdan:
`kill $!` and `kill 1` behave as you pasted. Non-interactive jobs die with the caller's context again — `sleep 5 &` under a 100ms context reaps the child in 0.1s where the old branch took 5.0s. `StopJobs(ctx)` returns immediately on my `echo <(sleep 20)` repro. `go test -race ./interp` passes with no leftover sleeps, the confirm suite runs 1404 subtests green including the three new cases without `#IGNORE`, and gofmt/vet plus windows/plan9/js/386/darwin builds are clean. The test split is the right shape now.

One real bug left, and this revision made it matter more: `signalNames` hardcodes Linux's numbering, and `killProcess` passes those numbers straight to `unix.Kill`. On darwin `SIGUSR1` is 30 and 10 is `SIGBUS`, so `kill -USR1 <pid>` on macOS sends SIGBUS to a real process. Same root cause makes `effectOf` switch on numbers (17–22), misclassifying signals off Linux, and `kill -l` print wrong names. Resolve names through the platform (`unix.SignalNum`) in `os_unix.go`; a fixed table only makes sense where there's nothing to ask.

Smaller, none blocking:

- `kill -STOP %1` is rejected as "no job control" while `kill -STOP <pid>` really stops a process — defensible, wants a comment.
- `kill -0` on a finished job succeeds; bash errors.
- `fg` returns 130 on cancellation where the rest of interp goes through `r.exit.fatal(ctx.Err())`.
- Finished jobs never leave the table, so `jobs` reprints `Done` forever and job numbers only grow; bash drops a completed job once reported.
- `bg.cmd == ""` doubles as "not a real job" in several places; an explicit field would read better.
- In #1418, `Reset` can now block — bounded by the exec handler's kill timeout, but new for a method with no context, so worth a doc line.
- `os_notunix.go` says "plan9 and windows"; it also covers js/wasm.

Finally, `kill` now signals arbitrary host PIDs from the interp process, which `ExecHandler` no longer sees. `CallHandler` does run for builtins, so embedders can still intercept it - worth naming that in the docs.

0pcom:
Signal names now resolve through the platform: `signalName`/`signalNum` in `os_unix.go` ask `x/sys/unix`, `effectOf` switches on the name, and `kill -l` walks the host's own numbering. The table survives only in `os_notunix.go`, where there is nothing to ask and the names never reach a kernel.

One correction to the diagnosis, because I went looking for the misclassification and it is not where either of us thought. On darwin — and on FreeBSD, OpenBSD, NetBSD and DragonFly — 17 through 22 are *exactly* the six job-control signals in a different order, so the old range was accidentally correct there. The platform it actually broke on is solaris, where 17 is USR2 and 19–22 are PWR, WINCH, URG and IO: five signals that should end a job, refused as "no job control". The `kill -USR1` fault you described is real and unaffected by this — that is the `killProcess` half, and 10 is BUS on darwin as you said.

**I cannot test darwin or solaris here**, only Linux. The new tests are written entirely in terms of names and mention no number, so they assert the same thing on every platform and would have failed against the old table — but they have only been *run* on Linux. `kill -l` output and the named lookups match bash here; `go test -race ./interp` is clean with no leftover sleeps.

The rest:

- `kill -0` on a finished job now fails with "no such job", as bash does.
- `fg` goes through `exit.fatal(ctx.Err())` on cancellation instead of returning 130.
- Process-substitution shells carry an explicit `substitution` field; every `cmd == ""` test was really asking that.
- `kill -STOP` on a job is still refused while a PID really stops — comment added saying why.
- `CallHandlerFunc` documents that it is the only handler which sees `kill`.
- `os_notunix.go` comments now mention js/wasm.

Two left undone, deliberately. Finished jobs still never leave the table: matching bash means reusing job numbers, which renumbers `%n` and destabilises `$!` for jobs already reported, and I would rather have your call on that than guess. And `Reset`'s new blocking behaviour is #1418's, so the doc line belongs there — I will add it in that PR.

0pcom:
CI caught that my `kill -0` change was wrong, and I have reverted it in 76ff8b6.

bash does report "no such job", but not for the reason I implemented. Its rule is about whether the job has been **reaped**, not whether it is still running — a finished job that nothing has reported still answers:

```
$ true & kill -0 $!; echo st=$?
st=0
$ true & wait; kill -0 %1; echo st=$?
bash: kill: %1: no such job
st=1
```

Testing `running()` therefore made the answer depend on whether the background job had got there first, which turned `true & kill -0 $!` — an existing case in `TestRunnerRun` — into a race. It passed here and on macos/1.26, and failed on ubuntu, windows and macos/1.27. My local full run passed too, which is exactly the false confidence a timing-dependent change gives you.

So this is not a separate smaller item: **`kill -0` cannot be right until a job can be reaped**, which makes it the same question as finished jobs never leaving the table. Both are waiting on your call about renumbering. I have left the old behaviour with a comment recording why.

The signal-name work is unaffected — that is the earlier commit and it was green on macos/1.26. Reverting only the `kill -0` hunk; everything else stands.

0pcom:
Green apart from one unrelated flake. On 76ff8b6: macos 1.26 and 1.27, ubuntu 1.26, windows 1.26 and alpine all pass; ubuntu 1.27 was cancelled by fail-fast rather than failing.

The one failure is `TestParseConfirm/bash/OK/363-0` and `364-0` on windows 1.27 — inputs `$"foo"` and `$"foo$"`, both `shell terminated by signal: context deadline exceeded`. That is `syntax/` shelling out to a real bash and timing out. Both commits together touch seven files, all under `interp/`, and `interp` passed on that same runner. I cannot rerun the job without admin rights on the repo.

Worth noting what the green macos runs bought: the signal tests are written in terms of names and had only ever been *run* on Linux, so this is the first time the platform lookup was actually exercised on darwin, where USR1 is 30 rather than 10. It passes there on both Go versions.

mvdan:
Two things on the current revision.

**`kill` can signal the host process itself.** `killProcess` hands the PID straight to `unix.Kill`, and kill(2) reads non-positive PIDs specially — 0 is the caller's process group, -1 is everything the user can reach:

```
$ setsid gosh -c 'echo before; kill 0; echo survived'
before
exit=143
```

For `gosh` that matches bash. For an embedded interpreter it means a script can SIGTERM the host application and its whole process group, and per the new `CallHandlerFunc` doc `ExecHandler` never sees it.

The comment justifying the bare-PID path isn't right either:

> before kill was a builtin these reached the system's kill program

On master they don't — `kill 99999999` gives `kill: unsupported builtin`, exit 2. `kill` went into `IsBuiltin` in 26d6d053, released in v3.13.1; it reached `/bin/kill` only up to v3.13.0. So this isn't preserving what scripts had last release, it's a new capability that bypasses the handler which used to gate it.

I don't have a strong opinion about what the right behavior here is, but at the very least, the comment is wrong.

**Process substitutions consume job numbers.** `byNumber` indexes `bgProcs` directly, so a substitution shell occupies a slot and shifts every `%n`:

```
$ gosh -c 'cat <(echo hi) >/dev/null; sleep 30 & kill %1'
kill: %1: no such job
```

`%2` works, but bash calls that job `%1`. The `substitution` field's own doc says these have no job number — they do, they're just skipped when listing. Job numbers should come off `jobIndexes()`. This affects spec resolution for `kill`, `fg`, `bg`, `disown` and `jobs %n`, not only the listing.

0pcom:
Both fixed in `40342216`.

**`kill` and non-positive PIDs.** `killProcess` now refuses `pid <= 0`, and the comment states that instead of the claim you corrected. One correction back: `26d6d053` shipped in `v3.13.0`, not `v3.13.1`, so `kill` has not reached `/bin/kill` in either released version — which only strengthens your point. I went with refusing rather than forwarding, since 0 and -1 name the host's process group rather than anything this runner started; the cost is that `gosh` no longer matches bash on `kill 0`. Easy to change if you would rather it went through a handler instead.

**Job numbers.** Numbering now skips process substitutions. I did not take it off `jobIndexes()` directly, because that also drops disowned jobs, and bash keeps the numbers of the jobs after a disowned one:

```
$ bash -c 'sleep 30 & sleep 30 & disown %1; jobs'
[2]+  Running                 sleep 30 &
```

So `jobNumber` counts non-substitution entries and still counts disowned ones. `jobs`, `kill`, `fg`, `bg` and `disown` all resolve through it, so the listing and spec resolution cannot disagree. Both of your cases now match bash.

Two regression tests cover them. Against the unfixed code the first one kills the test binary outright, which is the bug.

mvdan:
I'm not sure what to do with this feature, to be entirely honest. Every time I look at it, I uncover new bugs. We might need to carefully come up with a plan and design before looking at code. Or perhaps job control is not something that this library should aim to do, if it's not a good fit for Go or for the current v3 API.

1) `kill` seems to support both jobs and PIDs as arguments. A script running e.g. `kill 1` or `kill 321` could severely affect the host, in a way that could be highly dangerous to existing users of the interpreter. We don't promise a full sandbox or isolation, but it seems like a security issue to let a Go library kill the host machine.

I think we should only allow killing by PID - or killing PIDs which we have not started ourselves - behind an opt-in flag, like https://github.com/mvdan/sh/issues/171. But for now I probably would just leave that as a TODO and just forbid it entirely.

2) `{ { sleep 0.1; echo child-finished; } & } & wait; sleep 0.3` prints `child-finished` in master and in Bash, but not here.

3) `sleep 5 & p=$!; kill "$p"; wait "$p"` ends with a `context canceled` error rather than a standard shell exit status error.

PR #1384, opened 2026-08-16
interp: support js/wasm

0pcom:
Makes `interp` usable under `GOOS=js GOARCH=wasm`, where there are neither subprocesses nor OS pipes. Two independent changes.

**Abstract the runner's stdin.** `Runner.stdin` was an `*os.File` throughout. That type is load-bearing on OSes — it is the only thing a subprocess can inherit, and it is what supports cancellable reads via `SetReadDeadline` — but neither property exists on js/wasm, so the field cannot be constructed there at all.

It is now a small `stdinFile` interface (`io.ReadCloser` plus `SetReadDeadline`), with the OS path in `stdin_os.go` still backed by `*os.File` and unchanged in behaviour, and a js/wasm path in `stdin_js.go` backed by in-process pipes. `api.go` and `runner.go` go through the interface.

**Fall back to `/` when `Getwd` fails.** On js/wasm there is no working directory to report; `Runner` currently propagates the error and cannot be constructed. Seven lines.

Tested: `go test ./interp/` passes, and `GOOS=js GOARCH=wasm go build ./interp/` succeeds, which it did not before.

Happy to split these into two PRs if you would prefer — they share only the theme.

For context, these come out of running `interp` in a browser. There is one more portability patch in that work, an `os.SameFile` shim for TinyGo, which I have deliberately not included here: it is a workaround for a missing stdlib function rather than something js/wasm needs, and it seems more likely you would rather not carry a TinyGo build tag. Say if you would take it and I will send it separately.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -19,6 +19,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"strconv"
 	"time"
@@ -98,7 +99,7 @@ type Runner struct {
 	// accessHandler is a function responsible for checking file access. It must be non-nil.
 	accessHandler AccessHandlerFunc
 
-	stdin  *os.File // e.g. the read end of a pipe
+	stdin  stdinFile // e.g. the read end of a pipe
 	stdout io.Writer
 	stderr io.Writer
 
@@ -151,7 +152,7 @@ type Runner struct {
 	origDir    string
 	origParams []string
 	origOpts   runnerOpts
-	origStdin  *os.File
+	origStdin  stdinFile
 	origStdout io.Writer
 	origStderr io.Writer
 
@@ -326,6 +327,12 @@ func Dir(path string) RunnerOption {
 		if path == "" {
 			path, err := os.Getwd()
 			if err != nil {
+				if runtime.GOOS == "js" {
+					// js/wasm has no working directory; virtual
+					// filesystems set Runner.Dir themselves
+					r.Dir = "/"
+					return nil
+				}
 				return fmt.Errorf("could not get current dir: %w", err)
 			}
 			r.Dir = path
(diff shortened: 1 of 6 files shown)


mvdan:
Thanks. I have some thoughts on this. Rather than giving them as a review, I'll experiment with just pushing an incremental commit that shows them. You can then adapt the changes into your PR as you see fit.

Ideally, we end up with a single commit, and we also end up with some sort of regression test here. Because CI already runs `GOOS=js GOARCH=wasm go build ./...`, but seemingly that's not enough?

mvdan:
It looks like we could test on js/wasm with something like `PATH="$PATH:$(go env GOROOT)/lib/wasm" GOOS=js GOARCH=wasm go test ./interp`. I'll play with this on master after merging this PR.

As for my commit with the feedback, I can't push here, so I pushed to another branch: https://github.com/mvdan/sh/commit/321bf67568164b0f81f75033d4f1c55812def5ab

0pcom:
Thanks — pushing a commit rather than a review was much more useful than a list of remarks would have been. Adopted wholesale: the `stdinFile` alias on non-js, the interface moved into `stdin_js.go`, `stdin.go` gone, the `pipeWriter`/`jsPipeWriter`/`jsPipeReader` types dropped in favour of `io.WriteString` and the existing `jsReader`, `newOSPipe` renamed to `newPipe`, and your doc additions on `StdIO` and `HandlerContext.Stdin`. Squashed to a single commit.

On the regression test — I got part of the way and would rather report exactly where it stops than guess.

`GOOS=js GOARCH=wasm go test ./interp` did not previously get as far as running: the test binary would not compile, because `shortPathName` is defined in `unix_test.go` and `windows_test.go` and js is neither. The call site is already guarded by a runtime `GOOS == "windows"` check, so it just needed the third variant; that is in this commit, and `go test -c` for js/wasm now succeeds.

Running it then stops in `TestMain`, at `os.Executable()`:

```
panic: Executable not implemented for js
	mvdan.cc/sh/v3/interp_test.TestMain interp/interp_test.go:143
```

which is there to set `GOSH_PROG` for re-execing the test binary as a fake shell command. Subprocesses being precisely what js/wasm lacks, making the suite run there is a larger change than this PR — so I have left it at "compiles", since you said you would play with the lane on master after merging. Worth knowing that is the next obstacle, and that `PATH="$PATH:$(go env GOROOT)/lib/wasm"` does work: node runs the binary fine, it is only `TestMain` that stops it.

Also, a near miss worth flagging: squashing initially rebased across d6550df7 and my commit would have **deleted the new Contributing section** from the README. Caught before pushing, and README is untouched — but apologies for the noise if you saw an intermediate state.

Verified: `go test ./interp` passes natively, `GOOS=js GOARCH=wasm go build ./interp` succeeds, and `GOOS=js GOARCH=wasm go test -c ./interp` now compiles where it did not before.

0pcom:
Filed #1400 with a standalone reproducer, per the contributing guidelines — noticed the new section while rebasing. Happy to close it in favour of this PR if you would rather keep it in one place.

mvdan (approved):
Thanks!

PR #1395, opened 2026-08-18
interp: support exec's -a, -c and -l flags

### LINKED ISSUE #1386: interp: exec ignores its -a, -c and -l flags
`exec` accepts no flags. bash's is `exec [-cl] [-a name] [command [arguments]]`; the flag is passed through as the command name and fails PATH lookup.

```console
$ gosh -c 'exec -a nm /bin/echo hi'
"-a": executable file not found in $PATH   # exit 127
$ bash -c 'exec -a nm /bin/echo hi'
hi

$ gosh -c '(exec -c /bin/echo hi)'   → "-c": executable file not found in $PATH
$ gosh -c '(exec -l /bin/echo hi)'   → "-l": executable file not found in $PATH
```

`interp/builtin.go`'s `case "exec"` takes `args` and hands the first element to the exec handler with no option parsing, so the error comes from `interp/handler.go`'s PATH lookup and names the flag rather than the command.

`-a` is the one with a real consumer: it is how argv[0] is set without a wrapper binary, and it is what multi-call binaries dispatch on. A tool that ships several utilities behind one executable and installs them as shell functions (`(exec -a bfs "$bin" ...)`) has every one of those functions fail under `interp`, with a message that names neither the function nor the binary.
---

Found by a bash-vs-`gosh` differential sweep (128 snippets, each run direct and through `eval`) while integrating `interp` as the substrate of a shell. Reproduced on `gosh@v3.13.2-0.20260813221317-a04df9f0d4b8` against bash 5.3.15 on darwin/arm64. Happy to send a PR if the direction looks right.

blairham:
Fixes #1386.

`exec` took no flags, so the first one was treated as the command to run and failed PATH lookup with a message naming the flag:

```console
$ gosh -c 'exec -a nm /bin/echo hi'
"-a": executable file not found in $PATH
```

`-a` is the one with real consumers: it sets argv[0] without a wrapper binary, and multi-call binaries dispatch on it. A tool shipping several utilities behind one executable and installing them as shell functions has every one of those functions fail, with a diagnostic naming neither the function nor the binary.

## Approach

`-a` and `-l` need an argv[0] that differs from the file being looked up, and `-c` needs an empty environment, so all three need the exec handler's cooperation. `HandlerContext` gains two fields for that:

- `Argv0 string` — the name the program should see as its own argv[0].
- `ClearEnv bool` — run with an empty environment.

Both are **additive**: an `ExecHandlerFunc` that ignores them keeps working and simply does not support the flags; `DefaultExecHandler` honors them. Existing `Runner.exec` callers are untouched — the builtin goes through a new `execWith`, and `exec` delegates to it with the zero values.

`-c` is implemented rather than rejected because bash supports it, and the test suite confirms every case against real bash, which is what caught the first draft refusing it.

## Tests

Portable flag-error cases go in `runTests`; the three behavior cases need a program that reports its own argv[0] and environment, so they use `/bin/sh` in `runTestsUnix` (gosh sets `$0` itself, so it cannot observe argv[0]).

- `-a` overrides argv[0]; `-l` prefixes a dash; the two compose (`-l -a name` → `-name`)
- `-c` runs with an empty environment, and the same command without `-c` does not
- `exec -a` with no argument, and an unknown flag, both exit 2
- `exec -a name >/dev/null` with no command still keeps the redirection open, as bare `exec` does

`go test ./interp/` is at the same 4 pre-existing `TestRunnerRunConfirm` failures as `master` on this machine (three `HOME=…; echo ~` cases and one `>&-` write-error message, all environment-dependent), so no regressions. `gofmt` clean.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -528,12 +528,43 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		// the process. It's in theory what a shell should do,
 		// but in practice it would kill the entire Go process
 		// and it's not available on Windows.
+		argv0 := ""
+		login, clearEnv := false, false
+		fp := flagParser{remaining: args}
+		for fp.more() {
+			switch flag := fp.flag(); flag {
+			case "-a":
+				if len(fp.remaining) == 0 {
+					return failf(2, "exec: -a: option requires an argument\n")
+				}
+				argv0 = fp.value()
+			case "-l":
+				login = true
+			case "-c":
+				clearEnv = true
+			default:
+				return failf(2, "exec: invalid option %q\n", flag)
+			}
+		}
+		args := fp.args()
 		if len(args) == 0 {
+			// "exec" on its own keeps this statement's redirections open. Any
+			// flags then have nothing to apply to, as in bash.
 			r.keepRedirs = true
 			break
 		}
+		if argv0 == "" {
+			argv0 = args[0]
+		}
+		if login {
+			// A login shell is told an argv[0] prefixed with "-".
+			argv0 = "-" + argv0
+		}
 		r.exit.exiting = true
-		r.exec(ctx, pos, args)
+		if argv0 == args[0] {
+			argv0 = "" // nothing to override
+		}
+		r.execWith(ctx, pos, argv0, clearEnv, args)
 		exit = r.exit
 	case "command":
 		show := false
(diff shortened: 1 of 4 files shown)


mvdan:

blairham:
Hey, sorry, it was my Claude. If you don't want PR's that totally fine; I
don't mind forking and taking over. Sorry to have bothered you

mvdan:
You can still file issues if you wish, but you need to give me time to read them and process them. I can see you've closed them all now, I guess that's fine too.

Issue #1392, opened 2026-08-18
interp: several builtin options are unimplemented, and read -s reads the wrong file descriptor


blairham:
Several builtins print an "invalid option" diagnostic for an unimplemented flag and **still return 0**, so the caller cannot distinguish "refused" from "succeeded with an empty result".

```console
$ gosh -c 'printf "ab" | { read -r -d "" x; echo "[$x]"; }'
read: invalid option "-d"
[]                            # exit 0
$ bash -c 'printf "ab" | { read -r -d "" x; echo "[$x]"; }'
[a]
```

| construct | gosh | bash |
| --- | --- | --- |
| `read -r -d '' x` | `invalid option "-d"`, `x` empty, **exit 0** | `x=a` |
| `read -r -n 2 x` | `invalid option "-n"`, `x` empty, **exit 0** | `x=ab` |
| `read -r -s x` | **no message at all**, `x` empty, exit 0 | `x=hi` |
| `read -r -t 1 x` | `invalid option "-t"`, exit 0 | reads |
| `declare -i n; n=1+1` | `invalid option "-i"`, `n` is the string `1+1` | `n=2` |
| `declare -r v=1; v=2` | assignment accepted, exit 0 | `readonly variable`, exit 1 |
| `trap -p` | `"-p": NOT IMPLEMENTED flag`, exit 2 | prints traps |

`read -s` is the sharpest one: accepted with no diagnostic and no status, so a script reading a confirmation or a passphrase gets an empty string and continues. `read -d ''` is the standard safe way to consume `find -print0`, so its silent failure turns a filename-safe loop into a no-op.

`declare -i` leaves the literal source text in the variable, which makes every later arithmetic comparison wrong rather than absent. `declare -r` not rejecting an assignment means `readonly` used as a guard is decorative.

The exit status is the cheap half and worth separating from the features: an unsupported option should be non-zero regardless of which flags are eventually implemented. Today `read -d '' x` is indistinguishable to the caller from a successful read of an empty line, which is a legitimate outcome — so no amount of caller-side care detects it.

`declare -F` is implemented, and `-p` on a function name disagrees with bash (`declare: f: not found`).
---

Found by a bash-vs-`gosh` differential sweep (128 snippets, each run direct and through `eval`) while integrating `interp` as the substrate of a shell. Reproduced on `gosh@v3.13.2-0.20260813221317-a04df9f0d4b8` against bash 5.3.15 on darwin/arm64. Happy to send a PR if the direction looks right.

[...]

PR #1396, opened 2026-08-18
interp: implement read's -d, -n and -N, and fix -s and -a

blairham:
Fixes part of #1392.

`read` supported `-r`, `-p`, `-a` and `-s`. Four of bash's options were refused with a diagnostic and exit status 2, and two of the four matter beyond missing coverage:

* `read -d ''` is the standard way to consume `find -print0`, so without it a filename-safe loop cannot be written at all.
* `-s` read from `syscall.Stdin` — the process's own descriptor — rather than the shell's stdin. Under a redirect, in a pipeline, or when the caller supplied a reader via `StdIO`, it read the wrong input and assigned nothing. It also called `term.ReadPassword` on a descriptor which usually isn't a terminal, so it failed instead of falling back to a plain read:

```console
$ printf 'hi\n' > in.txt
$ gosh -c 'read -r -s x < in.txt; echo "st=$? x=[$x]"'   # before
st=1 x=[]
$ bash -c 'read -r -s x < in.txt; echo "st=$? x=[$x]"'
st=0 x=[hi]
```

`-a` had a separate bug, present regardless of this change: it did not consume its array name as the option's argument, relying on it being the first operand, so any option written after it was misparsed as an identifier. `read -a arr -d ''` — the print0-into-array idiom — failed with `read: invalid identifier "-d"`.

### What this implements

| option | behavior |
| --- | --- |
| `-d delim` | stop at `delim` instead of newline; an empty string means ASCII NUL, matching `mapfile -d ''` |
| `-n n` | stop after `n` characters or at the delimiter, whichever comes first |
| `-N n` | read exactly `n` characters, ignoring the delimiter, with no field splitting or trimming |
| `-s` | use the shell's stdin, and only suppress the echo when it is a terminal |
| `-a name` | take the array name as the option's argument |

`readLine` now takes the delimiter, a character count, and whether the delimiter should be ignored. The count is of characters that end up assigned rather than bytes read, since the backslashes which escape another character are dropped later — bash counts the same way, so `printf 'a\bc' | read -N 3 x` gives `x=abc` in both.

An escaped delimiter becomes a literal character rather than ending the line, except for the existing newline continuation.

### Still missing

`-t` and `-u` are unimplemented, and `-s` cannot suppress a terminal's echo when combined with `-d`, `-n` or `-N`, as `x/term` only exposes a whole-line password read. I left those out rather than approximate them; happy to follow up if you'd like them in this PR.

### Testing

18 cases added to `runTests`, so they run through `TestRunnerRunConfirm` and are confirmed against real bash 5.3 rather than only against what this commit implements. I also ran a separate 37-case differential harness (patched `gosh` vs bash 5.3.15 on darwin/arm64) covering statuses, IFS trimming, escape handling and the `-n`/`-N`/`-d` interactions: 37/37 agree.

`go test ./...` shows the same four pre-existing `TestRunnerRunConfirm` failures as `master` on this machine (three tilde/`HOME` cases and `echo foo >&- 2>&-`), confirmed by stashing the patch and re-running.

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -14,7 +14,6 @@ import (
 	"slices"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"golang.org/x/term"
@@ -666,7 +665,14 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 		var prompt string
 		raw := false
 		silent := false
-		readArray := false
+		arrayName := ""
+		delim := byte('\n')
+		// maxChars is the count given to -n or -N; a negative value means that
+		// the read is only stopped by the delimiter or by the end of the input.
+		maxChars := -1
+		// exactly is set by -N, which reads a fixed number of characters,
+		// ignoring the delimiter and doing no field splitting.
+		exactly := false
 		fp := flagParser{remaining: args}
 		for fp.more() {
 			switch flag := fp.flag(); flag {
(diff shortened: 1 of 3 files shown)


mvdan:
@blairham as I said in another thread, please slow down and stop sending PRs.

blairham:
One clarification on the evidence, so the claim is not stronger than it is: of the 18 cases, 13 assert exact combined output and so are fully confirmed against real bash by `TestRunnerRunConfirm`. The other 5 are the option-error cases marked `#JUSTERR`, where the harness only checks that bash also errors — it does not compare the message or the status. I did verify those five separately (bash gives status 2 for a missing option argument and 1 for an invalid number, which is what this implements), but that verification lives in my own harness rather than in the suite.

mvdan:

Issue #1398, opened 2026-08-18
interp: $- is unset, so it is empty and errors under set -u


blairham:
`$-`, the POSIX special parameter holding the current option flags, is never set.

```console
$ gosh -c 'echo "[$-]"'
[]
$ bash -c 'echo "[$-]"'
[hBc]

$ gosh -c 'set -e; echo "[$-]"'
[]
$ bash -c 'set -e; echo "[$-]"'
[ehBc]
```

Being empty rather than absent would already be a divergence, but it is genuinely unset, so `set -u` turns any reference into a fatal error:

```console
$ gosh -c 'set -u; echo "[$-]"'
-: unbound variable
$ bash -c 'set -u; echo "[$-]"'
[uhBc]
```

That combination is the awkward one: `set -euo pipefail` followed by anything that consults `$-` aborts. The common idioms are interactivity checks in shell rc files and libraries that save and restore options:

```sh
case $- in *i*) ... ;; esac        # sees no flags at all
[[ $- == *e* ]] && restore_errexit=1
```

The interactivity check happens to reach the right answer today, since an empty `$-` contains no `i` either — but only by accident, and only while `interp` has no interactive mode.

`$-` should hold the single-letter names of the enabled options, which `interp` already tracks in `posixOptsTable`; note that `pipefail` has no letter there (` `), so it is `-o`-only in bash too and should be left out.

Reproduced on `master` at d6550df7 against bash 5.3.15 on darwin/arm64, found while measuring `set` for #1392.

Issue #1389, opened 2026-08-18
interp: redirections to fd 3 and above are accepted and their data discarded


blairham:
A redirection to any file descriptor above 2 is accepted and its data is discarded. Exit status is 0.

```console
$ gosh -c 'f=$(mktemp); exec 3> "$f"; echo hi >&3; exec 3>&-; cat "$f"'
                              # empty, exit 0
$ bash -c 'f=$(mktemp); exec 3> "$f"; echo hi >&3; exec 3>&-; cat "$f"'
hi
```

Every spelling is affected, so it looks like the fd machinery rather than `exec`:

| construct | gosh | bash |
| --- | --- | --- |
| `exec 3> f; echo hi >&3` | nothing written | `hi` |
| `exec 3< f; read -r x <&3` | `x` empty | `x=hi` |
| `{ echo hi >&3; } 3> f` | nothing written | `hi` |
| `exec {v}> f; echo hi >&$v` | nothing written | `hi` |

The third row has no `exec` in it — a compound command with its own fd-3 redirection — and `{v}>` (automatic fd allocation) fails the same way.

fd 3+ is how scripts separate a log or trace channel from stdout/stderr, how `flock -x 3` gets its lock fd, how `read -u 3` reads a file while the loop body still owns stdin, and how the stdout/stderr swap idiom (`{ cmd 2>&1 1>&3; } 3>&1`) works. Each produces an empty artifact and reports success.

Possibly related: #1142.

`coproc` is separately unimplemented (`unhandled command node: *syntax.CoprocClause`) and needs the same plumbing.
---

Found by a bash-vs-`gosh` differential sweep (128 snippets, each run direct and through `eval`) while integrating `interp` as the substrate of a shell. Reproduced on `gosh@v3.13.2-0.20260813221317-a04df9f0d4b8` against bash 5.3.15 on darwin/arm64. Happy to send a PR if the direction looks right.

PR #1399, opened 2026-08-18
interp: implement the >| and <> redirections, and set -C

### LINKED ISSUE #1394: interp: the >| clobber redirect writes nothing
The `>|` redirect (clobber, overriding `noclobber`) writes nothing. No diagnostic, exit 0.

```console
$ gosh -c 'f=$(mktemp); echo hi >| "$f"; printf "content=[%s]
" "$(cat $f)"'
content=[]
$ bash -c 'f=$(mktemp); echo hi >| "$f"; printf "content=[%s]
" "$(cat $f)"'
content=[hi]
```

`RdrClob` does not appear anywhere in `interp` — the redirect switch in `interp/runner.go` handles `RdrIn`, `RdrOut`, `AppOut`, `RdrAll`, `AppAll` and `DplIn`/`DplOut`, so the op is not one it acts on. The parser accepts it, which is what makes the failure silent rather than an error.

Because `noclobber` (`set -C`) is also unsupported (#1387), `>|` currently has nothing to override, so treating it as a plain `RdrOut` is exactly right until it does — and it is the difference between writing the file and silently not writing it.

Worth pairing with #1387 for that reason: whoever implements `noclobber` needs `>|` to be the escape hatch, and whoever implements `>|` first makes it correct in the meantime.

Practical note on how this was found: a shell built on `interp` had to patch it downstream, rewriting `syntax.RdrClob` to `syntax.RdrOut` in the AST before handing it to the interpreter. That works, but only for the statements which pass through the rewrite — under `eval` and under `source` the original op survives and the write silently vanishes again. The consequence is worth stating because the construct's main consumer is machine-generated: `echo "# Snapshot file" >| "$SNAPSHOT_FILE"` is the first line of a shell snapshot generator, and a snapshot which is silently never written is a hard failure to attribute from the output.

---

Found by a bash-vs-`gosh` differential sweep while integrating `interp` as the substrate of a shell. Reproduced on `gosh@v3.13.2-0.20260813221317-a04df9f0d4b8` against bash 5.3.15 on darwin/arm64. Happy to send a PR.

blairham:
Fixes #1394. Also implements the `-C` part of #1392's `set` gaps.

Neither `>|` nor `<>` had a case in the redirect switch, so both fell through to its `default` and returned an error which `stmtSync` drops. They did nothing at all, with no diagnostic, and exit status 1:

```console
$ gosh -c 'echo hi >| f'; echo "st=$?"; cat f
st=1
cat: f: No such file or directory
$ bash -c 'echo hi >| f'; echo "st=$?"; cat f
st=0
hi

$ gosh -c 'echo hi > f; read -r l <> f; echo "[$l]"'
[]
$ bash -c 'echo hi > f; read -r l <> f; echo "[$l]"'
[hi]
```

`>|` is only meaningful as noclobber's escape hatch, and noclobber was itself missing, so this adds `set -C` too — they are two halves of one behavior and implementing either alone leaves the other wrong.

### noclobber

Matching bash: `-C` refuses to truncate an existing **regular** file, appending is unaffected, a file which does not exist yet is fine, and non-regular files such as `/dev/null` are not protected. `>|` ignores the option.

```console
$ gosh -c 'set -C; echo old >f; echo new >f; echo "st=$?"; cat f'
f: cannot overwrite existing file
st=1
old
```

### One implementation note

The diagnostic is printed where it is raised rather than by the caller of `Runner.redir`. I first tried reporting all of `redir`'s errors in `stmtSync`, which is tempting since it would also make `3>` and the unhandled dup arguments stop failing silently — but `Runner.open` already reports its own errors, and the handler's fatal errors go through `exit.fatal`, so a blanket report prints those twice (`TestRunnerHandlers/OpenForbidNonDev` catches it). Untangling that is worth doing, but it belongs with #1389 rather than here, so the sibling gaps in this switch are deliberately left silent.

### Testing

12 cases added to `runTests`. Three assert the diagnostic and are marked `#IGNORE bash prefixes its diagnostics`; each of those has a twin which suppresses stderr *before* the failing redirect, so bash confirms the status and the file contents through `TestRunnerRunConfirm`.

That ordering matters and is not cosmetic: with `2>/dev/null` written *after* the failing redirect, both shells still print the message, because redirections are applied left to right and the failure happens before stderr is replaced. My first draft of these tests put it after and normalized bash's `bash: line N:` prefix away, which hid the message entirely and made a correct implementation look wrong.

`go test ./...` shows the same four pre-existing `TestRunnerRunConfirm` failures as `master` on this machine (three tilde/`HOME` cases and `echo foo >&- 2>&-`), confirmed by stashing and re-running. Separately verified against bash 5.3.15 on darwin/arm64 with an 18-case differential harness: the only two divergences left are `$-` not listing `C` (#1398) and `set -o`'s column padding, neither introduced here.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -618,6 +618,7 @@ var posixOptsTable = [...]posixOpt{
 	// sorted alphabetically by name
 	{'a', "allexport"},
 	{'e', "errexit"},
+	{'C', "noclobber"},
 	{'n', "noexec"},
 	{'f', "noglob"},
 	{'u', "nounset"},
@@ -747,13 +748,14 @@ const (
 	// These correspond to indexes in [shellOptsTable]
 	optAllExport = iota
 	optErrExit
+	optNoClobber
 	optNoExec
 	optNoGlob
 	optNoUnset
 	optXTrace
 	optPipeFail
 
-	// These correspond to indexes (offset by the above seven items) of
+	// These correspond to indexes (offset by the above eight items) of
 	// supported options in [bashOptsTable]
 	optDotGlob
 	optExpandAliases
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -1608,6 +1608,57 @@ var runTests = []runTest{
 		"",
 	},
 	{
+		// `>|` overwrites the file whether or not noclobber is set
+		"echo old >f; echo new >|f; cat f",
+		"new\n",
+	},
+	{
+		// `<>` opens for reading and writing without truncating
+		"echo hi >f; read -r l <>f; echo \"[$l]\"; cat f",
+		"[hi]\nhi\n",
+	},
+
+	// noclobber
+	{
+		`set -C; echo old >f; echo new >f; echo "st=$?"; cat f`,
+		"f: cannot overwrite existing file\nst=1\nold\n #IGNORE bash prefixes its diagnostics",
+	},
+	{
+		// the same without the diagnostic, so that bash confirms the behavior
+		`set -C; echo old >f; echo new 2>/dev/null >f; echo "st=$?"; cat f`,
+		"st=1\nold\n",
+	},
+	{
+		// writing to a file which does not exist yet is allowed
+		`set -C; echo new >f; cat f`,
+		"new\n",
+	},
+	{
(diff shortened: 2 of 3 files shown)


mvdan:

Issue #1120, opened 2025-01-16
FIFO File Descriptors Created in process substitution aren't handled consistently


cfogrady:
At the moment an `OpenHandler` can  be provided to an `interp.Runner` as a `RunnerOption`.

When a process substitution creates a FIFO path it is created on the os filesystem; however, later operations sent to it treat it as a normal file sent to the `OpenHandler`. This has a side effect that if the OpenHandler is mocked or is sandboxed in some way, that the open file for the process substitution's named pipe hangs because writes are going to the wrong file.

I would propose one of the following:
1) Changing documentation such that users are specifically warned against sandboxing OpenHandler because of such scenarios
2) Changing the https://github.com/mvdan/sh/blob/master/interp/runner.go#L107 and https://github.com/mvdan/sh/blob/master/interp/runner.go#L120 to use the `OpenHandler` instead of directly interacting with the OS, which may also involve a handler for removing the file
3) Provide a way to internally differentiate between Named Pipes and regular files before the `OpenHandler` is called.

mvdan:
Thanks for filing this; I think this was simply an oversight. I'll need to think about this because, as you point out, there's no clear and easy solution here.

Named pipes themselves are also created directly and not via any handler, so that's also part of the problem. At least that one is excusable, given that one can't create a named pipe via `os.OpenFile`.

Given that named pipes are somewhat special (they need to be created in a special way, and they need to be removed later) perhaps they do need to be split apart from the open handler entirely. The named pipes are created by the interpreter internals anyway, and get placed in temporary directories so that they're out of the way - so I also don't think that anyone writing a "file handler sandbox" should care about them.

Hence I currently lean towards a form of option 3, perhaps even so that OpenHandler is not called at all for named pipes created by the interpreter.

mvdan:
Implemented the third option. I think FIFOs (named pipes) could become another handler if users need that, although I'm not sure if they can be implemented any way other than real named pipes on disk.

PR #1404, opened 2026-08-27
interp: add ProcSubstOpen handler for process substitution

aki-kong:
Process substitution always creates a real named pipe on disk via mkfifo
and os.OpenFile, no matter what OpenHandler/StatHandler/AccessHandler are
set to, so it's a hole straight to the real filesystem for anyone
sandboxing file I/O.

This is #1120, where you left the door open for a handler:

> perhaps they do need to be split apart from the open handler entirely...
> I think FIFOs could become another handler if users need that.

I've taken a crack at implementing it. ProcSubstOpen owns the whole FIFO
lifecycle (create, open, remove), since a plain (path, flag, perm) open
can't express mkfifo + cleanup. DefaultProcSubstOpen is the old logic
moved into handler.go, same behavior, except the actual open is now
deferred to the first read/write, since opening a FIFO blocks until a
peer connects and that peer isn't running yet when the path is returned.

This also meant removing the bit in r.open that special-cased FIFO
paths to skip OpenHandler, since creation is now handler-owned. Happy
to reconsider that if you'd rather keep it separate.

Updated the tests accordingly, including a new one (ProcSubstVirtual)
that pairs ProcSubstOpen with OpenHandler over an in-memory io.Pipe and
checks nothing touches the real filesystem.

## Motivation

I'm using this shell intepreter package to implement a simple Bash sandbox with WASI as the execution provider which supports [`fs.FS`](https://pkg.go.dev/io/fs#FS). I had implemented the `OpenHandler` based on the `fs.FS` but just for the ProcessSubstitution there seemed to be a sandbox breakage. In this case, I can implement the ProcSub + OpenHandler to handle the Pipe implementation (plan to use channels). Left the DefaultHandler to the original behaviour.

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -105,6 +105,10 @@ type Runner struct {
 	// accessHandler is a function responsible for checking file access. It must be non-nil.
 	accessHandler AccessHandlerFunc
 
+	// procSubstOpen is a function responsible for creating the communication
+	// channel for process substitutions. It must be non-nil.
+	procSubstOpen ProcSubstOpenFunc
+
 	stdin  stdinFile // e.g. the read end of a pipe
 	stdout io.Writer
 	stderr io.Writer
@@ -332,6 +336,7 @@ func New(opts ...RunnerOption) (*Runner, error) {
 		readDirHandler:       DefaultReadDirHandler2(),
 		statHandler:          DefaultStatHandler(),
 		accessHandler:        DefaultAccessHandler(),
+		procSubstOpen:        DefaultProcSubstOpen(),
 	}
 	r.dirStack = r.dirBootstrap[:0]
 	// turn "on" the default Bash options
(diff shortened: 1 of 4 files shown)


aki-kong:
cc @mvdan Can you please take a look

mvdan:
Thanks. This PR is largely in the right direction but I'm afraid it is incorrect. You are removing part of a test which existed explicitly to cover https://github.com/mvdan/sh/issues/1120, and that bug is reintroduced here. I will take a look at a fix myself.

aki-kong:
> Thanks. This PR is largely in the right direction but I'm afraid it is incorrect. You are removing part of a test which existed explicitly to cover https://github.com/mvdan/sh/issues/1120, and that bug is reintroduced here. I will take a look at a fix myself.

I felt the same. Somewhere there should be some sort of overlap/dependency between the pipe handler and the open handler. Let me know the approach, I can help with the changes as well (if you feel like it 😀)

Issue #1214, opened 2025-10-23
zsh: support double quotes around nested parameter expressions


LangLangBart:
```bash
./shfmt --language-dialect zsh <<<'${(f)"$(foo)"}'
```

#### actual

```bash
<standard input>:1:6: invalid parameter name
```

#### expected

```bash
${(f)"$(foo)"}
```

### related

- #1209
- #1208

### zsh docs

> _Note that double quotes may appear around nested expressions, in which case_
> _only the part inside is treated as quoted; for example, ${(f)"$(foo)"} quotes_
> _the result of $(foo), but the flag ‘(f)’ (see below) is applied using the_
> _rules for unquoted expansions._
>
> Source: https://zsh.sourceforge.io/Doc/Release/Expansion.html#Parameter-Expansion#:~:text=Note%20that%20double%20quotes%20may%20appear%20around%20nested%20expressions

mvdan:
I continue to be baffled by the amount of shenanigans that parameter expansions in Zsh allow...

PR #1406, opened 2026-08-29
syntax: fix parsing a quoted $# in zsh

### LINKED ISSUE #1405: zsh: quoted "$#" is corrupted to "$" (silent semantic change)
## Reproduction

```console
$ printf '%s\n' \
    'if [ "$#" -lt 2 ]; then' \
    '  echo "$#"' \
    'fi' | shfmt -ln=zsh
if [ "$" -lt 2 ]; then
	echo "$"
fi
```

Tested on v3.14.0 and on current master (built via `go install mvdan.cc/sh/v3/cmd/shfmt@master`).

## Expected

```sh
if [ "$#" -lt 2 ]; then
	echo "$#"
fi
```

`"$#"` must be preserved. The current output silently changes semantics: in Zsh, `$` alone expands to the PID of the shell, while `$#` is the number of positional parameters. `zsh -n` cannot catch this, as `"$"` is syntactically valid.

## Scope

Only the exact form `"$#"` (double-quoted, closing quote directly after `#`) is corrupted. These all round-trip correctly in `-ln=zsh`:

```sh
echo $#          # OK
echo "$#x"       # OK
echo "a$#b"      # OK
echo "${#}"      # OK
echo ${#arr}     # OK
echo $#PATH      # OK
```

And `"$#"` is fine in `-ln=bash`.

## Additional notes

This seems related to the area of quoted special parameters (e.g. #1298). The formatter producing output that silently changes program behavior seems worth fixing at high priority, as it defeats trust in `shfmt -w` for Zsh files.

StudentWeis:
Fixes #1405.

## The problem

In Zsh mode, a double quote directly following `$#` in the short form
was treated as the start of a quoted parameter name. The parser then
consumed both the `#` and the `"` as a length prefix, failed to parse
a parameter name afterwards, and the expansion collapsed to a bare `$`:

```console
$ printf 'if [ "$#" -lt 2 ]; then\n\techo "$#"\nfi\n' | shfmt -ln=zsh  # before
if [ "$" -lt 2 ]; then
	echo "$"
fi
```

This silently changes semantics, as `$` expands to the shell PID in Zsh,
and `zsh -n` cannot detect it.

## Root cause

`paramNameStart` accepted `"` as the start of a parameter name, which
is correct for the long form, where Zsh nested expansions in quotes
like `${#"${foo}"}` exist (#1214). But in the short form the `"` is
never part of the expansion: real Zsh treats `"$#"` as the special
parameter, and `$#"$foo"` as `$#` followed by a quoted word:

```console
% set -- x y; echo "$#"; echo $#"foo"
2
2foo
```

The same applied to the `%` and `+` short prefixes.

## The fix

Pass whether the expansion is the short form into `paramNameStart` and
refuse a leading `"` in that case. The long form keeps accepting
quoted parameter names, so `${#"${foo}"}` still parses as before.

Added parser regression tests for `"$#"` and `$#"$foo"`.

### DIFF
--- a/syntax/filetests_test.go
+++ b/syntax/filetests_test.go
@@ -2339,6 +2339,14 @@ var fileTests = []fileTestCase{
 		langFile(dblQuoted(litParamExp("!"))),
 	),
 	fileTest(
+		[]string{`"$#"`},
+		langFile(dblQuoted(litParamExp("#"))),
+	),
+	fileTest(
+		[]string{`$#"$foo"`},
+		langFile(word(litParamExp("#"), dblQuoted(litParamExp("foo")))),
+	),
+	fileTest(
 		[]string{`$`, `$ #`},
 		langFile(litWord("$")),
 	),
--- a/syntax/parser.go
+++ b/syntax/parser.go
@@ -1518,22 +1518,22 @@ zshPrefixLoop:
 		// Note that in Zsh, the short form like $#name is allowed too.
 		switch p.r {
 		case '#':
-			if p.paramNameStart() {
+			if p.paramNameStart(pe.Short) {
 				pe.Length = true
 			}
 		case '%':
-			if p.paramNameStart() {
+			if p.paramNameStart(pe.Short) {
 				p.checkLang(pe.Pos(), LangMirBSDKorn, "`${%%foo}`")
 				pe.Width = true
 			}
 		case '!':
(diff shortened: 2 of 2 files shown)


mvdan:
See: 019025c6e6697b6d114f61ffa8a067548fa62860

PR #1407, opened 2026-08-29
README: add PyPI to the list of packages

MaxWinterstein:
shfmt-py packages the released shfmt binaries as wheels on PyPI, so `pip install shfmt-py` (or `uv tool install` / `pipx`) puts shfmt on PATH. Packaging only: no patches, no wrapper code, and the wheels ship your BSD-3-Clause notice next to the binary.

I have maintained it since 2021, and Renovate opens the version bumps, so it usually follows a release within a day or two. 3.14.0 is already on master.

Why I think it earns a word in that list: in Python projects and CI images with no Go toolchain and no root for the system package manager, pip is the package manager that is already there. That is a different audience from the distro packages listed.

It does also ship a pre-commit hook, but that is not what I am asking about here, scop/pre-commit-shfmt has that covered. This is only about the PyPI entry.

Fine by me if you would rather keep that list to system package managers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ For more information, see [its manpage](cmd/shfmt/shfmt.1.scd), which can be
 viewed directly as Markdown or rendered with [scdoc].
 
 Packages are available on [Alpine], [Arch], [Debian], [Docker], [Fedora], [FreeBSD],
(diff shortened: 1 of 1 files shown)


mvdan (approved):
Thanks!

PR #1408, opened 2026-08-30
syntax: preserve indentation in nested tab heredocs

### LINKED ISSUE #1403: Nested tab heredoc formatting bug
Hi! First of all, thanks for the amazing project!

`shfmt` reformats my nested tab-prefixed heredoc the wrong way. An example:
```sh
sudo systemd-tmpfiles --create - <<-EOF
	f+~ /etc/configfile $(
		base64 -w0 <<-INNER
			content
		INNER
	)
EOF
```
is reformatted into
```sh
sudo systemd-tmpfiles --create - <<-EOF
	f+~ /etc/configfile $(
		base64 -w0 <<-INNER
content
				INNER
	)
EOF
```
with `shfmt -i0` (because those are tabs, take care when copying). The actual content of the nested heredoc is completely deindented, and the `INNER` end marker is indented too far (seems to gain an extra tab for every line in the inner heredoc).

I'm not in a hurry, but this seems wrong. `shfmt` support is important enough for me that I wont use this construct until it works 😄. Once again thanks for the project, and maybe for fixing this!

err0r4eve3:
## Summary

- route byte-slice writes through `extraIndenter.WriteByte`, matching its existing string path
- add a regression case for nested `<<-` heredocs

The bug is specific to nesting because an inner `extraIndenter` flushes each completed line with `e.bufWriter.Write`. When that underlying writer is an outer `extraIndenter`, its embedded `Write` method was promoted and bypassed the outer writer's per-line indentation state. Implementing `Write` on `extraIndenter` keeps every writer entry point on the same path.

## Test

- `go test ./...`

Fixes #1403

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1620,6 +1620,13 @@ func (e *extraIndenter) WriteString(s string) (int, error) {
 	return len(s), nil
 }
 
+func (e *extraIndenter) Write(b []byte) (int, error) {
+	for i := range len(b) {
+		e.WriteByte(b[i])
+	}
+	return len(b), nil
+}
+
 func startsWithLparen(node Node) bool {
 	switch node := node.(type) {
 	case *Stmt:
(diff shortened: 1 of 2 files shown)


milhnl:
?

mvdan:
Let's pretend you did not just copy paste someone else's patch and put your name on it.

err0r4eve3:
You are right. This PR reproduced the patch @milhnl had already posted in #1403, including the regression test, and I failed to credit that source before submitting it. I apologize to @milhnl and the maintainers. The attribution failure is entirely mine. The PR is closed, and I will not submit further work to this repository.

PR #1410, opened 2026-09-02
interp: support shopt -q to query option states

hktitof:
shopt -q was being rejected as an unsupported option like -p, but bash uses it to query the current value of one or more options

it prints nothing, and the exit status is 0 when all of the named options are set and 1 otherwise

so this adds a -q flag to the shopt builtin that reads the option states without printing, and reports the result via the exit status, like bash does

a few things match bash too:
- shopt -q with no names is a no-op that returns 0
- shopt -q -o <opt> queries posix options
- combining -q with -s or -u still does the set or unset, -q only suppresses the output
- an invalid option name still errors

tested against bash 5.2 with the confirm suite, plus new table cases that fail without the change

Signed-off-by: titof <hktitof@gmail.com>

### DIFF
--- a/interp/builtin.go
+++ b/interp/builtin.go
@@ -816,6 +816,7 @@ func (r *Runner) builtin(ctx context.Context, pos syntax.Pos, name string, args 
 
 	case "shopt":
 		mode := ""
+		quiet := false
 		posixOpts := false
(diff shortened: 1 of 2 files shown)


mvdan (approved):
thanks!

PR #1415, opened 2026-09-05
fix(printer): keep inline command with heredoc in command substitution

### LINKED ISSUE #1282: Formatting issue: Command in quoted subshell with HEREDOC should not be moved to newline
Thanks for a great tool.

If you disagree and think this is just a matter of personal preference, no problem. Feel free to close. I think it looks a bit odd, but perhaps it makes sense in other contexts.

I use [docopts](https://github.com/docopt/docopts) in my shell scripts. Using docopts requires running `eval "$(docopts <<EOF`. shfmt moves the word `docopts` and everything after it to a new line, which I think looks odd. Here's a file before and after formatting. I would rather "shfmt" leave _Before_ unchanged.

This is a case where I would have used something like `# shfmt-disable-next-line` if it were available, to prevent shfmt from changing this line, but the shfmt docs recommend opening an issue instead when one feels like doing that, so that's why I've opened this issue.

### Before

```bash
#!/usr/bin/env bash

eval "$(docopts --help=- -A ARGS : "$@" <<EOF
Run "ls". This is just an example for a bug report.

Usage:
	example-command
	example-command -h | --help

Options:
	-h, --help
		Show this documentation
EOF
)"

ls
```

### After

```bash
#!/usr/bin/env bash

eval "$(
	docopts --help=- -A ARGS : "$@" <<EOF
Run "ls". This is just an example for a bug report.

Usage:
	example-command
	example-command -h | --help

Options:
	-h, --help
		Show this documentation
EOF
)"

ls
```

dualfroz:
## Problem

- Repo: mvdan/sh
- Base: mvdan:master
- Head: dualfroz:fix/quoted-subshell-heredoc
- Closes: #1282

## Problem

A command written inline inside a command substitution, when it carries a
heredoc, was reformatted onto its own indented line. Reported in #1282 with
`eval "$(docopts ... <<EOF ... EOF)"`.

Before (input, and the wanted output):

```bash
eval "$(docopts --help=- -A ARGS : "$@" <<EOF
usage
EOF
)"
```

After (what shfmt produced, unwanted):

```bash
eval "$(
	docopts --help=- -A ARGS : "$@" <<EOF
usage
EOF
)"
```

## Root cause

`nestedStmts` forces a newline before the first statement when the closing
token (`)` here) sits on a line after the statements end:

```go
case closing.Line() > p.line && len(stmts) > 0 &&
	stmtsEnd(stmts, last).Line() < closing.Line():
	p.wantNewline = true
```

That heuristic is meant to preserve a deliberate layout such as:

```
$(stmt
)
```

When the last statement ends with a heredoc, `stmtsEnd` is the heredoc body
end, and the closing `)` is necessarily on the line after the heredoc
terminator. So the condition always fires, even though that newline is
mandatory shell syntax rather than a stylistic choice, and the command is
pushed onto its own line.

## Fix

Skip that heuristic when the last statement has a command and ends with a
heredoc redirect (`<<` or `<<-`). In that case only the heredoc terminator
forces the newline before the closing token, so the command stays inline.

A bare heredoc redirect with no command (for example `<(<<EOF ... EOF)`) is
left to the existing layout rules, so its current break-out behaviour and the
golden test that pins it are unchanged. A genuine user layout that puts the
command on its own line after `$(` is still preserved, because that is driven
by the statement's own position, not by this heuristic.

### DIFF
--- a/syntax/printer.go
+++ b/syntax/printer.go
@@ -1500,7 +1500,8 @@ func (p *Printer) nestedStmts(stmts []*Stmt, last []Comment, closing Pos) {
 		//     { stmt; stmt; }
 		p.wantNewline = true
 	case closing.Line() > p.line && len(stmts) > 0 &&
-		stmtsEnd(stmts, last).Line() < closing.Line():
+		stmtsEnd(stmts, last).Line() < closing.Line() &&
+		!lastStmtHeredocForcesClosing(stmts, last):
 		// Force a newline if we find:
 		//     { stmt
 		//     }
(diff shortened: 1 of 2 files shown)


mvdan:

PR #1417, opened 2026-09-05
interp: Fix brace expansion test case on 32-bit

nileshpatra:
9999999999 exceeds 32-bit range and does not lead to an error on 32-bit platforms. This can be observed in the debian package:

https://buildd.debian.org/status/fetch.php?pkg=golang-mvdan-sh&arch=armhf&ver=3.14.0-1&stamp=1788648229&raw=0

Change this to a sufficiently large value 1999999999 which still correctly triggers the stderr and fits in 32-bit range.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -3763,7 +3763,7 @@ done <<< 2`,
 		"brace expansion would exceed 16384 elements\n #IGNORE bash has no defensive limit below MaxInt",
 	},
 	{
-		"echo a{0..9999999999}b",
+		"echo a{0..1999999999}b",
 		"brace expansion would exceed 16384 elements\n #JUSTERR bash errors with a different message",
 	},
 


mvdan:
Oops, thanks for catching this. We already test on 32-bit in CI via `GOARCH=386 go test ./...`, but that only covers the Go code, and not the external tests with Bash, which we only test against 64 bits.

There is indeed a bug but I'm afraid that the fix here is not quite correct. You're fixing 32 bits but breaking 64 bits. The better fix is to not check this against Bash. I'll do that fix myself.

nileshpatra:
With this fix I can see 64 bir tests also passing on debian.

buildd.debian.org/golang-mvdan-sh which is why I proposed the patch. 

On 6 September 2026 5:53:40 pm IST, "Daniel Martí" ***@***.***> wrote:
>mvdan left a comment (mvdan/sh#1417)
>
>Oops, thanks for catching this. We already test on 32-bit in CI via `GOARCH=386 go test ./...`, but that only covers the Go code, and not the external tests with Bash, which we only test against 64 bits.
>
>There is indeed a bug but I'm afraid that the fix here is not quite correct. You're fixing 32 bits but breaking 64 bits. The better fix is to not check this against Bash. I'll do that fix myself.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/mvdan/sh/pull/1417#issuecomment-5559186686
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
Best,
Nilesh

mvdan:
Bash uses different bit sizes on 32 vs 64 bits, so I'm afraid it was causing inconsistent results on my side. Anyway, that test didn't need to cover Bash.

mvdan:
Fix released with https://github.com/mvdan/sh/releases/tag/v3.14.1.

PR #1420, opened 2026-09-07
cmd/shfmt,syntax: preserve zsh associative keys

### LINKED ISSUE #1233: zsh: associative arrays with hyphens in key get mangled
zsh allows hyphens in the keys for associative arrays, but shfmt adds spaces around the hyphens breaking them

Input
```
foo() {
  local -A dot_parse_opts_errors=()
  dot_parse_opts_errors[too-many-positional]="You may only specify one array name."

  <...>
}
```

Output
```
foo() {
  local -A dot_parse_opts_errors=()
  dot_parse_opts_errors[too - many - positional]="You may only specify one array name."

  <...>
}
```

StudentWeis:
Formatting `h[foo-bar]=1` in Zsh mode currently inserts spaces and breaks a valid associative-array assignment. The same assumption corrupts keys containing `/` or `+` and rejects `:`. Zsh determines whether a subscript is arithmetic or an associative key at runtime, so the parser cannot safely infer its meaning from the expression or a preceding declaration.

Add the opt-in parser option `ZshSubscriptsAsWords` and enable it in shfmt. It parses unflagged Zsh subscripts as words, preserving literal operators and whitespace while continuing to parse nested shell expansions. Track nested and escaped brackets in the lexer. Retain existing flagged-subscript parsing and Bash/mksh arithmetic parsing.

The library default remains unchanged for v3 compatibility. Existing parser fixtures retain their original expectations; new tests cover opt-in word nodes, disabling the option on a reused parser, and unchanged Bash/mksh behavior. A `TODO(v4)` records switching the default in a future major version. The first commit records the current incorrect output; the second fixes CLI formatting and adds behavior regressions.

Validation:

- Full Go package tests, `go vet ./...`, and a 20-second `FuzzZshSubscripts` run covering the opt-in path.
- Zsh 5.9 comparisons for associative keys, indexed arithmetic, slices, negative indices, flags, nested subscripts, command/arithmetic expansions, and escaped brackets.
- Default, minified, and single-line printing; one-byte readers and a long key crossing lexer buffers.
- CLI testscript coverage of explicit Zsh mode, filename detection, minification, and unchanged Bash formatting.
- Assertions use `qt`; external-shell subtests use the existing Zsh requirement helper. Missing-Zsh checks verify skipping normally and failing with `REQUIRE_SHELLS=1`.

This adds a parser option without changing existing exported signatures or the default AST. Library callers formatting Zsh should explicitly enable the option; callers using the legacy arithmetic AST can keep their existing configuration. The proposed option still needs the usual upstream API review.

Fixes #1233.

### DIFF
--- a/cmd/shfmt/main.go
+++ b/cmd/shfmt/main.go
@@ -210,7 +210,7 @@ For more information and to report bugs, see https://github.com/mvdan/sh.
 			useEditorConfig = false
 		}
 	})
-	parser = syntax.NewParser(syntax.KeepComments(true))
+	parser = syntax.NewParser(syntax.KeepComments(true), syntax.ZshSubscriptsAsWords(true))
 	printer = syntax.NewPrinter(syntax.Minify(minify.val))
 
 	syntax.RecoverErrors(expRecover.val)(parser)
--- a/cmd/shfmt/testdata/script/zsh-subscripts.txtar
+++ b/cmd/shfmt/testdata/script/zsh-subscripts.txtar
@@ -1,0 +1,30 @@
+stdin keys.zsh
+exec shfmt -ln=zsh
+cmp stdout keys.zsh
+! stderr .
+
+exec shfmt keys.zsh
+cmp stdout keys.zsh
+! stderr .
+
+stdin keys.zsh
+exec shfmt -ln=zsh -mn
+cmp stdout keys.zsh
+! stderr .
+
+stdin index.sh
+exec shfmt -ln=bash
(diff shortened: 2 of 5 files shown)


mvdan:
Thanks. I want help with the discussion, but no help is needed with the implementation. We also most likely do not want an option for this.

StudentWeis:
Thanks for the clarification and for looking into this. I appreciate your time and look forward to seeing your fix for Zsh associative subscripts!

